Issue #20, opened 2023-07-13
Improve signature API


franziskuskiefer:
The signature API isn't great right now. It should be updated to be easier to use and cover all signature algorithms.

franziskuskiefer:
We'll do this at some point, but closing for now as it is not on the roadmap for now.

PR #21, opened 2023-07-13
Cleanup & Version bumps

franziskuskiefer:
* bump hacl version
* add RSA PSS signature API (not nice, see #20)
* add 32-bit tests to CI


franziskuskiefer:
This is obsolete now.

PR #23, opened 2023-07-31
Added derandomized kyber to libjade-sys.

xvzcf:


src/jasmin.rs
```diff
@@ -1,5 +1,14 @@
 pub mod chacha20;
 pub mod poly1305;
+
+#[cfg(any(target_os = "linux", target_os = "macos"))]
```
franziskuskiefer, line 4: Let's add this guard to the `mod jasmin` in `lib.rs`.

src/digest.rs
```diff
@@ -121,13 +121,19 @@ pub type Sha3_512Digest = [u8; digest_size(Algorithm::Sha3_512)];
 
 macro_rules! sha3_impl {
     ($fun_name:ident, $output:ty, $jasmin_fun:expr, $hacl_fun:expr) => {
-        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+        #[cfg(all(
```
franziskuskiefer, line 124: Pulling up the guard, the ones in here can go as well.

franziskuskiefer (approved):
Thanks, looks good.

There are two follow ups
- [ ] add the avx2 version
- [ ] add kats for kyber on the top level

But lets get this in.

PR #24, opened 2023-08-01
kyber spec update

franziskuskiefer:
There are still a number of things that I think we should change.
But let's get something merged to `dev` to work against.

@xvzcf Please do a review to check if I messed up something.

Then we can do a quick round of review on your branch and get it in.


specs/hacspec-lib/src/bit_vector.rs
```diff
@@ -0,0 +1,208 @@
...
+    #[test]
+    fn better_bits() {
+
+        // 00000001 00000010 00000011 00000100 00000101 00000110 ...
+        //        1        2        3        4        5        6
+        let v = vec![1u8, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16];
+
+        let mut it = v.iter_sliced(7);
+        for i in bit..7 {
+            assert_eq!(v.bit(i), 0);
```
xvzcf, line 195: This test fails to build since no variable `bit` was declared in the function; also, did you mean `assert_eq!(it.bytes.bit(i), 0)` to test the iterator?

xvzcf:
Aside from the one comment, looks good!

specs/hacspec-lib/src/bit_vector.rs
franziskuskiefer: Ah, I thought I removed this again.
This was part of an experiment that was unnecessary. I'll delete it and what it pretends to test.

PR #33, opened 2023-08-04
Exposing reference implementation of Kyber in libcrux.

xvzcf:


src/hpke/kem.rs
```diff
@@ -243,7 +243,7 @@ fn shared_secret_from_dh(alg: KEM, mut secret: Vec<u8>) -> Vec<u8> {
 /// [`ValidationError`](`HpkeError::ValidationError`) as described in
 /// [validation](#validation-of-inputs-and-outputs).
 pub fn DH(alg: KEM, sk: &PrivateKeyIn, pk: &PublicKeyIn) -> Result<SharedSecret, HpkeError> {
-    match crate::ecdh::derive(kem_to_named_group(alg).into(), pk, sk) {
+    match crate::ecdh::derive(kem_to_named_group(alg).try_into().unwrap(), pk, sk) {
```
franziskuskiefer, line 246: Using the KEM API here requires us to unwrap now. That's not great. But nothing to fix in here.
Can you file an issue to investigate whether to re-design the KEM - EC interactions/APIs?

src/kem.rs
```diff
@@ -49,19 +54,21 @@ impl From<ecdh::Error> for Error {
...
 
 /// A KEM public key.
 pub enum PublicKey {
     X25519([u8; 32]),
     P256([u8; 64]),
+    Kyber768([u8; kyber768::PUBLIC_KEY_SIZE]),
 }
 
 /// Compute the public key for a private key of the given [`Algorithm`].
 pub fn secret_to_public(alg: Algorithm, sk: impl AsRef<[u8]>) -> Result<Vec<u8>, Error> {
```
franziskuskiefer, line 68: I'm not sure yet if we need this for Kyber. But the comment should say that it's only for EC algorithms.

src/kem.rs
```diff
@@ -78,7 +85,18 @@ pub fn key_gen(
 ) -> Result<(Vec<u8>, Vec<u8>), Error> {
     match alg {
         Algorithm::X25519 | Algorithm::Secp256r1 => {
-            ecdh::key_gen(alg.into(), rng).map_err(|e| e.into())
+            ecdh::key_gen(alg.try_into().unwrap(), rng).map_err(|e| e.into())
+        }
+
+        Algorithm::Kyber768 => {
+            let mut seed = [0; kyber768::KEY_GENERATION_SEED_SIZE];
+            rng.try_fill_bytes(&mut seed).map_err(|_| Error::KeyGen)?;
```
franziskuskiefer, line 93: We have the same in `encapsulate`. Maybe this should be moved out into a function.

src/kem.rs
```diff
@@ -93,9 +111,21 @@ pub fn encapsulate(
...
-            let gxy = ecdh::derive(alg.into(), pk, &new_sk)?;
+            let gxy = ecdh::derive(alg.try_into().unwrap(), pk, &new_sk)?;
             Ok((gxy, new_pk))
         }
+
+        Algorithm::Kyber768 => {
+            let mut seed = [0; kyber768::SHARED_SECRET_SIZE];
+            rng.try_fill_bytes(&mut seed).map_err(|_| Error::KeyGen)?;
+
+            if let Ok((ct, ss)) = kyber768::encapsulate(pk.try_into().unwrap(), seed) {
```
franziskuskiefer, line 122: Lets not unwrap the key conversions. They really could panic if the API isn't used correctly.
Same in `decapsulate`.

src/lib.rs
```diff
@@ -23,3 +23,4 @@ pub mod hmac;
 pub mod hpke;
 pub mod kem;
 pub mod signature;
+pub(crate) mod kyber768;
```
franziskuskiefer, line 26: This is not a top level module. Lets move into `kem`.

src/kem.rs
franziskuskiefer, line 1: We'll need another layer for Kyber to multiplex between the different implementations. But lets do that in a follow up (to use libjade on Linux/Mac x64).

franziskuskiefer (changes requested):
Thanks! Let's get this in quickly to start getting the improvements in.
Just a few small comments.
I assume that the Kyber code itself is copied from the spec. I didn't really review it.

franziskuskiefer (approved):
Thanks!

PR #37, opened 2023-08-08
Optimize kyber key generation

xvzcf:


src/kem/kyber768/sampling.rs
```diff
@@ -37,28 +36,35 @@ pub fn sample_from_uniform_distribution(
...
+
+        let coin_toss_outcomes = even_bits + odd_bits;
 
-        sampled[i] = coin_tosses_a - coin_tosses_b;
+        for outcome_set in (0..u32::BITS).step_by(4) {
+            let outcome_1: i16 = ((coin_toss_outcomes >> outcome_set) & 0x3)
+                .try_into()
+                .unwrap();
+            let outcome_2: i16 = ((coin_toss_outcomes >> (outcome_set + 2)) & 0x3)
+                .try_into()
```
franziskuskiefer, line 62: We could use `as i16` here. These things will need proofs for correctness anyway.

src/kem/kyber768/serialize.rs
```diff
@@ -27,6 +27,23 @@ pub fn serialize_little_endian(
...
+pub fn serialize_little_endian_12(re: KyberPolynomialRingElement) -> [u8; 384] {
+    let mut serialized = [0u8; 384];
+
+    for (i, chunks) in re.coefficients.chunks(2).enumerate() {
+        let coefficient = chunks[0].value;
+        let coefficient1 = chunks[1].value;
+
+        serialized[3 * i] = (coefficient & 0xFF).try_into().unwrap();
+        serialized[3 * i + 1] = ((coefficient >> 8) | ((coefficient1 & 0xF) << 4))
+            .try_into()
```
franziskuskiefer, line 39: Here we can use `as u8` as well for readability and it will need proofs.

franziskuskiefer (changes requested):
Nice improvements already.
Some comments and a few things that might get us another couple easy microseconds.

src/kem/kyber768/sampling.rs
xvzcf: Switched to using u16 with a new implementation of field element subtraction.

src/kem/kyber768/ntt.rs
```diff
@@ -39,26 +35,6 @@ pub(crate) mod kyber_polynomial_ring_element_mod {
 
     const NTT_LAYERS: [usize; 7] = [2, 4, 8, 16, 32, 64, 128];
 
-    /// Use the Cooley–Tukey butterfly to compute an in-place NTT representation
```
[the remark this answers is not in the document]
xvzcf: Just to make things easier while I break out code/reorder functions and all that. It's temporary, I'll add it back in later.

src/kem/kyber768/utils/field.rs
```diff
@@ -64,34 +63,39 @@ impl<const MODULUS: u16> From<u32> for PrimeFieldElement<MODULUS> {
...
-            i32::try_from(self.value).unwrap() - i32::try_from(other.value).unwrap();
-        let representative = difference.rem_euclid(MODULUS.into());
+        let lhs = self.value;
+        let rhs = MODULUS - other.value;
 
-        u16::try_from(representative).unwrap().into()
+        let sum: u16 = lhs + rhs;
+        let difference: u16 = sum.wrapping_sub(MODULUS);
+
+        let mask = 0u16.wrapping_sub((difference >> 15) & 1);
```
franziskuskiefer, line 91: llvm-14 is pretty good in optimizing the masks into conditionals. We should look at the assembly of this at some point.

franziskuskiefer (approved):
Thanks, let's get this in

PR #38, opened 2023-08-09
Reworking field operations.

xvzcf:
I added some code for barrett reduction, and tried to remove spurious modulo operations whereever I could.

### DIFF
--- a/src/kem/kyber768/field_element.rs
+++ b/src/kem/kyber768/field_element.rs
@@ -15,10 +15,10 @@ impl KyberFieldElement {
 
     pub fn barrett_reduce(value : u32) -> Self {
         let product : u64 = u64::from(value) * u64::from(Self::BARRETT_MULTIPLIER);
-        let quotient : u32 = (product >> Self::BARRETT_SHIFT).try_into().unwrap();
+        let quotient : u32 = (product >> Self::BARRETT_SHIFT) as u32;
 
         let remainder = value - (quotient * u32::from(Self::MODULUS));
-        let remainder : u16 = remainder.try_into().unwrap();
+        let remainder : u16 = remainder as u16;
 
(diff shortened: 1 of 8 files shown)


src/kem/kyber768/field_element.rs
```diff
@@ -0,0 +1,111 @@
...
+
+impl KyberFieldElement {
+    pub const MODULUS: u16 = FIELD_MODULUS;
+
+    const BARRETT_SHIFT : u32 = 24; // 2 * ceil(log_2(FIELD_MODULUS))
+    const BARRETT_MULTIPLIER : u32 = (1u32 << Self::BARRETT_SHIFT) / (Self::MODULUS as u32);
+
+    pub fn barrett_reduce(value : u32) -> Self {
+        let product : u64 = u64::from(value) * u64::from(Self::BARRETT_MULTIPLIER);
+        let quotient : u32 = (product >> Self::BARRETT_SHIFT).try_into().unwrap();
```
franziskuskiefer, line 18: Using `as` everywhere here gives you a 25% performance improvement.
`unwrap` is pretty expensive.

src/kem/kyber768/field_element.rs
```diff
@@ -0,0 +1,111 @@
...
+        Self {
+            value: (mask & sum) | (!mask & difference),
+        }
+    }
+}
+
+impl ops::Mul for KyberFieldElement {
+    type Output = Self;
+
+    fn mul(self, other: Self) -> Self {
```
franziskuskiefer, line 96: Multiplication here is pretty expensive compared to the `i16` based montgomery reduce used in pqclean.
Did you try that approach?

franziskuskiefer (changes requested):
Thanks.
Two comments. One to address in this PR and the other one we can discuss later and do in the next  one if we decide to do it.

franziskuskiefer (approved):

PR #43, opened 2023-08-18
Use signed representatives and lazy reduction.

xvzcf:


franziskuskiefer (approved):
Thanks, getting closer 🙂🎉

PR #49, opened 2023-08-22
Use montgomery multiplication everywhere.

xvzcf:


franziskuskiefer (approved):
Thanks, looks good!

On my machine I see the following
- Key generation is ~7% slower (32µs vs 30.5µs)
- encaps is significantly slower (68µs vs 38µs)
- decaps is significantly faster 🎉 (35µs vs 43µs)

PR #50, opened 2023-08-22
When encapsulating, only call key_gen() for EC algorithms

xvzcf:


examples/kyber768_encapsulate.rs
franziskuskiefer, line 1: Please run rustfmt on all the code.

franziskuskiefer (approved):
Good catch! Looks like we're faster than pqclean now 🥳
You only need to run rustfmt, then you can merge it.

franziskuskiefer (approved):

PR #51, opened 2023-08-23
Import sys/libjade/build.rs from franziskus/hacl-sys branch

mamonet:
The patch uses `cc` library to compile libjade instead of clang commands and create `libcrux_platform` package to detect cpu features.

### DIFF
--- a/sys/libjade/Cargo.toml
+++ b/sys/libjade/Cargo.toml
@@ -13,6 +13,7 @@ log = "0.4"
 libc = { version = "0.2", default-features = false }
 fs_extra = "1.2"
 cc = { version = "1.0", features = ["parallel"] }
+libcrux_platform = { path = "../platform" }
 
 [target.'cfg(not(windows))'.build-dependencies]
 bindgen = "0.66"
--- a/sys/libjade/Cargo.toml
+++ b/sys/libjade/Cargo.toml
@@ -12,6 +12,7 @@ log = "0.4"
 [build-dependencies]
 libc = { version = "0.2", default-features = false }
 fs_extra = "1.2"
+cc = { version = "1.0", features = ["parallel"] }
 
 [target.'cfg(not(windows))'.build-dependencies]
 bindgen = "0.66"
--- a/sys/libjade/Cargo.toml
+++ b/sys/libjade/Cargo.toml
@@ -13,7 +13,7 @@ log = "0.4"
 libc = { version = "0.2", default-features = false }
 fs_extra = "1.2"
 cc = { version = "1.0", features = ["parallel"] }
-libcrux_platform = { path = "../platform" }
+libcrux_platform = { version = "=0.0.1", path = "../platform" }
 
 [target.'cfg(not(windows))'.build-dependencies]
 bindgen = "0.66"
--- a/sys/libjade/build.rs
+++ b/sys/libjade/build.rs
@@ -193,9 +193,11 @@ pub fn main() -> Result<(), u8> {
     let mode = "static";
     println!("cargo:rustc-link-lib={}={}", mode, library_name);
     if platform.simd128 {
+        println!("cargo:rustc-cfg=simd128");
         println!("cargo:rustc-link-lib={}={}", mode, "jade_128");
     }
     if platform.simd256 {
+        println!("cargo:rustc-cfg=simd256");
         println!("cargo:rustc-link-lib={}={}", mode, "jade_256");
     }
     println!("cargo:rustc-link-search=native={}", out_path.display());
--- a/sys/libjade/build.rs
+++ b/sys/libjade/build.rs
@@ -168,8 +168,8 @@ pub fn main() -> Result<(), u8> {
         }
     } else {
         Platform {
-            simd128: std::arch::is_x86_feature_detected!("avx"),
-            simd256: std::arch::is_x86_feature_detected!("avx2"),
+            simd128: libcrux_platform::simd128_support(),
+            simd256: libcrux_platform::simd256_support(),
         }
     };
 
--- a/sys/libjade/build.rs
+++ b/sys/libjade/build.rs
@@ -1,211 +1,184 @@
-#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-mod x64_build {
-    use std::{path::Path, process::Command};
+use std::{env, path::Path};
 
-    macro_rules! svec {
+macro_rules! svec {
         ($($x:expr),*$(,)?) => (vec![$($x.to_string()),*]);
     }
 
-    pub fn copy_files(home_path: &Path, out_path: &Path) {
-        let mut options = fs_extra::dir::CopyOptions::new();
-        options.overwrite = true;
-        fs_extra::dir::copy(home_path.join("jazz"), out_path, &options).unwrap();
-    }
+fn copy_files(home_path: &Path, out_path: &Path) {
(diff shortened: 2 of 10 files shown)


sys/libjade/build.rs
```diff
@@ -1,211 +1,184 @@
...
+
+    // If cross compiling, we assume to have it all.
+    let platform = if cross_target.is_some() {
+        Platform {
+            simd128: true,
+            simd256: true,
+        }
+    } else {
+        Platform {
+            simd128: std::arch::is_x86_feature_detected!("avx"),
```
franziskuskiefer, line 171: This doesn't work.
The build script is built for the host, not the target platform. So this throws an error when cross compiling.
You need to guard this with the host platform.

franziskuskiefer (changes requested):
Thanks, looks good except for one issue.

sys/libjade/Cargo.toml
```diff
@@ -12,6 +12,8 @@ log = "0.4"
 [build-dependencies]
 libc = { version = "0.2", default-features = false }
 fs_extra = "1.2"
+cc = { version = "1.0", features = ["parallel"] }
+libcrux_platform = { path = "../platform" }
```
franziskuskiefer, line 16: Please fix this to the version number of the platform crate (`=0.0.1`)

sys/platform/src/lib.rs
```diff
@@ -0,0 +1,199 @@
+#![no_std]
```
franziskuskiefer, line 1: Please add some doc comments to the crate to describe what it's doing.
Same for each module.

sys/platform/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    cpu_id::supported(Feature::bmi2) && cpu_id::supported(Feature::adx)
+}
+
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+pub fn bmi2_adx_support() -> bool {
+    false
+}
+
+/// Check whether p(cl)mull is supported
+pub fn pmull_support() -> bool {
```
franziskuskiefer, line 82: There are three different approaches in here for the guards
1. multiple functions, each guarded with the platform
2. one function with guards within the function
3. a module with the guards and all the functions

Each of the approaches has advantages and disadvantages. Please pick one and make all of them use the same approach.

franziskuskiefer (changes requested):
Thanks, looks pretty good already. A few points inline.

Please run rustfmt on the code.

sys/platform/src/test.rs
```diff
@@ -0,0 +1,84 @@
...
+    eprintln!("aes\t\t{:?}", aes_ni_support());
+    eprintln!("advSimd\t\t{:?}", adv_simd_support());
+    eprintln!("pmull\t\t{:?}", pmull_support());
+    eprintln!("sha256\t\t{:?}", sha256_support());
+}
+
+// XXX[windows]: Something is running into a STATUS_ACCESS_VIOLATION here.
+#[cfg(all(
+    any(target_arch = "x86", target_arch = "x86_64"),
+    not(target_os = "windows")
```
franziskuskiefer, line 18: I don't think we need this guard. It works for me on Windows.

sys/platform/src/test.rs
```diff
@@ -0,0 +1,84 @@
...
+    eprintln!("avx512er\t{:?}", supported(Feature::avx512er));
+    eprintln!("avx512cd\t{:?}", supported(Feature::avx512cd));
+    eprintln!("sha\t\t{:?}", supported(Feature::sha));
+    eprintln!("avx512bw\t{:?}", supported(Feature::avx512bw));
+    eprintln!("avx512vl\t{:?}", supported(Feature::avx512vl));
+}
+
+#[cfg(all(
+    any(target_arch = "x86", target_arch = "x86_64"),
+    not(target_os = "windows")
```
franziskuskiefer, line 58: Same for this one. This works for me on Windows as well.

sys/platform/src/test.rs
```diff
@@ -0,0 +1,84 @@
...
+))]
+#[test]
+fn cpuid() {
+    use super::x86::supported;
+    use std::time::Instant;
+
+    let now = Instant::now();
+    let _avx2 = supported(Feature::avx2);
+    let elapsed = now.elapsed();
+    eprintln!("libcrux init: {elapsed:.2?}");
```
franziskuskiefer, line 68: What do these timings tell you? Is our feature detection fast?

sys/libjade/Cargo.toml
mamonet: Done!

sys/platform/src/lib.rs
mamonet: Accomplished with approach 2 for all functions.

sys/libjade/build.rs
mamonet: Replaced with functions from platform package, should be fixed now.

sys/platform/src/lib.rs
mamonet: I added header doc for each src module in platform crate. Let me know if there's room for improvement.

sys/platform/src/test.rs
mamonet: Done!

sys/platform/src/test.rs
mamonet: Done!

sys/platform/src/test.rs
mamonet: when I run `cargo test`, I don't get input from this specific module. Would appreciate any hints to get the output here.

sys/platform/src/linux_arm.rs
```diff
@@ -0,0 +1,55 @@
+/* Obtain particular CPU features for AArch64 on Linux */
+
+use libc::{getauxval, AT_HWCAP, HWCAP_AES, HWCAP_ASIMD, HWCAP_PMULL, HWCAP_SHA2};
+
+#[inline(always)]
+fn auxval() {
+    let val = unsafe { getauxval(AT_HWCAP) };
+    unsafe { ADV_SIMD = val & HWCAP_ASIMD != 0 };
```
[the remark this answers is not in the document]
mamonet: Right. Edited to tighten unsafe circle.

sys/platform/src/test.rs
```diff
@@ -0,0 +1,77 @@
...
+    eprintln!("x25519\t\t{:?}", x25519_support());
+    eprintln!("bmi2 & adx\t{:?}", bmi2_adx_support());
+    eprintln!("aes\t\t{:?}", aes_ni_support());
+    eprintln!("advSimd\t\t{:?}", adv_simd_support());
+    eprintln!("pmull\t\t{:?}", pmull_support());
+    eprintln!("sha256\t\t{:?}", sha256_support());
+}
+
+#[test]
+fn dump_raw() {
```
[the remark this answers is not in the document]
mamonet: I dropped `x86|x64` guard by unconscious editing. Brought it back now.

sys/platform/src/test.rs
mamonet: Thanks! On x86_64, libcrux init function is ~2x faster than `std` one while the actual detection procedure is 2x slower but it's a matter of 100ns, doesn't make any difference.

franziskuskiefer (approved):
Thanks! Let's get this in.

PR #52, opened 2023-08-24
More cleanup

xvzcf:
- Refactored test harness and added a check to see that the right implicit rejection secret was being calculated
- If re-encryption succeeds during decryption, compare the ciphertexts and choose the value to be hashed in constant time
- Added barrett reduction before returning from the inverse ntt to try and avoid overflows
- Renamed `utils` to `conversions` and removed some conversion related traits and methods


src/kem/kyber768/constant_time_ops.rs
```diff
@@ -0,0 +1,41 @@
+use crate::kem::kyber768::{CIPHERTEXT_SIZE, SHARED_SECRET_SIZE};
+
+// TODO: Attribute this to RustCrypto
+#[inline]
+fn is_non_zero(value: u8) -> u8 {
+    const SHIFT_BITS: usize = core::mem::size_of::<u64>() - 1;
+    let value = value as u64;
```
franziskuskiefer, line 7: Can we use `u32` here? `u64` seems overkill and might not be great on 32-bit systems.

src/kem/kyber768/constant_time_ops.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    (((value | (!value).wrapping_add(1)) >> SHIFT_BITS) & 1) as u8
+}
+
+pub(crate) fn compare_ciphertexts_in_constant_time(lhs: &[u8], rhs: &[u8]) -> u8 {
```
franziskuskiefer, line 12: Let's make sure this is not optimized out. From what I see, it doesn't look good. https://godbolt.org/z/obsMY3YYM
I'm fine to leave this in and replace then later with something actually constant time.

src/kem/kyber768.rs
```diff
@@ -55,15 +59,18 @@ pub fn encapsulate(
 ) -> Result<(Kyber768Ciphertext, Kyber768SharedSecret), BadRejectionSamplingRandomnessError> {
     let randomness_hashed = H(&randomness);
 
-    let to_hash: [u8; 2 * H_DIGEST_SIZE] = randomness_hashed.push(&H(&public_key));
+    let mut to_hash: [u8; 2 * H_DIGEST_SIZE] = randomness_hashed.as_ref().into_padded_array();
```
franziskuskiefer, line 62: Do these changes make a performance difference? Or what is the motivation for the change?

franziskuskiefer (changes requested):
Thanks. Generally looks good to me. Just a few questions.

src/kem/kyber768/constant_time_ops.rs
xvzcf: Switched to a simpler expression for now, we can complicate it depending on what LLVM outputs.

src/kem/kyber768/constant_time_ops.rs
xvzcf: Added a comment about this.

src/kem/kyber768.rs
xvzcf: No I don't believe they do,  I found the `UpdatingArray2` traits and methods a bit confusing and I decided to remove them to keep just one. I might've been a little too quick to do so.

franziskuskiefer: Simplifying the code is a good reason as well 🙂

franziskuskiefer (approved):

PR #53, opened 2023-08-28
Utilize platform crate in src files

mamonet:
This patch removes old hardware detection module and utilizes platform package in `src` files to detect CPU features.


build.rs
```diff
@@ -1,57 +1,9 @@
...
-        && std::arch::is_x86_feature_detected!("sse")
-        && std::arch::is_x86_feature_detected!("aes")
-        && std::arch::is_x86_feature_detected!("pclmulqdq")
-}
-
-#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
-fn aes_ni_support() -> bool {
-    false
-}
+use libcrux_platform::{
```
franziskuskiefer, line 1: Just a note: We'll want to remove all this once hacl is moved over to the new build.

franziskuskiefer (commented):
Thanks. Generally looks good to me. I'll need to look at it again when #51 is merged. But I don't see anything here that would need chnages.

The hacl module is unused for now. So the changes don't make any difference. But they also don't hurt.

franziskuskiefer:
@mamonet can you update this branch to main and resolve the conflicts?

franziskuskiefer (approved):
Thanks

PR #57, opened 2023-08-29
Import sys/hacl/build.rs from franziskus/hacl-sys branch

mamonet:
The patch uses cc library to compile `hacl` package instead of clang commands and utilizes `platform` crate to detect cpu features. It also updates Hacl C files and create `config.h`, `hacl.h`, `hacl128.h`, and `hacl256.h` headers to help building the package.


sys/hacl/build.rs
```diff
@@ -1,19 +1,24 @@
 #[cfg(not(windows))]
 extern crate bindgen;
```
franziskuskiefer, line 3: This shouldn't be necessary.

sys/hacl/Cargo.toml
```diff
@@ -10,6 +10,11 @@ edition = "2021"
 [build-dependencies]
 libc = { version = "0.2", default-features = false }
 fs_extra = "1.2"
```
franziskuskiefer, line 12: This needs to be bumped to `1.3`.

sys/hacl/build.rs
```diff
@@ -190,6 +267,18 @@ fn main() {
...
+        simd256: libcrux_platform::simd256_support(),
+        aes_ni: libcrux_platform::aes_ni_support(),
+        x25519: libcrux_platform::x25519_support(),
+        bmi2_adx_support: libcrux_platform::bmi2_adx_support(),
+        pmull: libcrux_platform::pmull_support(),
+        adv_simd: libcrux_platform::adv_simd_support(),
+        sha256: libcrux_platform::sha256_support(),
+    };
+
     // Set re-run trigger for all of c
```
franziskuskiefer, line 282: For some reason we always rebuild (I think because of the copies, ideally we get rid of them).
But to avoid rebuilding every time. Let's remove these two lines and add `println!("cargo:rerun-if-changed=build.rs");` instead.
Alternatively, you could try to build without copying the source files to see if that solves the issue.

sys/hacl/build.rs
```diff
@@ -199,17 +288,23 @@ fn main() {
     eprintln!(" >>> out {:?}", out_path);
 
     // Build the C files
-    build(out_path);
+    build(platform, out_path);
 
     // Set library name to look up
     let library_name = "hacl";
```
franziskuskiefer, line 294: Let's move this out into a `const` together with the other library names for 128 and 256 etc.

sys/hacl/build.rs
```diff
@@ -95,92 +117,147 @@ fn copy_files(home_path: &Path, out_path: &Path) {
...
+        // XXX: There are too many warnings for now
+        // .warnings_into_errors(true)
+        .warnings(true)
+        .extra_warnings(true);
+    // .no_default_flags(true);
+
+    for include in includes(out_path, "") {
+        build.include(include);
+    }
+    build.flag("-O3").flag("-c");
```
franziskuskiefer, line 141: I have to check this later but I expect that this will fail on Windows when compiling with MSVC.
Let's use the `opt_level` function instead. The `-c` we don't need at all I think.

sys/hacl/build.rs
```diff
@@ -25,34 +30,42 @@ fn includes(home_dir: &Path) -> Vec<String> {
...
-    }
+    flags.push("-mavx2".to_string());
 }
 
 #[cfg(not(windows))]
-fn create_bindings(home_dir: &Path) {
-    let mut clang_args = includes(home_dir);
-    append_simd128_flags(&mut clang_args);
-    append_simd256_flags(&mut clang_args);
+fn create_bindings(platform: Platform, home_dir: &Path) {
```
franziskuskiefer, line 45: The Windows function has a different function signature now and will fail.
Ideally we don't need this guard, but I think Windows has some issues here.
And since we want to move away from creating bindings every time, let's not try fix this for now. But it should build on Windows too.

sys/hacl/build.rs
```diff
@@ -25,34 +30,42 @@ fn includes(home_dir: &Path) -> Vec<String> {
...
-    // Platform detection
-    if cfg!(simd128) {
-        flags.push("-DSIMD128".to_string());
-        if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
-            flags.push("-msse4.1".to_string());
-            flags.push("-msse4.2".to_string());
-            flags.push("-mavx".to_string());
-        }
+    if cfg!(any(target_arch = "x86", target_arch = "x86_64")) {
+        flags.push("-msse4.1".to_string());
```
franziskuskiefer, line 34: These flags probably won't work on Windows. We need to make this msvc compatible.

sys/hacl/build.rs
```diff
@@ -25,34 +30,42 @@ fn includes(home_dir: &Path) -> Vec<String> {
...
-    let bindings = bindgen::Builder::default()
+    let mut bindings = bindgen::Builder::default();
+    bindings = bindings
         // Header to wrap HACL headers
-        .header("c/include/hacl.h")
+        .header("c/include/hacl.h");
+
+    if platform.simd128 {
+        append_simd128_flags(&mut clang_args);
+        clang_args.push("-DSIMD128".to_string());
```
franziskuskiefer, line 55: These are still used in the hand-written `config.h`. But we should probably set the right defines here instead.

sys/hacl/build.rs
```diff
@@ -25,34 +30,42 @@ fn includes(home_dir: &Path) -> Vec<String> {
...
+    let mut clang_args = includes(home_dir, "-I");
 
-    let bindings = bindgen::Builder::default()
+    let mut bindings = bindgen::Builder::default();
+    bindings = bindings
         // Header to wrap HACL headers
-        .header("c/include/hacl.h")
+        .header("c/include/hacl.h");
+
+    if platform.simd128 {
```
franziskuskiefer, line 53: There are some more platform dependent defines we should set.
The following values are usually set in `config.h`. We need to set them depending on the platform we are building for.
But let's file a follow up issue to tackle it in another PR and get a first version in that works.

- `TARGET_ARCHITECTURE`
- `HACL_CAN_COMPILE_UINT128`
- `HACL_CAN_COMPILE_VALE`
- `LINUX_NO_EXPLICIT_BZERO`

franziskuskiefer (changes requested):
Thanks, some comments below and inline.

I'm seeing lots of no upper case warnings. Please add something to the crate that allows them to avoid the warning.

I didn't look any further at the `c` directory assuming that it's imported code.

Please add a CI job that is building this hacl crate. Right now it's not used and hence, not even built. So we don't know if it's working.

We should move the hand written header files into a different directory to avoid clobbering them when importing code from HACL. How about putting them into c/ directly?

sys/hacl/Cargo.toml
mamonet: Edited!

sys/hacl/build.rs
mamonet: Fixed!

sys/hacl/build.rs
mamonet: Edited!

sys/hacl/build.rs
mamonet: Agreed. We can make `create_bindings` work on Windows at later point.

sys/hacl/build.rs
mamonet: msvc uses to ignore such flags, I used a proper guard to push these flags where necessary.

sys/hacl/build.rs
mamonet: Here is what happens when we don't add `-DSIMD128/-DSIMD256` flags
libhacl128 built first and so it includes `libintvector.h` with only `HACL_CAN_COMPILE_VEC128` defined, see `hacl128.h` and `config.h`. when libhacl256 is building, it has `libintvector.h` with no `HACL_CAN_COMPILE_VEC256` defined and the build fails.

sys/hacl/build.rs
mamonet: I agree.

sys/hacl/build.rs
mamonet: I switched `library_name` to `const`, let me know if I missed something here.

sys/hacl/build.rs
mamonet: Done~

sys/hacl/c/config/hacl128.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * Meta header to include all the main header files we care about.
```
franziskuskiefer, line 2:
```suggestion
 * Meta header to include all the header files for SIMD 128 we care about.
```

sys/hacl/c/config/hacl256.h
```diff
@@ -0,0 +1,10 @@
+/*
+ * Meta header to include all the main header files we care about.
```
franziskuskiefer, line 2:
```suggestion
 * Meta header to include all the header files for SIMD 256 we care about.
```

sys/hacl/c/config/vale-aes.h
```diff
@@ -0,0 +1,8 @@
+/*
+ * Meta header to include all the main header files we care about.
+ */
+#pragma once
+
+#include "config.h"
+
+#include "EverCrypt_AEAD.h"
```
franziskuskiefer, line 8:
```suggestion
#include "EverCrypt_AEAD.h"

```

sys/libjade/build.rs
```diff
@@ -66,10 +60,10 @@ fn create_bindings(platform: Platform, home_dir: &Path) {
 }
 
 #[cfg(windows)]
-fn create_bindings(platform: Platform, _: &Path) {}
+fn create_bindings(_: Platform, _: &Path) {}
```
franziskuskiefer, line 63: We can remove this completely as libjade isn't built for Windows right now.

sys/hacl/build.rs
franziskuskiefer: Ok, I still think we can simplify this. But let's not do it right now if it's working as is.

sys/hacl/build.rs
franziskuskiefer: MSVC builds all this without any flags? It looks like it's building but I'm a little surprised. But let's get it in like this. We'll see if it works as expected once we use it.

sys/hacl/build.rs
franziskuskiefer: Can you file an issue for it?

sys/hacl/build.rs
```diff
@@ -24,35 +30,44 @@ fn includes(home_dir: &Path) -> Vec<String> {
...
         flags.push("-mavx2".to_string());
     }
 }
 
-#[cfg(not(windows))]
-fn create_bindings(home_dir: &Path) {
-    let mut clang_args = includes(home_dir);
-    append_simd128_flags(&mut clang_args);
-    append_simd256_flags(&mut clang_args);
+fn create_bindings(platform: Platform, home_dir: &Path) {
```
franziskuskiefer, line 47: Let's add a note to the readme that in order to build this the libclang path needs to be set correclty.

sys/hacl/build.rs
mamonet: Yes, I made sure that simd128/256 code are built on MSVC.

sys/hacl/build.rs
mamonet: Maybe I misunderstood the comment, but I see Cargo compile `clang-sys` to use it for `bindgen`

sys/hacl/build.rs
mamonet: Filed! https://github.com/cryspen/libcrux/issues/60

sys/hacl/build.rs
franziskuskiefer: I filed #61 as a follow up for this.

franziskuskiefer (approved):
Thanks, let's get this in.

Issue #54, opened 2023-08-28
Update CI


franziskuskiefer:
Update the CI

- [ ] add merge queues
- [x] add rust fmt checks
- [ ] ~~add individual jobs for the sys libraries.~~

Platform support
- [x] macOS x64
- [x] macOS aarch64 (builds only)
- [ ] iOS aarch64 (builds only)
- [ ] Windows x86
- [x] Windows x64
- [x] Linux x64
- [x] Linux x86
- [ ] Linux aarch64
- [ ] Android aarch64
- [x] wasm-32

Ergonomics & Maintainability
- [ ] Pull out common things like setup into reusable workflows

There are some CI improvements esp. for 32-bit in #21.

franziskuskiefer:
- merge queues is being worked on in #525 
- we should add the remaining architectures to CI for Rust only code

franziskuskiefer:
Closing this. Let's file more specific issues when we want to extend the CI.

PR #58, opened 2023-08-29
Only Enable aes-ni configuration on x86_64/x86

mamonet:
Some functions in `aead.rs` don't support `aarch64` yet, this patch only enables `aes-ni` configuration on `x86_64/x86` architectures.


franziskuskiefer (approved):
Thanks! We should tackle #54 soon to avoid this.
I'll merge it when the CI is happy.

PR #62, opened 2023-08-31
Compile Vale AESGCM in hacl

mamonet:
This patch imports Vale assembly files of AES-GCM mode and supports building these files in Hacl.


franziskuskiefer (changes requested):
Thanks, just two nits for warnings on other machines.

franziskuskiefer (approved):
Thanks

PR #63, opened 2023-09-01
Use libcrux-hacl crate in src files

mamonet:
This patch replaces the dependency of `hacl [hazmat]` with `libcrux-hacl` by utilizing the latter in `src` files and creating wrappers for certain algorithms.
It also imports Evercrypt modules and Vale assembly files of `aesgcm` and `cpuid` from hacl distributions and compiles it in the local hacl build.


franziskuskiefer (changes requested):
Cancelling until the CI is green.

Cargo.toml
```diff
@@ -17,7 +17,7 @@ crate-type = ["staticlib", "cdylib", "lib"]
 libcrux_platform = { version = "=0.0.1", path = "sys/platform" }
 
 [dependencies]
-hacl = { version = "=0.0.2", features = ["hazmat"] }
+libcrux-hacl = { version = "=0.0.1", path = "sys/hacl" }
```
franziskuskiefer, line 20: So the consistent naming part is still open.

sys/hacl/c/config/vale-aes.h
```diff
@@ -5,4 +5,5 @@
 
 #include "config.h"
 
-#include "Vale.h"
+#include "EverCrypt_AutoConfig2.h"
```
franziskuskiefer, line 8: How much work would it be to sidestep evercrypt for AES GCM?
That's the only one that doesn't have a HACL API.

sys/hacl/c/src/msvc/Hacl_Curve25519_64.c
```diff
@@ -27,9 +27,7 @@
 
 #include "internal/Vale.h"
 #include "internal/Hacl_Krmllib.h"
```
franziskuskiefer, line 29: Is this a manual change?

sys/hacl/c/src/Hacl_Curve25519_64.c
```diff
@@ -27,9 +27,7 @@
 
 #include "internal/Vale.h"
 #include "internal/Hacl_Krmllib.h"
```
franziskuskiefer, line 29: Same here. A manual change?

src/hacl.rs
```diff
@@ -1,7 +1,62 @@
...
+pub(crate) mod ed25519;
+pub(crate) mod hkdf;
+pub(crate) mod hmac;
+pub(crate) mod p256;
+pub(crate) mod sha2;
+pub(crate) mod sha3;
+
+/// Unified error type.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Error {
```
franziskuskiefer, line 26: Do we need this error type or shall would it make more sense to pull this up into the main error type of libcrux?
This was necessary in the `hazmat` case. But I think it might be nicer if we skip it here.

src/hacl/hkdf.rs
```diff
@@ -0,0 +1,150 @@
...
+                    }
+                    Ok(okm)
+                }
+
+                /// HKDF using the `salt`, input key material `ikm`, `info`. The output length
+                /// is defined through the result type.
+                /// Calls `extract` and `expand` with the given input.
+                ///
+                /// Returns the key material in an array of length `okm_len`.
+                pub fn hkdf(
```
franziskuskiefer, line 117: This function is unused because it's implemented on the top level. Lets remove it.

src/hacl/hkdf.rs
```diff
@@ -0,0 +1,150 @@
...
+                }
+                Ok(okm)
+            }
+
+            /// HKDF using the `salt`, input key material `ikm`, `info`. The output length
+            /// is defined through the result type.
+            /// Calls `extract` and `expand` with the given input.
+            ///
+            /// Returns the key material in an array of length `okm_len`.
+            pub fn hkdf<const OKM_LEN: usize>(
```
franziskuskiefer, line 70: This one is unused as well.
We don't have this one on the top level. Let's add it there.

src/hacl/hkdf.rs
```diff
@@ -0,0 +1,150 @@
...
+            }
+
+            /// HKDF expand using the pre-key material `prk` and `info`. The output length
+            /// is defined through the result type.
+            /// Returns the key material in an array of length `okm_len` or
+            /// [`Error::OkmTooLarge`] if the requested `okm_len` is too large.
+            ///
+            /// Note that this function returns an [`Error::ArgumentsTooLarge`]
+            /// if `salt`, `ikm`, or `OKM_LEN` is larger than 2**32 bytes.
+            pub fn expand<const OKM_LEN: usize>(
```
franziskuskiefer, line 41: The top level should probably expose something like this.
With all the warnings in here. We should make sure the top level APIs are done well and offer these different options, and then drop the functions here that are not necessary.
You can do that here, or file a follow-up.

src/aead.rs
```diff
@@ -321,9 +321,9 @@ fn decrypt_32(
 fn aes_encrypt_128(key: &Aes128Key, msg_ctxt: &mut [u8], iv: Iv, aad: &[u8]) -> Result<Tag, Error> {
     // XXX: Note that this might still fail because we don't check all
```
franziskuskiefer, line 322: Aren't these comments outdated now?

franziskuskiefer (changes requested):
Thanks!
I left some comments. They are all about improving the API and making things less complex
But I'd be fine with pushing that into follow-ups.
I only want to make sure we don't have any warnings before merging. You can just surpress them for now and file follow-ups if you prefer so that we can get this landed today.


- We have a warning in the build.rs that shold get fixed. `aes_ni_support` is only used on some architectures.

sys/hacl/c/config/vale-aes.h
mamonet: I hate using Evercrypt as much as you, let's keep it on TODO list and get back to it when we have time for.

src/hacl/hkdf.rs
mamonet: Removed!

src/hacl/hkdf.rs
mamonet: I allowed `dead_code` at crate level to suppress relative warnings.

src/aead.rs
mamonet: Yes, removed now.

Cargo.toml
mamonet: Do you want to change hacl package naming?

src/hacl.rs
mamonet: This `enum` is used in `signature.rs` and `ecdh`. I don't see where the main error type of libcrux.

src/hacl/hkdf.rs
mamonet: Yeah. Let's keep this PR for hack package utilization and file a follow-up for other changes.

src/hacl.rs
franziskuskiefer: #65

sys/hacl/c/src/msvc/Hacl_Curve25519_64.c
franziskuskiefer: There's no need to touch these files.
This makes reviewing things unnecessarily hard.
I'll update all the C files when this PR is in to make sure we have the correct code.
`config.h` is used in hacl packages and everywhere else.

sys/hacl/c/config/vale-aes.h
franziskuskiefer: #66

sys/hacl/c/src/msvc/Hacl_Curve25519_64.c
franziskuskiefer: #67

Issue #47, opened 2023-08-22
build hacl-sys directly instead of using hacl-packages


franziskuskiefer:

mamonet:
Implemented in https://github.com/cryspen/libcrux/pull/63

franziskuskiefer:
Reopening this as the build isn't working properly yet #69

franziskuskiefer:
Fixed in #69

PR #69, opened 2023-09-04
Misc cleanup and updated hacl bindings

franziskuskiefer:
Follow up to #47 updating the HACL code.


franziskuskiefer:
@mamonet can you look at this? Looks like the build isn't working with the code from upstream.

PR #70, opened 2023-09-04
Automatically sync sha256 and sha3-224 from libjade upstream.

xvzcf:
I formatted the python script using [black](https://github.com/psf/black).


franziskuskiefer:
> Note that the SHA256 code in libjade is unverified, as far as I know, so merging that to dev is fine but we should be careful not to include or at least expose it in main.

Please look at the code first. This is not actually using any new libjade code in libcrux.

karthikbhargavan:
Added an issue: https://github.com/cryspen/libcrux/issues/71

PR #72, opened 2023-09-06
kyber extraction

franziskuskiefer:
Some small changes to make kyber extract to F*.
The F* extraction is in the proofs directory.

This needs https://github.com/hacspec/hacspec-v2/pull/246 to be merged for extraction.
Then the extraction can be done with `cargo hax into -i '-** +kem::kyber768::** -kem::kyber768::arithmetic::mutable_operations::**' fstar`.

### DIFF
--- a/proofs/fstar/extraction/Libcrux.Kem.Kyber768.Arithmetic.fst
+++ b/proofs/fstar/extraction/Libcrux.Kem.Kyber768.Arithmetic.fst
@@ -1,0 +1,145 @@
+module Libcrux.Kem.Kyber768.Arithmetic
+#set-options "--fuel 0 --ifuel 1 --z3rlimit 15"
+open Core
+
+let t_KyberFieldElement = i32
+
+let v_BARRETT_SHIFT: i32 = 26l
+
+let v_BARRETT_R: i32 = 1l >>. v_BARRETT_SHIFT
+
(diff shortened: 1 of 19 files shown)


proofs/fstar/extraction/Libcrux.Kem.Kyber768.Arithmetic.fst
```diff
@@ -0,0 +1,145 @@
+module Libcrux.Kem.Kyber768.Arithmetic
```
xvzcf, line 1: What is the difference between this and the `formal_verification` directory? Also should these files be under a `kyber768` sub-directory in `proofs`?

franziskuskiefer: There is no relation yet.
The `formal_verification` dir is a manual thing about all the other code. This is the first piece of code that will be proven here in libcrux directly. And we will have to figure out how to structure that.
There is work on hax (e.g. https://github.com/hacspec/hacspec-v2/pull/202) to work towards "verifiable crates" that define how to integrate the proofs and where this code lives.

For now we do what hax is doing, that is putting the extracted code into the `proofs` directory. Note that this is an extraction of the libcrux crate, not kyber.

Issue #42, opened 2023-08-17
Update build system


franziskuskiefer:
Update the build to

- [x] #46
- [x] #47
- [x] #48
- [x] add basic CI support to ensure all of this is working

Some open ToDos on branch franziskus/hacl-sys
- [x] consistent naming of sys crates (libjade-sys, libcrux-hacl-sys, libcrux-platform, ... ?)
- [x] consistent version numbers across crates
- [x] fix all crates to the right version (`"=0.0.1"`)
- [x] use new libcrux-hacl-sys functions in hacl module with hardware detection from libcrux-platform
- [x] use functions from hacl module in libcrux
- [x] #93

**Platform support**
- [x] macOS (arm64, x64)
- [x] Windows (x64, x86)
- [x] Linux (x64, x86)

Platform support for follow ups
- [x] #90
- [x] #91
- [x] wasm (#10)
- [ ] Windows (arm64)
- [x] #92

franziskuskiefer:
Open follow ups
- [ ] #60 
- [ ] #61 
- [ ] #54

PR #73, opened 2023-09-08
wasm

franziskuskiefer:
Add wasm support and CI for it.
This PR also adds CI for 32-bit Linux. (No Windows yet as that's not working for some reason.)
The PR further cleans up some of the sys builds.

This works towards #42


build.rs
```diff
@@ -1,16 +1,19 @@
...
     }
-    if libcrux_platform::simd256_support() {
+    if libcrux_platform::simd256_support() && target_arch == "x86_64" {
         println!("cargo:rustc-cfg=simd256");
     }
     if libcrux_platform::bmi2_adx_support() {
         println!("cargo:rustc-cfg=bmi2");
         println!("cargo:rustc-cfg=adx");
     }
-    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
mamonet, line 12: I think we can do further cleanup by removing checks of `x86` arch along with features like `aesni` and `avx` in `sys/hacl/build.rs`

sys/hacl/build.rs
```diff
@@ -1,5 +1,13 @@
 use std::{env, path::Path};
 
+const LIB_NAME: &str = "hacl";
+const LIB_128_NAME: &str = "hacl_128";
+const LIB_256_NAME: &str = "hacl_256";
+#[allow(dead_code)] // FIXME: remove wrong cfg guards
```
mamonet, line 6: You mean replacing cfg guards with conditional checks in `if` statements? You may want to remove checking for `x86` arch there too?

sys/hacl/build.rs
```diff
@@ -30,6 +38,7 @@ fn includes(home_dir: &Path, include_str: &str) -> Vec<String> {
     ]
 }
 
+// FIXME: drop all the host specific checks and move them to target.
 #[cfg(all(
     any(target_arch = "x86", target_arch = "x86_64"),
     any(
```
mamonet, line 44: Isn't `target_arch` meant to check for target arch?

sys/hacl/build.rs
```diff
@@ -171,6 +187,10 @@ fn compile_files(
         build.include(include);
     }
     build.opt_level(3);
+    build.static_crt(true);
+    if let Some(target) = platform.target {
+        build.flag(&format!("--target={target}"));
+    }
```
mamonet, line 193: I'm assuming this flag is to enable cross-compile. We may need to check for `msvc` to avoid using `--target` flag there.

sys/hacl/build.rs
```diff
@@ -370,71 +394,54 @@ struct Platform {
...
+    build(&platform, home_path);
 
     // Generate new bindings.
-    create_bindings(platform, home_path);
-
-    // Link hacl library.
-    const MODE: &str = "static";
-    println!("cargo:rustc-link-lib={}={}", MODE, LIB_NAME);
-    if platform.simd128 {
-        println!("cargo:rustc-link-lib={}={}", MODE, "hacl_128");
```
mamonet, line 410: By having the different lib naming, we no longer have to link the additional libraries manually?

sys/hacl/build.rs
```diff
@@ -370,71 +394,54 @@ struct Platform {
...
-        adv_simd: libcrux_platform::adv_simd_support(),
-        sha256: libcrux_platform::sha256_support(),
+    let platform = if target_arch != "wasm32" {
+        let x86 = target_arch == "x86";
+        Platform {
+            simd128: !x86 && libcrux_platform::simd128_support(),
+            simd256: !x86 && libcrux_platform::simd256_support(),
+            aes_ni: libcrux_platform::aes_ni_support(),
+            x25519: !x86 && libcrux_platform::x25519_support(),
+            bmi2_adx_support: libcrux_platform::bmi2_adx_support(),
```
mamonet, line 419: Both `x25519` and `bmi2_adx_support` check for same flags underneath. We don't use `bmi2_adx_support`, should we remove it?

sys/hacl/build.rs
```diff
@@ -370,71 +394,54 @@ struct Platform {
...
-        x25519: libcrux_platform::x25519_support(),
-        bmi2_adx_support: libcrux_platform::bmi2_adx_support(),
-        pmull: libcrux_platform::pmull_support(),
-        adv_simd: libcrux_platform::adv_simd_support(),
-        sha256: libcrux_platform::sha256_support(),
+    let platform = if target_arch != "wasm32" {
+        let x86 = target_arch == "x86";
+        Platform {
+            simd128: !x86 && libcrux_platform::simd128_support(),
+            simd256: !x86 && libcrux_platform::simd256_support(),
```
mamonet, line 416: `libcrux_platform` checks for `avx` in `simd128_support()` besides `avx2` in `simd256_support()` and both aren't supported in `x86` 32-bit

mamonet, line 416: `libcrux_platform` checks for `avx` in simd128_support() besides `avx2` in simd256_support(), AFIAK both aren't supported in x86 32-bit processors but can be enabled with `-m32` mode on x64 targets

sys/hacl/build.rs
franziskuskiefer: Actually this is done automatically. This is a leftover. I'll delete it.

sys/hacl/build.rs
franziskuskiefer: The cfg guards are wrong right now because they are for the host system, not the target. That's why we have dead code here.

sys/hacl/build.rs
franziskuskiefer: All these checks are compile time checks right now. But compile time of this build script, not the library. Hence this is wrong, because it's not for the target, but the host.

sys/hacl/build.rs
franziskuskiefer: Yeah, we might want to enable something on x86 at some point. But for now we just want it to work on 32-bit platforms.

sys/hacl/build.rs
franziskuskiefer: bmi2 and adx is used for some other things as well. So it's worth keeping independently.

sys/hacl/build.rs
```diff
@@ -38,55 +38,49 @@ fn includes(home_dir: &Path, include_str: &str) -> Vec<String> {
...
     }
 }
 
-fn append_simd128_flags(flags: &mut Vec<String>, is_bindgen: bool) {
-    if cfg!(any(target_arch = "x86", target_arch = "x86_64"))
-        && (is_bindgen || cfg!(not(target_env = "msvc")))
+#[allow(dead_code)]
+fn append_simd128_flags(platform: &Platform, flags: &mut Vec<String>, is_bindgen: bool) {
+    if platform.target_arch == "x86"
+        || platform.target_arch == "x86_64" && (is_bindgen || platform.target_env != "msvc")
```
mamonet, line 52: `&&` operator has precedence over `||` in Rust, I think we should set `platform.target_arch == "x86" || platform.target_arch == "x86_64"` in parentheses.

sys/hacl/build.rs
franziskuskiefer: Oh looks like the comment didn't make it here.
This is unrelated to names. `cc` adds this boilerplate if not disabled.

PR #76, opened 2023-09-13
Use google benchmark to benchmark BoringSSL code.

xvzcf:


franziskuskiefer (approved):
Thanks, this looks way nicer

PR #77, opened 2023-09-13
Added CIRCL and BoringSSL shake benchmarks.

xvzcf:


benches/boringssl/shake.cxx
```diff
@@ -0,0 +1,33 @@
+#include <stdlib.h>
+#include <string>
+
+#include "crypto/kyber/internal.h"
+
+#include <benchmark/benchmark.h>
+
+#define BYTES_TO_OUTPUT 10000
```
franziskuskiefer, line 8: The interesting values here are probably 128 and 840.

franziskuskiefer (approved):
Thanks. Looks like keccak is the bottleneck again.

The error we see on CI here is probably the same one you encountered recently.
The Linux image got updated two days ago and apparently there's an LLVM issue (version mismatch?) with bindgen.

We need to get this fixed before merging.

PR #78, opened 2023-09-15
Begin updating spec to FIPS 203 draft.

xvzcf:


specs/kyber/src/sampling.rs
```diff
@@ -100,60 +111,60 @@ pub fn sample_from_uniform_distribution(
...
+    let mut f: KyberPolynomialRingElement = KyberPolynomialRingElement::ZERO;
+
+    for i in 0..f.len() {
         let mut coin_tosses: u8 = 0;
-        for bit in random_bits
+
+        // x ← ∑(j = 0 to η-1) b[2iη + j]
+        for bit in bits
             .next()
             .expect("the assertion ensures there are enough sampling coins")
```
franziskuskiefer, line 146: I know we looked at these unwrap/expect things before.
But looking at this again I think this makes things a little hard to read.
How about adding a comment and using an unwrap instead? Or even better, hiding the unwrap completely?

Looking at the entire function, we should probably factor out the inner for loops over the `bits` to only have a `sum` function call. Then this looks like the spec.

specs/kyber/src/sampling.rs
```diff
@@ -5,67 +5,78 @@ use crate::{
...
-///     if d_1 < q then
-///         aˆ_{j} := d_1
-///         j := j + 1
+/// Input: byte stream B ∈ B*
+/// Output: array â ∈ ℤ₂₅₆
+///
+/// i ← 0
+/// j ← 0
+/// while j < 256 do
+///     d₁ ← B[i] + 256·(B[i+1] mod 16) d2 ← ⌊B[i+1]/16⌋+16·B[i+2] ifd1 <qthen
```
franziskuskiefer, line 28:
```suggestion
///     d₁ ← B[i] + 256·(B[i+1] mod 16) d2 ← ⌊B[i+1]/16⌋+16·B[i+2] if d1 < q then
```

specs/kyber/src/sampling.rs
```diff
@@ -5,67 +5,78 @@ use crate::{
...
+        // end if
+        if d2 < parameters::FIELD_MODULUS && sampled_coefficients < a_hat.len() {
+            a_hat[sampled_coefficients] = d2.into();
             sampled_coefficients += 1;
         }
 
-        if sampled_coefficients == out.len() {
-            return Ok(out);
+        if sampled_coefficients == a_hat.len() {
+            return Ok(a_hat);
```
franziskuskiefer, line 79: This will make trouble in hax because of early returns. I have to figure out if we can get this fixed in the tool or should work around for now.

franziskuskiefer (changes requested):
Thanks, looks good. Only a few nits.

specs/kyber/src/ntt.rs
```diff
@@ -1,154 +1,225 @@
 use crate::parameters::{KyberPolynomialRingElement, RANK};
 
-use self::kyber_polynomial_ring_element_mod::ntt_multiply;
+use self::kyber_polynomial_ring_element_mod::multiply_ntts;
 
 pub(crate) mod kyber_polynomial_ring_element_mod {
```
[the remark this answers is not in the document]
xvzcf: Removed the module, and actually moved the matrix multiplication code into `matrix.rs`.

specs/kyber/src/sampling.rs
xvzcf: This was text that wasn't supposed to be there; I removed it.

specs/kyber/src/sampling.rs
xvzcf: Moved the return out of the loop.

specs/Cargo.toml
```diff
@@ -1,2 +1,3 @@
 [workspace]
 members = ["hacspec-lib", "kyber"]
+resolver = "1"
```
franziskuskiefer, line 3: This shouldn't really do anything because it's the default. Only 2 will set it to 2.

franziskuskiefer (approved):
There's enough good stuff in here to get it in.

I took a pass and did a bunch of changes on top of this to see what we could do and to do some examples of what I think would be good to do. It's [here](https://github.com/cryspen/libcrux/compare/update-spec-to-fips-203...franziskus/update-spec-1?expand=1).
Maybe let's look at that later briefly.

specs/kyber/src/ind_cpa.rs
```diff
@@ -225,178 +203,224 @@ fn encode_and_compress_u(input: [KyberPolynomialRingElement; RANK]) -> Vec<u8> {
...
+    let error_2 = sample_poly_cbd(2, &prf_output);
 
-    // u := NTT^{-1}(AˆT ◦ rˆ) + e_1
-    let mut u = multiply_matrix_by_column(&A_transpose, &r_as_ntt).map(|r| invert_ntt(r));
+    // r̂ ← NTT(r)
+    let mut r_as_ntt = [KyberPolynomialRingElement::ZERO; RANK];
+    for i in 0..r.len() {
+        r_as_ntt[i] = ntt(r[i]);
+    }
+
```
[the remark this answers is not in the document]
xvzcf: Will do this in a separate PR.

PR #82, opened 2023-09-19
Update more of the spec to fips 203

xvzcf:


specs/hacspec-lib/src/bit_vector.rs
franziskuskiefer, line 1: I wonder if we need this at all or if this could just be a type alias now.

specs/kyber/src/serialize.rs
```diff
@@ -1,108 +1,203 @@
...
+    assert!(bits.len() <= BITS_PER_COEFFICIENT);
+
+    let modulus = 2u16.pow(bits.len().as_u32());
     let mut value: u16 = 0;
-    for (i, bit) in bit_vector.into_iter().enumerate() {
-        value |= u16::from(bit) << i;
+
+    for (i, bit) in bits.into_iter().enumerate() {
+        value += ((bit as u16) * 2u16.pow(i.as_u32())) % modulus;
     }
```
franziskuskiefer, line 152: How about this? For better readability.

```suggestion
        value += ((bit as u16) * (1 << i)) % modulus;
```

specs/kyber/src/sampling.rs
```diff
@@ -149,8 +150,7 @@ fn sum_coins(coins: &mut BitVectorChunks<'_>) -> KyberFieldElement {
 pub fn sample_poly_cbd(eta: usize, bytes: &[u8]) -> KyberPolynomialRingElement {
     assert_eq!(bytes.len(), eta * 64);
 
-    // b ← BytesToBits(B)
-    let bits: BitVector = bytes.into();
+    let bits = bytes_to_bits(bytes);
 
     let mut bits = bits.chunks(eta);
```
franziskuskiefer, line 156: Ugh, where is the `sum_coins` coming from? Did I miss this in the last review.
That probably won't work. hax might be able to deal with it by now, but we'll have to look at it tomorrow when extracting.

Please stop force pushing. That breaks the review workflow and have been a reason for this.

specs/kyber/src/sampling.rs
```diff
@@ -149,8 +150,7 @@ fn sum_coins(coins: &mut BitVectorChunks<'_>) -> KyberFieldElement {
 pub fn sample_poly_cbd(eta: usize, bytes: &[u8]) -> KyberPolynomialRingElement {
```
franziskuskiefer, line 150: We don't need an iterator here really, but could just use a slice as in the spec (`&bits[2 * i * eta..2 * i * eta + eta]` etc.)

franziskuskiefer (changes requested):
Thanks, this is already pretty close.
I think we might be able to simplify even more.

specs/kyber/src/sampling.rs
xvzcf: I did `bits.next().unwrap()`instead since I thought that'd be easier to read.

specs/kyber/src/sampling.rs
xvzcf: I'll refrain from force pushes, my bad!

specs/kyber/src/serialize.rs
xvzcf: Changed.

specs/kyber/src/serialize.rs
```diff
@@ -1,108 +1,203 @@
...
+/// return B
+/// ```
+///
+/// The NIST FIPS 203 standard can be found at
+/// <https://csrc.nist.gov/pubs/fips/203/ipd>.
+pub(crate) fn bits_to_bytes(bits: BitVector) -> Vec<u8> {
+    assert!(bits.len() % 8 == 0);
+
+    // B ← (0,...,0)
+    let mut bytes = Vec::<u8>::with_capacity(bits.len() / 8);
```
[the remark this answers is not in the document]
xvzcf: Made the change.

specs/hacspec-lib/src/bit_vector.rs
xvzcf: Removed all this code and replaced it by two `BitVector`and `BitSlice`aliases.

franziskuskiefer (approved):

PR #87, opened 2023-09-21
Created KyberVector struct and refactored code accordingly.

xvzcf:


franziskuskiefer (approved):
Thanks. Let's get this in.

PR #95, opened 2023-09-26
Don't inadvertently perform explicit rejection in Kyber.

xvzcf:


src/kem/kyber768.rs
```diff
@@ -52,11 +52,16 @@ pub fn generate_keypair(
...
     let implicit_rejection_value = &randomness[parameters::CPA_PKE_KEY_GENERATION_SEED_SIZE..];
 
-    let ind_cpa_key_pair = ind_cpa::generate_keypair(ind_cpa_keypair_randomness)?;
+    let (ind_cpa_key_pair, sampling_a_error) =
+        ind_cpa::generate_keypair(ind_cpa_keypair_randomness);
 
     let secret_key_serialized = ind_cpa_key_pair.serialize_secret_key(implicit_rejection_value);
 
-    Ok((ind_cpa_key_pair.pk(), secret_key_serialized))
+    if sampling_a_error.is_some() {
```
franziskuskiefer, line 60: Is this not going through hax?
Would allow you to avoid the `unwrap` and much nicer.

```suggestion
    `if let Some(error) = sampling_a_error {
```

src/kem/kyber768.rs
```diff
@@ -97,28 +107,26 @@ pub fn decapsulate(
...
+    // |expected_ciphertext| cannot equal |ciphertext|, thereby resulting in
+    // implicit rejection.
+    //
+    // If C is ill-formed, due to the use of hashing to obtain |pseudorandomness|
+    // as well as the fact that the Kyber CPA-PKE is sparse pseudo-random, it is
+    // highly likely that |expected_ciphertext| will not equal |ciphertext|, thereby
+    // also resulting in implicit rejection.
+    //
+    // Thus, we ignore the second return value of |ind_cpa::encrypt|.
+    let (expected_ciphertext, _) =
```
franziskuskiefer, line 125: Should we put the implicit rejection value in the second part or something instead of the constant time select? No concrete idea though ...

franziskuskiefer (changes requested):
Just a small thing

franziskuskiefer (approved):
Thanks

PR #97, opened 2023-09-27
boring kyber bench fix

franziskuskiefer:


benches/boringssl/kyber768.cxx
```diff
@@ -1,4 +1,4 @@
-#include <string>
+#include <cstring>
```
xvzcf, line 1: We could just remove this line altogether

franziskuskiefer: sure, if it's not used ... let's see what the CI says.
Also there's something off on the CI. It works on main but fails here. I have to look into that.

franziskuskiefer: It's needed, you use memset

xvzcf (approved):

benches/boringssl/kyber768.cxx
xvzcf: Strange that I can build locally without the header, but oh well

PR #107, opened 2023-10-09
move benchmarks out

franziskuskiefer:
This PR moves the benchmarks out of the main crate into a separate crate to make comparisons easier.
It also adds running the benchmarks to CI to make sure they don't break.


.github/workflows/rust.yml
```diff
@@ -134,3 +134,79 @@ jobs:
 
       - name: 🏃🏻‍♀️ Test
         run: CC=emcc AR=emar wasm-pack test --node --features wasm
+
+  benchmarks:
+    strategy:
+      fail-fast: false
```
xvzcf, line 140: Should we set this to true? If an error is reported it'll be in the setup and install code as opposed to in the benchmarking itself, and it'd be better to just fail the whole run (if a failure is reported the whole thing has to be run anyway another time before it can be merged in).

.github/workflows/rust.yml
```diff
@@ -134,3 +134,79 @@ jobs:
...
+      matrix:
+        bits: [32, 64]
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+        exclude:
+          - bits: 32
```
xvzcf, line 148: Why are we excluding 32-bit macOS? Maybe add a comment for that too?

.github/workflows/rust.yml
```diff
@@ -134,3 +134,79 @@ jobs:
...
+
+      - run: echo "RUST_TARGET_FLAG=" > $GITHUB_ENV
+        if: ${{ matrix.bits == 64 }}
+
+      - name: ⚙️ Setup Ubuntu x86
+        if: ${{ matrix.bits == 32 &&  matrix.os == 'ubuntu-latest' }}
+        run: |
+          rustup target add i686-unknown-linux-gnu
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib g++-multilib
```
xvzcf, line 170: Couldn't you also set `RUST_TARGET_FLAG` here and do all the setup for one arch+OS in one step?

.github/workflows/rust.yml
franziskuskiefer: Good point, changed it

.github/workflows/rust.yml
franziskuskiefer: There is no such thing. macOS is 64-bit only.

.github/workflows/rust.yml
franziskuskiefer: This is all c&p from above 😬 Really, all of the setup should be moved out into a reusable workflow and just called here. I added it to #54. Feel free to to do some of this if you need something in between things 🙂

.github/workflows/rust.yml
xvzcf: oops, my bad

xvzcf (approved):

PR #108, opened 2023-10-09
fstar kyber usize workaround

franziskuskiefer:
Because of https://github.com/hacspec/hacspec-v2/pull/274 `usize` can't be used to match in hax right now. This PR works around this issue for now.


karthikbhargavan (approved):
Looks good to me. 

In the future, this particular cast will be added automatically in the F* backend by hax and we can remove it from the Rust code.

PR #109, opened 2023-10-10
Add Kyber512 and Kyber1024 reference implementation

xvzcf:


src/kem/kyber/ind_cpa.rs
```diff
@@ -211,6 +211,10 @@ pub(crate) fn encrypt<
     const VECTOR_U_COMPRESSION_FACTOR: usize,
     const VECTOR_V_COMPRESSION_FACTOR: usize,
     const BLOCK_LEN: usize,
+    const ETA1: usize,
+    const ETA1_RANDOMNESS_SIZE: usize,
+    const ETA2: usize,
+    const ETA2_RANDOMNESS_SIZE: usize,
```
franziskuskiefer, line 217: eta2 is always the same so this isn't necessary. But let's keep in case it changes. Two more doesn't hurt at this point 😆

src/kem/kyber/ind_cpa.rs
```diff
@@ -309,8 +311,11 @@ pub(crate) fn decrypt<
         .chunks_exact((COEFFICIENTS_IN_RING_ELEMENT * VECTOR_U_COMPRESSION_FACTOR) / 8)
         .enumerate()
     {
-        let u = deserialize_little_endian::<VECTOR_U_COMPRESSION_FACTOR>(u_bytes);
-        u_as_ntt[i] = ntt_representation(decompress(u, 10));
+        let u = decompress(
```
franziskuskiefer, line 314: Lets be consistent and make `bits_per_compressed_coefficient` a const generic on this function. I forgot that and only did it in compress. But we should handle the two the same.

src/kem/kyber/sampling.rs
```diff
@@ -37,6 +37,19 @@ pub fn sample_from_uniform_distribution<const SEED_SIZE: usize>(
     (out, Some(BadRejectionSamplingRandomnessError))
 }
 
+#[inline(always)]
+pub(super) fn sample_from_binomial_distribution<const ETA: usize>(
+    randomness: &[u8],
+) -> KyberPolynomialRingElement {
+    debug_assert_eq!(randomness.len(), ETA * 64);
+
+    match ETA {
```
franziskuskiefer, line 46: This will fail with the current version of hax.

```suggestion
    match ETA as u32 {
```

src/kem/kyber/serialize.rs
```diff
@@ -236,43 +258,59 @@ fn serialize_little_endian_11<const OUT_LEN: usize>(
...
-
-        serialized[5 * i] = (coefficient1 & 0xFF) as u8;
-        serialized[5 * i + 1] =
-            ((coefficient2 & 0x3F) as u8) << 2 | ((coefficient1 >> 8) & 0x03) as u8;
-        serialized[5 * i + 2] =
-            ((coefficient3 & 0x0F) as u8) << 4 | ((coefficient2 >> 6) & 0x0F) as u8;
-        serialized[5 * i + 3] =
-            ((coefficient4 & 0x03) as u8) << 6 | ((coefficient3 >> 4) & 0x3F) as u8;
-        serialized[5 * i + 4] = ((coefficient4 >> 2) & 0xFF) as u8;
+    for (i, coefficients) in re.coefficients.chunks_exact(8).enumerate() {
```
franziskuskiefer, line 261: Please remove the FIXME on this function.

tests/kyber.rs
```diff
@@ -3,16 +3,14 @@ use libcrux::{
     kem::{self, Algorithm, Ct, PrivateKey},
 };
 
+use rand::{CryptoRng, Rng};
+
 #[cfg(not(target_arch = "wasm32"))]
-use libcrux::drbg::{self, RngCore};
+use libcrux::drbg;
 #[cfg(target_arch = "wasm32")]
 use rand_core::{OsRng, RngCore};
```
franziskuskiefer, line 12: `RngCore` here is unused now.

```suggestion
use rand_core::OsRng;
```

tests/kyber_nistkats.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    #[serde(with = "hex::serde")]
+    sha3_256_hash_of_ciphertext: [u8; 32],
+
+    #[serde(with = "hex::serde")]
+    shared_secret: [u8; 32],
+}
+
+#[test]
+fn kyber768_nist_known_answer_tests() {
```
franziskuskiefer, line 31: Let's put these functions in a macro or into one to avoid duplicating the code three times.

tests/kyber_nistkats.rs
```diff
@@ -0,0 +1,157 @@
...
+        for i in 0..shared_secret.len() {
+            assert_eq!(shared_secret[i], kat.shared_secret[i]);
+        }
+
+        let shared_secret_from_decapsulate =
+            kem::kyber768_decapsulate_derand(key_pair.private_key(), &ciphertext);
+        for i in 0..shared_secret.len() {
+            assert_eq!(shared_secret_from_decapsulate[i], shared_secret[i]);
+        }
+    }
```
franziskuskiefer, line 70: Why not compare the entire secret?

```suggestion
        assert_eq!(shared_secret_from_decapsulate, shared_secret.as_ref());
```

franziskuskiefer (changes requested):
Thanks!
Only a couple nits.

Also
- [ ] Please run the extraction again with hax from main. Some significant changes were merged yesterday.
- [ ] Please do a follow up with a description of how to generate the kats and the code for it.

src/kem/kyber/ind_cpa.rs
xvzcf: Heh yeah, I also thought I'd include it in to help document the code according to the specification.

tests/kyber_nistkats.rs
xvzcf: This was a relic from when I'd print all/some of the buffers for debugging, changed it all over the place.

franziskuskiefer (approved):
Thanks!

PR #120, opened 2023-10-24
More debug asserts

xvzcf:


src/kem/kyber/ntt.rs
```diff
@@ -17,6 +17,74 @@ const ZETAS_MONTGOMERY_DOMAIN: [KyberFieldElement; 128] = [
     -1530, -1278, 794, -1510, -854, -870, 478, -108, -308, 996, 991, 958, -1460, 1522, 1628,
 ];
 
+// Over time, all invocations of ntt_representation() will be replaced by
+// invocations to this function, upon which this function will be renamed back to
+// ntt_representation().
+#[inline(always)]
+pub(in crate::kem::kyber) fn ntt_with_debug_asserts(
```
franziskuskiefer, line 24: I don't think `coefficient_bound` should be an argument here.
This makes things look pretty weird and requires this additional function for now.
It looks this value should always be static, so it could be a const generic.
So it's probably something that requires some hax support to be nice in the end. But until then maybe a const generic is the way to go?
The remaining function that still uses the oldversion, could get `usize::MAX - 319584` to avoid having two functions. But then the reduction actually fails. So looks likes that's not all correct yet.

The other question is whether we need the strict individual bound or if a slightly larger, but generic bound is enough.

src/kem/kyber/ntt.rs
```diff
@@ -17,6 +17,74 @@ const ZETAS_MONTGOMERY_DOMAIN: [KyberFieldElement; 128] = [
...
+                    re[j + $layer] = re[j] - t;
+                    re[j] = re[j] + t;
+                }
+            }
+
+            layer_number += 1;
+            re.coefficients.into_iter().all(|coefficient| {
+                usize::try_from(coefficient.abs()).unwrap()
+                    < coefficient_bound + (layer_number * 3 * (FIELD_MODULUS as usize) / 2)
+            });
```
franziskuskiefer, line 72: This looks like it should only be done in debug mode? And then also actually check it with an assert?

src/kem/kyber/ntt.rs
```diff
@@ -17,6 +17,74 @@ const ZETAS_MONTGOMERY_DOMAIN: [KyberFieldElement; 128] = [
...
+                    let t = montgomery_reduce(re[j + $layer] * ZETAS_MONTGOMERY_DOMAIN[zeta_i]);
+                    re[j + $layer] = re[j] - t;
+                    re[j] = re[j] + t;
+                }
+            }
+
+            layer_number += 1;
+            re.coefficients.into_iter().all(|coefficient| {
+                usize::try_from(coefficient.abs()).unwrap()
+                    < coefficient_bound + (layer_number * 3 * (FIELD_MODULUS as usize) / 2)
```
franziskuskiefer, line 71: This may overflow. That's fine as longs we only do it in debug mode. (see next comment)

src/kem/kyber/ntt.rs
```diff
@@ -17,6 +17,74 @@ const ZETAS_MONTGOMERY_DOMAIN: [KyberFieldElement; 128] = [
     -1530, -1278, 794, -1510, -854, -870, 478, -108, -308, 996, 991, 958, -1460, 1522, 1628,
 ];
 
+// Over time, all invocations of ntt_representation() will be replaced by
+// invocations to this function, upon which this function will be renamed back to
+// ntt_representation().
+#[inline(always)]
+pub(in crate::kem::kyber) fn ntt_with_debug_asserts(
+    mut re: KyberPolynomialRingElement,
+    coefficient_bound: usize,
```
franziskuskiefer, line 26: Independent whether this should be here or not. It should probably be the same type as `coefficients`.

src/kem/kyber/ntt.rs
```diff
@@ -171,24 +250,32 @@ pub(in crate::kem::kyber) fn multiply_matrix_by_column_montgomery<const K: usize
 // this function after conversion from montgomery form lets us skip an extra
 // barrett reduction step in generate_keypair itself.
 #[inline(always)]
-pub(in crate::kem::kyber) fn multiply_matrix_by_column<const K: usize>(
-    matrix: &[[KyberPolynomialRingElement; K]; K],
-    vector: &[KyberPolynomialRingElement; K],
+#[allow(non_snake_case)]
+pub(in crate::kem::kyber) fn compute_As_plus_e<const K: usize>(
```
franziskuskiefer, line 254: Why the name change? And why a non-idiomatic name?

franziskuskiefer (changes requested):
Thanks. lgtm but I have some questions.

src/kem/kyber/ntt.rs
xvzcf: Yes, they should've been guarded by a `debug_assert`.

src/kem/kyber/ntt.rs
franziskuskiefer: ok, so let's get this in and then do a follow up.
We should make sure to get the F* extract to at least lax type check this week to make sure whatever we do here actually makes sense for the proofs.

franziskuskiefer (approved):

PR #121, opened 2023-10-24
Added nightly hax extraction job and git pre-commit hook.

xvzcf:


franziskuskiefer (changes requested):
Thanks. lgtm with two nits.

* While renaming the python script. Shall we rename it to something more generic to not only extract but also do proofs etc?
* Please add instructions on how to use the git hooks to the readme or a separate `CONTRIBUTING.md`

xvzcf:
>     * While renaming the python script. Shall we rename it to something more generic to not only extract but also do proofs etc?

What do you envision? Should we name it to something like `fstar-utils`?

>     * Please add instructions on how to use the git hooks to the readme or a separate `CONTRIBUTING.md`

Done.

franziskuskiefer:
> What do you envision? Should we name it to something like `fstar-utils`?

Not sure.. I'm bad at naming things 😬
But I see this as the helper script to extract and run proofs, and do whatever else needed for the verification workflows. So this will probably include lean as well at some point. So I'd go more towards `verify` (which might be misleading), or just `hax` (as it's all hax based). I think I like `hax` ...

franziskuskiefer (approved):
Thanks

PR #122, opened 2023-10-25
Update Kyber reference implementation to FIPS-203 draft.

xvzcf:


specs/kyber/tests/interop_with_libcrux.rs
```diff
@@ -0,0 +1,93 @@
...
+fn same_inputs_result_in_same_output() {
+    let mut keygen_seed = [0u8; KYBER768_KEY_GENERATION_SEED_SIZE];
+    OsRng.fill_bytes(&mut keygen_seed);
+
+    let libcrux_key_pair_result = libcrux::kem::kyber768_generate_keypair_derand(keygen_seed);
+    let spec_key_pair_result = hacspec_kyber::generate_keypair(keygen_seed);
+
+    if libcrux_key_pair_result.is_err() {
+        spec_key_pair_result.expect_err("If rejection sampling failed in libcrux, it should fail in the spec as well.");
+        return
```
franziskuskiefer, line 17: Let's not return here, but try again. It's very unlikely that this happens. But we don't want this test to always get out here.

src/kem/kyber.rs
```diff
@@ -107,47 +107,36 @@ pub(super) fn encapsulate<
...
+        VECTOR_U_COMPRESSION_FACTOR,
+        VECTOR_V_COMPRESSION_FACTOR,
+        VECTOR_U_BLOCK_LEN,
+        ETA1,
+        ETA1_RANDOMNESS_SIZE,
+        ETA2,
+        ETA2_RANDOMNESS_SIZE,
+    >(public_key.as_slice(), randomness, pseudorandomness);
 
     if sampling_a_error.is_some() {
```
franziskuskiefer, line 136: While we're here, let's remove the `unwrap` here (matching instead).

src/kem/kyber.rs
```diff
@@ -107,47 +107,36 @@ pub(super) fn encapsulate<
...
         Err(sampling_a_error.unwrap())
     } else {
-        Ok((ciphertext, shared_secret))
+        Ok((ciphertext, shared_secret.try_into().unwrap()))
```
franziskuskiefer, line 139: This `unwrap` here should be safe, but needs a proof of course.

src/kem.rs
```diff
@@ -167,10 +166,10 @@ pub enum Ct {
 pub enum Ss {
     X25519(x25519::PublicKey),
     P256(p256::PublicKey),
-    Kyber512(Kyber512SharedSecret),
-    Kyber768(Kyber768SharedSecret),
-    Kyber768X25519(Kyber768SharedSecret, x25519::PublicKey),
-    Kyber1024(Kyber1024SharedSecret),
+    Kyber512(KyberSharedSecret),
```
franziskuskiefer, line 169: We don't need the different variants here anymore. Let's collaps them into a `Kyber` variant. Or do we get something from the typesafety here?

franziskuskiefer (changes requested):
Thanks! lgtm with a few nits.

src/kem.rs
xvzcf: I think having type safety here still makes sense; it'd be nice to enforce the fact that `Kyber768X25519` has a `Kyber768SharedSecret` at the type level. I'm mulling over how best to do this ...

specs/kyber/tests/interop_with_libcrux.rs
franziskuskiefer, line 1: This needs a `cargo fmt`. We should maybe the spec fmt to CI as well.

src/kem/kyber.rs
```diff
@@ -25,10 +25,11 @@ pub mod kyber512;
 pub mod kyber768;
 
 impl_generic_struct!(KyberCiphertext);
-impl_generic_struct!(KyberSharedSecret);
 impl_generic_struct!(KyberPrivateKey);
 impl_generic_struct!(KyberPublicKey);
 
+pub type KyberSharedSecret = [u8; constants::SHARED_SECRET_SIZE];
```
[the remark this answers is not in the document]
xvzcf: Opened https://github.com/cryspen/libcrux/issues/123 to track this and added a comment in the code.

franziskuskiefer (approved):
Let's get this in

PR #124, opened 2023-10-27
Rearranged some code to speed things up and get bounds on compress_q and decompress_q.

xvzcf:


src/kem/kyber/ntt.rs
```diff
@@ -84,12 +83,12 @@ pub(in crate::kem::kyber) fn ntt_with_debug_asserts(
...
 
 #[inline(always)]
-pub(in crate::kem::kyber) fn ntt_representation(
+pub(in crate::kem::kyber) fn ntt_vector_u<const VECTOR_U_COMPRESSION_FACTOR: usize>(
     mut re: KyberPolynomialRingElement,
 ) -> KyberPolynomialRingElement {
     let mut zeta_i = 0;
 
-    macro_rules! layers {
+    macro_rules! ntt_at_layer {
```
franziskuskiefer, line 91: It's not great to duplicate the code in this macro. Can you pull that out so that we have it here only once?

franziskuskiefer (changes requested):
Thanks, lgtm with one nit.

franziskuskiefer (approved):
Thanks

PR #127, opened 2023-10-31
start hax requires for kyber

franziskuskiefer:


xvzcf (approved):

xvzcf (changes requested):
On second thought actually, I'm not sure if the `.depend` file should be checked-in to version control, since it's autogenerated has a bunch of system-specific paths.

franziskuskiefer:
Oh, yes. That shouldn't be there. I dropped it.
Note that I also removed the `hax` cfg environment. This will now always be set by `cargo hax` with https://github.com/hacspec/hax/pull/314.

xvzcf:
> Oh, yes. That shouldn't be there. I dropped it.

Could you also add the `.depend` to the `.gitignore` just to make things easier?

xvzcf (approved):

PR #135, opened 2023-11-22
More kyber code refactoring

xvzcf:
This PR is a grab-bag of various changes to Kyber to make things clearer, fix constant time issues and such:

-  `ntt.rs` had code for both the NTT and functions that used the NTT in matrix/vector multiplication. These latter functions have been moved to a `matrix.rs` file.
- I renamed `KyberPolynomialRingElement` and `KyberFieldElement` to just `PolynomialRingElement` and `FieldElement` (they're in the Kyber code, so the Kyber prefix is uncessary)
- Some field elements are in the montgomery domain as they've been multiplied by `R^{-1}`, others have been multiplied by `R`, so multiplications take elements from the montgomery domain to the standard domain, some the other way around, so I added some type aliases and corresponding functions to try and document this. These type aliases could become tuple structs later
- I renamed the `cbd` function in `ind_cpa.rs` to `sample_vector_cbd_then_ntt` and expanded its use.
- I added two new constant-time functions for compressing and decompressing the coefficients of the message ring element, since the message is secret information
- The `at_layer` macros used in the NTT have been converted to functions that get inlined


franziskuskiefer (changes requested):
Please describe what this PR is doing.

src/kem/kyber/arithmetic.rs
```diff
@@ -1,6 +1,16 @@
 use super::constants::{COEFFICIENTS_IN_RING_ELEMENT, FIELD_MODULUS};
 
-pub(crate) type KyberFieldElement = i32;
+// if this is 'x'
```
franziskuskiefer, line 3: Let's make these doc comments for better usability.

src/kem/kyber/arithmetic.rs
```diff
@@ -62,29 +72,40 @@ pub(crate) fn montgomery_reduce(value: KyberFieldElement) -> KyberFieldElement {
     value_high - c
 }
 
+pub(crate) fn montgomery_multiply_sfe_by_fer(
```
franziskuskiefer, line 75: Let's inline these. They are probably inlined anyway, but let's make sure to tell the compiler.

src/kem/kyber/arithmetic.rs
```diff
@@ -62,29 +72,40 @@ pub(crate) fn montgomery_reduce(value: KyberFieldElement) -> KyberFieldElement {
...
+    fe: StandardFieldElement,
+    fer: FieldElementTimesMontgomeryR,
+) -> StandardFieldElement {
+    montgomery_reduce(fe * fer)
+}
+
+pub(crate) fn to_standard_domain(mfe: MontgomeryFieldElement) -> StandardFieldElement {
+    montgomery_reduce(mfe * 1353)
```
franziskuskiefer, line 83: Can we pull out the constant and give it a name to make it clear what it is?

src/kem/kyber/arithmetic.rs
```diff
@@ -1,6 +1,16 @@
 use super::constants::{COEFFICIENTS_IN_RING_ELEMENT, FIELD_MODULUS};
 
-pub(crate) type KyberFieldElement = i32;
+// if this is 'x'
+// let 'fe' be the shorthand for this type
+pub(crate) type StandardFieldElement = i32;
```
franziskuskiefer, line 5: Maybe just call this one `FieldElement`?

src/kem/kyber/compress.rs
```diff
@@ -1,36 +1,87 @@
...
+    // if 0 <= shifted <= 831 then 0 <= shifted_positive <= 831
+    let shifted_to_positive = (shifted >> 15) ^ shifted;
+
+    let shifted_positive_in_range = shifted_to_positive - 832;
+
+    // If x <= 831, then x - 832 <= -1 => x - 832 < 0.
+    ((shifted_positive_in_range >> 15) & 1) as u8
+}
+
+#[cfg_attr(hax, hax_lib_macros::requires((coefficient_bits == 4 || coefficient_bits == 5 || coefficient_bits == 10 || coefficient_bits == 11) && fe < (FIELD_MODULUS as u16)))]
```
franziskuskiefer, line 37: Can we break these long lines? They're a little hard to read.

src/kem/kyber/serialize.rs
```diff
@@ -1,35 +1,25 @@
 use super::{
-    arithmetic::{to_unsigned_representative, KyberFieldElement, KyberPolynomialRingElement},
-    compress::{compress_q, decompress_q},
+    arithmetic::{to_unsigned_representative, PolynomialRingElement, StandardFieldElement},
+    compress::{
+        compress_ciphertext_coefficient, compress_message_coefficient,
+        decompress_ciphertext_coefficient, decompress_message_coefficient,
+    },
     constants::{BYTES_PER_RING_ELEMENT, COEFFICIENTS_IN_RING_ELEMENT, SHARED_SECRET_SIZE},
```
franziskuskiefer, line 7: `COEFFICIENTS_IN_RING_ELEMENT` seems unused? It's only used in hax debug asserts. So either only include it for hax, or use the qualified path to avoid the warning.

src/kem/kyber/ind_cpa.rs
```diff
@@ -68,11 +69,15 @@ fn sample_matrix_A<const K: usize>(
 }
 
 #[inline(always)]
-fn cbd<const K: usize, const ETA: usize, const ETA_RANDOMNESS_SIZE: usize>(
+fn sample_vector_cbd_then_ntt<
```
franziskuskiefer, line 72: Please add comments to functions.

src/kem/kyber/matrix.rs
franziskuskiefer, line 1: Please add comments here as well.

src/kem/kyber/matrix.rs
```diff
@@ -0,0 +1,113 @@
...
+    let mut result = PolynomialRingElement::ZERO;
+
+    for i in 0..K {
+        let product = ntt_multiply(&secret_as_ntt[i], &u_as_ntt[i]);
+        result = add_to_ring_element::<K>(result, &product);
+    }
+
+    result = invert_ntt_montgomery::<K>(result);
+
+    for i in 0..result.coefficients.len() {
```
franziskuskiefer, line 24: Let's use the constant size `COEFFICIENTS_IN_RING_ELEMENT` here to avoid the `len()`.

src/kem/kyber/matrix.rs
```diff
@@ -0,0 +1,113 @@
...
+    let mut result = PolynomialRingElement::ZERO;
+
+    for i in 0..K {
+        let product = ntt_multiply(&t_as_ntt[i], &r_as_ntt[i]);
+        result = add_to_ring_element::<K>(result, &product);
+    }
+
+    result = invert_ntt_montgomery::<K>(result);
+
+    for i in 0..result.coefficients.len() {
```
franziskuskiefer, line 49: Same here `COEFFICIENTS_IN_RING_ELEMENT`.

src/kem/kyber/matrix.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    for (i, row) in a_as_ntt.iter().enumerate() {
+        for (j, a_element) in row.iter().enumerate() {
+            let product = ntt_multiply(a_element, &r_as_ntt[j]);
+            result[i] = add_to_ring_element::<K>(result[i], &product);
+        }
+
+        result[i] = invert_ntt_montgomery::<K>(result[i]);
+
+        for j in 0..result[i].coefficients.len() {
```
franziskuskiefer, line 76:
```suggestion
        for j in 0..COEFFICIENTS_IN_RING_ELEMENT {
```

src/kem/kyber/matrix.rs
```diff
@@ -0,0 +1,113 @@
...
+) -> [PolynomialRingElement; K] {
+    let mut result = [PolynomialRingElement::ZERO; K];
+
+    for (i, row) in matrix_A.iter().enumerate() {
+        for (j, matrix_element) in row.iter().enumerate() {
+            let product = ntt_multiply(matrix_element, &s_as_ntt[j]);
+            result[i] = add_to_ring_element::<K>(result[i], &product);
+        }
+
+        for j in 0..result[i].coefficients.len() {
```
franziskuskiefer, line 102: `COEFFICIENTS_IN_RING_ELEMENT`

src/kem/kyber/ntt.rs
```diff
@@ -17,47 +18,83 @@ const ZETAS_MONTGOMERY_DOMAIN: [KyberFieldElement; 128] = [
     -1530, -1278, 794, -1510, -854, -870, 478, -108, -308, 996, 991, 958, -1460, 1522, 1628,
 ];
 
-macro_rules! ntt_at_layer {
-    ($layer:literal, $zeta_i:ident, $re:ident, $initial_coefficient_bound:literal) => {
-        let step = 1 << $layer;
+#[inline(always)]
+fn ntt_at_layer(
```
franziskuskiefer, line 22: This file needs comments too.

src/kem/kyber/ntt.rs
```diff
@@ -85,80 +122,56 @@ pub(in crate::kem::kyber) fn ntt_binomially_sampled_ring_element(
...
-            }
-        };
-    }
+    (re, zetas_index) = invert_ntt_at_layer(1, zetas_index, re, initial_coefficient_bound);
+    (re, zetas_index) = invert_ntt_at_layer(2, zetas_index, re, initial_coefficient_bound);
+    (re, zetas_index) = invert_ntt_at_layer(3, zetas_index, re, initial_coefficient_bound);
+    (re, zetas_index) = invert_ntt_at_layer(4, zetas_index, re, initial_coefficient_bound);
+    (re, zetas_index) = invert_ntt_at_layer(5, zetas_index, re, initial_coefficient_bound);
+    (re, zetas_index) = invert_ntt_at_layer(6, zetas_index, re, initial_coefficient_bound);
+    (re, _) = invert_ntt_at_layer(7, zetas_index, re, initial_coefficient_bound);
```
franziskuskiefer, line 168: Not using mutability for `re` and `zetas_index` in these functions might make the extaction nicer.
I also see a slight performance increase with that change. But it's pretty noisy so I'm not sure if it's real.

src/kem/kyber/serialize.rs
```diff
@@ -89,16 +77,18 @@ pub(super) fn deserialize_to_uncompressed_ring_element(
...
+fn compress_then_serialize_10<const OUT_LEN: usize>(re: PolynomialRingElement) -> [u8; OUT_LEN] {
     let mut serialized = [0u8; OUT_LEN];
 
     for (i, coefficients) in re.coefficients.chunks_exact(4).enumerate() {
-        let coefficient1 = compress_q(10, to_unsigned_representative(coefficients[0]));
-        let coefficient2 = compress_q(10, to_unsigned_representative(coefficients[1]));
-        let coefficient3 = compress_q(10, to_unsigned_representative(coefficients[2]));
-        let coefficient4 = compress_q(10, to_unsigned_representative(coefficients[3]));
+        let coefficient1 =
+            compress_ciphertext_coefficient(10, to_unsigned_representative(coefficients[0]));
```
franziskuskiefer, line 85: Couldn't `compress_ciphertext_coefficient` take an `i32` and call `to_unsigned_representative` itself? This would make the code less noisy because you don't need to call `to_unsigned_representative` so many times.
If you want to keep the preconditions on `compress_ciphertext_coefficient`, you could add one function in between that does the conversion.

franziskuskiefer (changes requested):
Thanks. Generally looks like good cleanup.
Let's add comments to things you're touching to get comments back in here.

src/kem/kyber/serialize.rs
xvzcf: Added a `cfg` guard.

src/kem/kyber/ntt.rs
xvzcf: I had to revert this change and convert it back to a macro, since hax does not extract this properly.

src/kem/kyber/serialize.rs
xvzcf: This could work, but I'd rather investigate it and put up code in a separate PR.

src/kem/kyber/ntt.rs
```diff
@@ -106,72 +73,101 @@ pub(in crate::kem::kyber) fn ntt_binomially_sampled_ring_element(
...
+        re.coefficients[0].abs() < 128 * (K as i32) * 3329
+            && re.coefficients[1].abs() < 128 * (K as i32) * 3329
+    );
     hax_lib::debug_assert!(re
         .coefficients
         .into_iter()
         .enumerate()
         .skip(2)
-        .all(|(i, coefficient)| coefficient.abs() < (128 / (1 << i.ilog2())) * FIELD_MODULUS));
+        .all(|(i, coefficient)| coefficient.abs() < (128 / (1 << i.ilog2())) * 3329));
```
franziskuskiefer, line 170: Why don't you use the `FIELD_MODULUS`?
Let's use constants whenever possible and add comments when not.

franziskuskiefer (approved):
Thanks, only one nit.
Let's make sure to get laxing on CI before merging this. But otherwise this is good to go.

src/kem/kyber/ntt.rs
xvzcf: Ah, I made that change, but it got lost with the revert.

src/kem/kyber/arithmetic.rs
```diff
@@ -1,6 +1,16 @@
 use super::constants::{COEFFICIENTS_IN_RING_ELEMENT, FIELD_MODULUS};
 
-pub(crate) type KyberFieldElement = i32;
+/// Values having this type hold a representative 'x' of the Kyber Field.
+/// Let 'fe' be the shorthand for this type
+pub(crate) type FieldElement = i32;
```
karthikbhargavan, line 5: I wonder if we could go one step further and make this an anonymous struct, so that the Rust type system enforces these types. But for this PR, this is already good documentation.

src/kem/kyber/serialize.rs
```diff
@@ -1,35 +1,28 @@
...
 ) -> [u8; SHARED_SECRET_SIZE] {
     let mut serialized = [0u8; SHARED_SECRET_SIZE];
 
     for (i, coefficients) in re.coefficients.chunks_exact(8).enumerate() {
         for (j, coefficient) in coefficients.iter().enumerate() {
             let coefficient = to_unsigned_representative(*coefficient);
 
-            let coefficient_compressed = compress_q(1, coefficient);
-
-            // At this point, the following should hold for |coefficient_compressed|:
```
karthikbhargavan, line 19: Does this comment no longer apply?

src/kem/kyber/arithmetic.rs
```diff
@@ -17,7 +27,7 @@ const BARRETT_MULTIPLIER: i64 = 20159; // floor((BARRETT_R / FIELD_MODULUS) + 0.
 
 #[cfg_attr(hax, hax_lib_macros::requires((i64::from(value) > -BARRETT_R && i64::from(value) < BARRETT_R)))]
 #[cfg_attr(hax, hax_lib_macros::ensures(|result| result > -FIELD_MODULUS && result < FIELD_MODULUS))]
-pub(crate) fn barrett_reduce(value: KyberFieldElement) -> KyberFieldElement {
```
karthikbhargavan, line 20: Should we generally be removing all references to Kyber and replacing them with ML-KEM-based nomenclature?

karthikbhargavan (approved):
All looks good, thanks.

src/kem/kyber/serialize.rs
xvzcf: It does, it's just that it's now enforced by the way `compress_message_coefficient` is written.

src/kem/kyber/arithmetic.rs
xvzcf: I'll slowly try and make them structs as well, the main thing is we'd need traits/multiple type definitions/some other way to hold all these types in RingElements

src/kem/kyber/arithmetic.rs
xvzcf: Yeah, might as well, unless there's another name change on the horizon ...

PR #136, opened 2023-11-23
kyber crate extraction script

franziskuskiefer:
A simple shell script to pull out kyber as a separate crate for Aeneas.

cc @protz

### DIFF
--- a/kyber-crate.sh
+++ b/kyber-crate.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -e
+
+mkdir -p kyber-crate/src
+cp -r src/kem/kyber* kyber-crate/src
+cd kyber-crate
+mv src/kyber.rs src/lib.rs
+mv src/kyber/* src
+
+cat > Cargo.toml <<EOF
+[package]
+name = "libcrux_kyber"
+version = "0.1.0"
+authors = ["Cryspen"]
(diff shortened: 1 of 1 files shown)


xvzcf:
When running the script I get

```
fixing up src/PERFORMANCE.md ...
sed: 1: "src/PERFORMANCE.md": unterminated substitute pattern
```

speaking of which, I need to start regularly updating this file with every PR again ...

franziskuskiefer:
> When running the script I get

hm, do you use mac `sed`? `gsed` should be able to handle this. I didn't want to make this more complex than necessary ...

> speaking of which, I need to start regularly updating this file with every PR again ...

Yeah we should update the numbers from time to time....

xvzcf:
> > When running the script I get
> 
> hm, do you use mac `sed`? `gsed` should be able to handle this. I didn't want to make this more complex than necessary ...

Yeah, I used `gsed` and it worked. We can just leave the script as is since it's temporary.

franziskuskiefer:
> Yeah, I used `gsed` and it worked. We can just leave the script as is since it's temporary.

Yes, we should delete this as soon as it's not needed anymore.

xvzcf (approved):
This is fine with me, I'll wait for Jonathan to approve.

PR #138, opened 2023-11-29
Added some more preconditions using hax::implies and hax::forall and slight refactoring.

xvzcf:
- Removed all use of `map()` for the C-extraction


xvzcf:
> Looks like hax is failing

Yeah, I opened https://github.com/hacspec/hax/pull/380 to fix this.

PR #140, opened 2023-12-04
Makefile for verification

karthikbhargavan:
This Makefile lists files that should be fully typechecked, and those that should be typechecked while admitting SMT queries. Over time, we expect more modules to be promoted to VERIFIED.


franziskuskiefer (approved):
lgtm, merge it when the CI is happy

PR #153, opened 2023-12-13
Kyber C extraction

franziskuskiefer:
This PR modifies kyber such that it can be extracted to C.
The `kyber-crate.sh` can be used to extract and copy the C files to hacl-packages (see cryspen/hacl-packages#442).

Main changes
* use the `cloop` macro to rewrite all iterators into range loops (this macro does not change the code for F* extraction)
* rewrite an `unwrap()` into a `match` with a `panic` to work around a Eurydice issue
* rewrite an `iter` into an `into_iter` to work around a Eurydice issue


src/kem/kyber.rs
```diff
@@ -133,7 +136,11 @@ pub(super) fn encapsulate<
         ETA2_RANDOMNESS_SIZE,
     >(public_key.as_slice(), randomness, pseudorandomness);
 
-    (ciphertext.into(), shared_secret.try_into().unwrap())
+    let shared_secret = match shared_secret.try_into() {
+        Ok(shared_secret) => shared_secret,
+        Err(_) => panic!(),
```
dconnolly, line 141: should this panic or return an Error?

franziskuskiefer: This should indeed panic because it can not fail (as proven in F* as well, Rust just doesn't know it unfortunately). The panic is just explicit here (instead of the unwrap) because of some shortcomings in the tooling we work around for now.

src/kem/kyber/helper.rs
```diff
@@ -0,0 +1,54 @@
...
+        for $i in $start..$end / $step {
+            let $i = $i * $step;
+            $body
+        }
+    };
+}
+
+#[cfg(hax)]
+#[macro_export]
+macro_rules! cloop {
```
karthikbhargavan, line 38: This looks good to me, we may redefine this differently for proofs in another PR, if needed.

src/kem/kyber/ind_cpa.rs
```diff
@@ -48,11 +49,13 @@ fn deserialize_public_key<const K: usize, const T_AS_NTT_ENCODED_SIZE: usize>(
     public_key: &[u8],
 ) -> [PolynomialRingElement; K] {
     let mut t_as_ntt = [PolynomialRingElement::ZERO; K];
-    for (i, t_as_ntt_bytes) in public_key[..T_AS_NTT_ENCODED_SIZE]
```
karthikbhargavan, line 51: This seems strange. If the function does not need this slice, it does not need to have T_AS_NTT_ENCODED_SIZE as a const parameter either?

src/kem/kyber/ind_cpa.rs
```diff
@@ -261,12 +268,14 @@ fn deserialize_then_decompress_u<
     ciphertext: &[u8; CIPHERTEXT_SIZE],
 ) -> [PolynomialRingElement; K] {
     let mut u_as_ntt = [PolynomialRingElement::ZERO; K];
-    for (i, u_bytes) in ciphertext[..VECTOR_U_ENCODED_SIZE]
```
karthikbhargavan, line 264: Same here, why do we not to slice to VECTOR_U_ENCODED_SIZE any more?

src/kem/kyber/helper.rs
```diff
@@ -0,0 +1,54 @@
...
+    };
+    (for ($i:ident, $chunk:ident) in $val:ident.iter().enumerate() $body:block) => {
+        for $i in 0..$val.len() {
+            let $chunk = &$val[$i];
+            $body
+        }
+    };
+    (for ($i:ident, $chunk:ident) in $val:ident.into_iter().enumerate() $body:block) => {
+        for $i in 0..$val.len() {
+            let $chunk = $val[$i];
```
karthikbhargavan, line 24: should this be $item instead of $chunk?

src/kem/kyber/helper.rs
```diff
@@ -0,0 +1,54 @@
...
+    };
+    (for ($i:ident, $chunk:ident) in $val:ident.chunks_exact($($chunk_size:expr),*).enumerate() $body:block) => {
+        for $i in 0..$val.len() / ($($chunk_size)*) {
+            let $chunk = &$val[$i*($($chunk_size)*) .. $i*($($chunk_size)*)+($($chunk_size)*)];
+            $body
+        }
+    };
+    (for ($i:ident, $chunk:ident) in $val:ident.iter().enumerate() $body:block) => {
+        for $i in 0..$val.len() {
+            let $chunk = &$val[$i];
```
karthikbhargavan, line 18: should this be $item instead of $chunk?

src/kem/kyber/helper.rs
franziskuskiefer: No, this should be `$chunk`. Otherwise the loop has no way to accessing it. Adding another `$item` would require changing the syntax of the loop that goes into the macro. We don't want that.

src/kem/kyber/helper.rs
franziskuskiefer: See the other one.

src/kem/kyber/helper.rs
franziskuskiefer: Yep makes sense. This is really just keeping the same code for extraction as before.

src/kem/kyber/ind_cpa.rs
franziskuskiefer: These were drive-by changes because we don't need the slices. I didn't want to change more. But we indeed don't need the generic anymore here. If that doesn't create too much churn for the proofs, I'm happy to drop the unnecessary const generics.

src/kem/kyber/helper.rs
karthikbhargavan: I mean that this variable should not be called $chunk, since it is an item of the enumerate, not a chunk.

PR #155, opened 2023-12-31
Remove more divisions in compress and elsewhere.

xvzcf:


dconnolly:
Any good comments to include on why this and not the previous technique?

Also:

karthikbhargavan (commented):
This fix appears to be a bit different from what PQ-Crystals does? Shall we verify and then justify why we chose this patch instead?

franziskuskiefer (approved):

xvzcf:
> Any good comments to include on why this and not the previous technique?

I will do a big documentation overhaul of the code next week; will add comments about this then (along with a link to the announcement on the mailing list).

PR #156, opened 2024-01-03
Expose the SHAKE-128 incremental API in libcrux and add 100,000 Kyber KAT tests using this API

xvzcf:


tests/kyber_shake_kats.rs
```diff
@@ -0,0 +1,76 @@
...
+#[test]
+fn kyber_768_100000_kats() {
+    let mut rng = Shake128Rng::new();
+    let mut shake128 = AbsorbManySqueezeOnceShake128::new();
+
+    for _ in 0..100_000 {
+        let key_generation_seed = rng.read::<64>();
+        let key_pair = kem::kyber768_generate_keypair_derand(key_generation_seed);
+
+        shake128.absorb(key_pair.public_key().as_ref());
```
franziskuskiefer, line 54: I can't think of a good way but maybe you can. If we could parallelise this, it would go much faster. But because we need `shake128` we can't really do that.

src/hacl/sha3.rs
```diff
@@ -94,6 +94,89 @@ pub(crate) fn shake256<const BYTES: usize>(data: &[u8]) -> [u8; BYTES] {
     out
 }
 
+pub mod incremental {
```
franziskuskiefer, line 97: This module needs comments.
Can you file a follow up issue to not make this public but wrap it in a proper API in `digest`?

src/hacl/sha3.rs
```diff
@@ -94,6 +94,89 @@ pub(crate) fn shake256<const BYTES: usize>(data: &[u8]) -> [u8; BYTES] {
...
+                Hacl_Hash_SHA3_Scalar_shake128_absorb(
+                    self.state,
+                    input.as_ptr() as _,
+                    input.len() as u32,
+                )
+            };
+        }
+
+        pub fn squeeze_nblocks<const OUTPUT_BYTES: usize>(&mut self) -> [u8; OUTPUT_BYTES] {
+            debug_assert!(OUTPUT_BYTES % 168 == 0);
```
franziskuskiefer, line 127: especially something like this needs to go into the comment.

src/hacl/sha3.rs
```diff
@@ -94,6 +94,89 @@ pub(crate) fn shake256<const BYTES: usize>(data: &[u8]) -> [u8; BYTES] {
...
+    };
+
+    pub struct AbsorbManySqueezeOnceShake128 {
+        state: *mut Hacl_Hash_SHA3_state_t_s,
+    }
+
+    impl AbsorbManySqueezeOnceShake128 {
+        pub fn new() -> Self {
+            Self {
+                state: unsafe { Hacl_Hash_SHA3_malloc(12) },
```
franziskuskiefer, line 158: Where is the `12` coming from?

franziskuskiefer (commented):
Generally looks good. A couple comments.

tests/kyber_kats/generate_shake128_kats_hash.py
```diff
@@ -0,0 +1,65 @@
+#! /usr/bin/env pypy3
```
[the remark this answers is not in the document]
xvzcf: I use `pypy3` for this script since using the `python3` interpreter would be much much slower, especially when generating 100,000 KATs.

sys/hacl/build.rs
```diff
@@ -107,9 +107,14 @@ fn create_bindings(platform: &Platform, home_dir: &Path) {
         // Allow function we want to have in
         .allowlist_function("Hacl_AEAD_Chacha20Poly1305_.*")
         .allowlist_function("Hacl_Curve25519.*")
+        .allowlist_function("Hacl_Hash_SHA1.*")
         .allowlist_function("Hacl_Hash_SHA2.*")
-        .allowlist_function("Hacl_Hash_.*")
-        .allowlist_function("Hacl_Blake2.*")
+        .allowlist_function("Hacl_Hash_SHA3.*")
+        .allowlist_function("Hacl_Hash_Blake2.*")
```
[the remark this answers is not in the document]
xvzcf: I thought it would just document better what exactly we want and what we don't want. Also, this let me exclude `Hacl_Hash_SHA3_Scalar_sha3*`, `Hacl_Hash_SHA3_Scalar_shake128` and the like from the extraction (they weren't being used and I didn't want to add them in).

tests/kyber_shake_kats.rs
xvzcf: I can't off the top of my head, I added a TODO in there for now and we can get to it at some point.

franziskuskiefer:
This is blocked on fixing the includes for avx2 in https://github.com/cryspen/hacl-packages/pull/443.

karthikbhargavan:
Is this PR still alive @xvzcf?
It seems we already did this elsewhere.

xvzcf:
> Is this PR still alive @xvzcf? It seems we already did this elsewhere.

I think the SHA3 incremental API is exposed, but I don't think we have the 100k KAT test, I can open a fresh PR adding that.

PR #162, opened 2024-01-09
Kyber: F*: enable the `--interfaces` option for most modules

W95Psp:
This PR enables the `--interfaces` option to `cargo hax` and refreshes the extraction of Kyber


franziskuskiefer (approved):
Thanks

PR #169, opened 2024-01-11
Secret independence for Kyber

karthikbhargavan:
This PR adds secret independence proofs on top of the existing proofs for Kyber.

It correctly flags the timing leak in this older version of our code:
https://github.com/cryspen/libcrux/blob/2a3cd5d4b09de1fb214c0ae9c6ee915877794bbc/src/kem/kyber/compress.rs#L59

It also flags potential leaks in these lines:
https://github.com/cryspen/libcrux/blob/ea52806f4961c53f71e819740370f05895ac16e7/src/kem/kyber/sampling.rs#L33

https://github.com/cryspen/libcrux/blob/ea52806f4961c53f71e819740370f05895ac16e7/src/kem/kyber/sampling.rs#L37


karthikbhargavan:
Actually, this PR should not be merged in the current state.

karthikbhargavan:
Closed in favor of #171

PR #171, opened 2024-01-13
Secret Independence for Kyber 

karthikbhargavan:
This PR adds secret independence proofs on top of the existing proofs for Kyber.

It correctly flags the timing leak in this older version of our code:
https://github.com/cryspen/libcrux/blob/2a3cd5d4b09de1fb214c0ae9c6ee915877794bbc/src/kem/kyber/compress.rs#L59

This leak has now been fixed and our analysis verifies that the leak no longer exits.

The analysis also flags potential leaks in these lines:
https://github.com/cryspen/libcrux/blob/ea52806f4961c53f71e819740370f05895ac16e7/src/kem/kyber/sampling.rs#L33

https://github.com/cryspen/libcrux/blob/ea52806f4961c53f71e819740370f05895ac16e7/src/kem/kyber/sampling.rs#L37

For these instances, we add declassification operations to allow them.

The PR adds a new directory that annotates the extracted F* code with type annotations needed for secret independence (classifications or public/secret declarations.) In a future PR, we would like to backport these changes to Rust so that the new directory becomes unnecessary.

### DIFF
--- a/.github/workflows/hax.yml
+++ b/.github/workflows/hax.yml
@@ -80,6 +80,14 @@ jobs:
               PATH="${PATH}:${{ github.workspace }}/fstar/bin" \
               make -C proofs/fstar/extraction-edited
 
+      - name: 🏃 Verify Kyber `extraction-secret-independent` F* code
+        run: |
+          env FSTAR_HOME=${{ github.workspace }}/fstar \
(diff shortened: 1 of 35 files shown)


franziskuskiefer (approved):
🥳 let's get it in

PR #172, opened 2024-01-14
hacl build update

franziskuskiefer:
Set some more platform dependent flags for hacl.

### DIFF
--- a/sys/hacl/build.rs
+++ b/sys/hacl/build.rs
@@ -237,6 +237,13 @@ fn build(platform: &Platform, home_path: &Path) {
     let mut defines = vec![];
     defines.push(("RELOCATABLE", "1"));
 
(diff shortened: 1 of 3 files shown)


sys/hacl/build.rs
```diff
@@ -416,6 +425,7 @@ fn main() {
             target_arch: target_arch.clone(),
             target_env: target_env.clone(),
             target_os: target_os.clone(),
+            int128: target_arch == "x86_64" || target_arch == "aarch64",
```
karthikbhargavan, line 428: I will add an issue to address arm32 in libcrux in a future PR.

karthikbhargavan (approved):
Looks good, thanks for this, perf looks much more respectable.

PR #177, opened 2024-01-19
Working around regressions in typeclass inference in F*

karthikbhargavan:
This PR fixes a couple of places where F* 2024.01.13 fails but F* 2023.09.03 used to succeed.
The issue appears when a typeclass instance is used within an arithmetic or logical expression.


karthikbhargavan:
I also added a drive-by fix that adds new hints for the correctness proofs. These are needed when upgrading F* and Z3. Since no code is changed, and it is related to the new F*, I folded it into this PR.

franziskuskiefer (approved):
Let's get this in to unbreak ci.

franziskuskiefer (changes requested):
Looks like more changes are needed.
This is blocking everything else, so let's get this fixed.

W95Psp:
I think I understood why the TC resolution is failing, I think https://github.com/hacspec/hax/pull/444 is solving the issue

I've pushed a temporary commit that pins Hax to the branch of https://github.com/hacspec/hax/pull/444, let's revert it when the PR is merged

W95Psp (approved):

franziskuskiefer (approved):

PR #183, opened 2024-01-24
Kyber documentation update

xvzcf:
- Added more comments to the Kyber code.
- Added an `implementation_notes.pdf` file describing, among other things, Barrett and Montgomery reduction and justifying some of the constants and operations used in the code.

I moved `sample_ring_element_cbd` above `sample_vector_cbd_then_ntt` since the former could be renamed to `sample_vector_cbd` and is a more basic operation than `sample_vector_cbd` and then ntt. I couldn't rerun hax on this change, could one of you do that? if it causes an issue in the extraction it can be easily reversed.


franziskuskiefer (approved):

karthikbhargavan (approved):
The F* changes look fine.

PR #189, opened 2024-01-26
mlkem pk validation

franziskuskiefer:
* pk validation for ml-kem
* kem API cleanup (move encaps and decaps functions into public key and ciphertext)
* ml-kem rename
* fixed fuzzers


franziskuskiefer:
> * Do we still leave the folder name and file names as kyber* ?

I'd prefer renaming, but felt like it would make things harder than necessary right now for the proofs.

> * There is a weird structure in the current code, where the IND-CCA code is in `kyber.rs` and is one folder up from the rest of the code, and this file is then called within the 3 kyber instances. It would be better to have an `indcca.rs` file within the `kem/kyber/` folder and have the current `kyber.rs` act as a wrapper for the three instances?

That's been an open issue for a while. We didn't do it because of the proofs. The correct structure would be to have an empty `kyber` module and everything in the kyber directory, or the three variants in the `kyber` module and everything else below.

I'm fine either way.

hax-driver.py
```diff
@@ -115,11 +115,39 @@ def shell(command, expect=0, cwd=None, env={}):
...
+            "proofs/fstar/extraction/",
+        ]
+    )
+    # Extract ml-kem from libcrux
     shell(
         cargo_hax_into
         + [
             "-i",
-            "-** +libcrux::kem::kyber::** {} -libcrux::digest::** -libcrux::**::types::index_impls::**".format(
+            "-** +libcrux::kem::kyber::** +libcrux_platform::** {} -libcrux::**::types::index_impls::**".format(
```
W95Psp, line 150: I think we should just `hax into` `libcrux` and `libcrux-platform` in one go:
```python
shell(
        cargo_hax_into
        + [
            "-C", "-p", "libcrux", "-p", "libcrux-platform", ";",
            "-i",
            "-** +libcrux::kem::kyber::** +libcrux_platform::** {} -libcrux::**::types::index_impls::**".format(
                exclude_sha3_implementations
            ),
            "fstar",
            "--interfaces",
            "+* -libcrux::kem::kyber::types",
        ],
        cwd=".",
        env=hax_env,
    )
```

and then either copy over the fsti or just include then in the makefile.
Including in the makefile might actually be the best option.

franziskuskiefer: Well, if that would actually work, yes :) But this fails just like `--deps` (hacspec/hax#468).
But changing the makefile rather than copying the file over is a good idea!

W95Psp: Oh, that's not great... https://github.com/hacspec/hax/pull/469 fixes it though.
Shall we just merge that PR in Hax and use that?

Ok, do you want me to do that in this PR?

franziskuskiefer: Yeah, can you do the makefile and hax-driver changes here? I'm looking at https://github.com/hacspec/hax/pull/469 then we can get both in.

W95Psp: I pushed those changes.
https://github.com/hacspec/hax/pull/469 will be soon in, I stopped the CI for now, will re-trigger it once hax' PR is merged in main

PR #190, opened 2024-01-26
Allow libcrux to be compiled for no_std (embedded) environments

sivadeilra:
This allows libcrux to be compiled with the minimal Rust runtime, also known as `no_std`. This is necessary for some highly-constrained code environments, such as embedded devices, kernels, etc., where the full dependency set of the Rust standard library is too large or requires capabilities that do not exist in the target environment.

To do this, I added a new Cargo "feature" to the crate, called `std`. This feature is enabled by default. To target a `no_std` environment, you must first disable the default feature, then opt-in to only the features that your target environment needs.

The `rand` crate itself currently has a dependency on `std`, so the `rand` feature depends on `std`. If the `rand` crate itself provides a `no_std` environment, then this requirement could be lifted.


sivadeilra:
Assuming this goes forward, the CI/CD will need a new build job that verifies building with `cargo build --no-default-features`, to verify that `no_std` builds correctly.

franziskuskiefer (changes requested):
Thanks!

I moved it back against main. I think we want this on main right away.
There are some (easy to resolve) conflicts. Can you resolve them?

And then I noticed that `rand` is still being used. The solution there should be to make `rand` optional (only enable with the `rand` feature), and convert everything, except the `rand` feature guarded drbg pieces to use `RngCore` and include `rand_core` instead.

franziskuskiefer:
@sivadeilra do you plan on updating this PR?

Issue #192, opened 2024-01-29
[Meta] Kyber F* Proof Cleanup and CI


karthikbhargavan:
- [x] CI: use diffs for auto-generation of `extraction-{edited,secret-independent}` folders (see #193)
 - [ ] Proof cleanup: minimize a bit the initial diff

PR #193, opened 2024-01-29
F*: kyber: add and use diffs for `proofs/fstar/extraction-*` folders

W95Psp:
Working towards #192 

TODOs:
  - [x] add readme describing the workflow
  - ~check ALL snapshots are up to date~ will do in a subsequent PR


franziskuskiefer (commented):
Should the hints really be in there? That makes the patch huge and I don't see why it needs to be there. We could just keep the files around. Same for the Makefile.
So that the patch is really a diff on the F* files.

W95Psp:
Mmh, right. This makes sense for hint files, indeed.
But maybe for makefile it's good? things should be pretty similar.

W95Psp:
Fixed that, now the script drops hints and makefiles!

franziskuskiefer (commented):
Thanks, already much smaller 🙂
But it doesn't apply right now 🤔

I see the vscode config (`Kyber.fst.config.json`) in there and the `PROOFS.md` in the patch as well. Should we pull them out too?

No strong opinion on them. This depends on the workflow as well. Speaking of which, let's document how to work with this somewhere in here.

proofs/fstar/README.md
```diff
@@ -0,0 +1,36 @@
...
+are hand-edited snapshots of `extraction` at some point in time.
+
+The edits (1) on the one hand between `extraction` and
+`extraction-edited` and (2) on the other hand between
+`extraction-edited` and `extraction-secret-independent` are tracked
+via diff files.
+
+Whenever the rust code changes, the extracted F* code in `extraction`
+changes as well. The CI then apply the diff files: if the patch
+process fails or if the resulting patched F* doesn't typecheck, the CI
```
franziskuskiefer, line 13:
```suggestion
changes as well. The CI then applies the diff files: if the patch
```

proofs/fstar/README.md
```diff
@@ -0,0 +1,36 @@
...
+fails.
+
+The bash script `./patches.sh` takes care of the diffs:
+ - `./patches.sh create` creates patches out of the `extraction*` folders;
+ - `./patches.sh apply` drops the `extraction-edited` and
+   `extraction-secret-independent` folders and re-create them out of
+   the `extraction-edited.patch` and
```
franziskuskiefer, line 20:
```suggestion
   `extraction-secret-independent` folders and re-creates them out of
```

proofs/fstar/README.md
```diff
@@ -0,0 +1,36 @@
...
+`extraction-edited` and (2) on the other hand between
+`extraction-edited` and `extraction-secret-independent` are tracked
+via diff files.
+
+Whenever the rust code changes, the extracted F* code in `extraction`
+changes as well. The CI then apply the diff files: if the patch
+process fails or if the resulting patched F* doesn't typecheck, the CI
+fails.
+
+The bash script `./patches.sh` takes care of the diffs:
```
franziskuskiefer, line 16: I assume the `extraction` directory needs to be populated for this? Can you add that.

proofs/fstar/README.md
```diff
@@ -0,0 +1,36 @@
...
+   toolchain, we should not commit it in the repository;
+ - since `extraction-edited` and `extraction-secret-independent` are
+   created via the diff files, we should not commit them either in the
+   repository.
+
+Currently, we do check those 3 folders in. Thus, when we change
+anything in the F* proofs, we should always make sure we keep the diff
+files and those folder in a coherent state.
+
+In other words, this means that the last commit of each PR should be
```
franziskuskiefer, line 35: You can create a git hook for this if you want. There's one for extraction already.

franziskuskiefer (approved):
Thanks, let's get this in and then clean things up.

proofs/fstar/README.md
W95Psp: Let's do that later, this PR is already quite a nightmare, I'd like to make it go through asap!
But I was not aware one could share git hooks actually! How do you do that? :o

W95Psp:
This PR now needs https://github.com/cryspen/libcrux/pull/194.......

proofs/fstar/README.md
franziskuskiefer: Need to be installed individually, but having the code in the repo makes this easier. https://github.com/cryspen/libcrux/tree/main/git-hooks

PR #198, opened 2024-02-03
Merge some changes from extraction to extraction-edited

karthikbhargavan:
This is the first of a few passes to update the F* proofs in `extraction-edited` to match the code now being generated in the the `extraction` directory.

The PR has three kinds of changes:
- Changes to call sites of the `deserialize_public_key` function in `ind_cpa.rs` and `kyber.rs` to slice the public key before passing it to this function. This change makes the F* precondition and post-condition of the resulting function cleaner and more consistent with the previous version; it also makes the Rust code less fragile to future changes.
- Superficial changes to files like `Libcrux.Digest.fsti` that update the code but do not require new proofs
- Propagate recent renamings of `Kyber*` types from `types.rs` to `MlKem*` in the verified F* code and reverify files.

We also check in the hints files to help CI verify the new F* code.


proofs/fstar/extraction-edited/Libcrux.Kem.Kyber.Hash_functions.fst
```diff
@@ -18,12 +18,13 @@ let v_PRF (v_LEN: usize) (input: t_Slice u8) =
...
 
-let v_XOFx4 (v_K: usize) (input: t_Array (t_Array u8 (sz 34)) v_K) =
+let v_XOFx4 v_K (input: t_Array (t_Array u8 (sz 34)) v_K) =
+  assert (v v_K >= 2);
   let out:t_Array (t_Array u8 (sz 840)) v_K =
     Rust_primitives.Hax.repeat (Rust_primitives.Hax.repeat 0uy (sz 840) <: t_Array u8 (sz 840)) v_K
   in
   let out:t_Array (t_Array u8 (sz 840)) v_K =
-    if ~.(Libcrux_platform.simd256_support <: bool) || ~.false
+    if ~.(Libcrux_platform.simd256_support () <: bool) || ~.true
```
W95Psp, line 27: This I think is now `~.(Libcrux_platform.simd256_support () <: bool)` without the `~.true` in `extraction`

W95Psp (approved):

PR #201, opened 2024-02-05
Fix hax cfg issues

W95Psp:
This PR:
 - update F*;
 - makes F* extraction stable across hardware (well, I hope, I cannot try);
 - fixes the Makefiles in `proofs/extraction/fstar` so that they use the correct folder for Libcrux_platform;
 - makes CI check snapshots are up to date;
 - cut CI time for the `hax` job in two by installing Hax from Nix (going from ~12 minutes to ~1 minute)


karthikbhargavan (approved):
Hopefully this unblocks CI. Changes look good.

PR #202, opened 2024-02-19
Kyber: add proofs for `serialize_uncompressed_ring_element`

W95Psp:
This PR:
 - introduces a `BitVecEq` module (we should move that to hax lib at some point, see https://github.com/cryspen/libcrux/issues/207);
 - adds a proof for the last function of the serialize module that will serve as a basis for all other proofs (which are similar) in this module;
 - fixes a regression.


karthikbhargavan (approved):
Looks good but yes, BitVecEq should go to rust_primitives

W95Psp:
(I'm not moving BitVecEq yet, because we will probably need more things in there, and I will discover that while proving the other functions, so I will do the move once everything is proven)

PR #206, opened 2024-02-20
fix(kyber/F*): spec, force subtype on bit vectors

W95Psp:
Type inference sometimes fails with fun. ext. bit vectors: when F* wants a `bit_vec n` but gets a ` bit_vec n'` where `n == n'` but `n` up to some reasoning, F* fails.


W95Psp:
Closing in favor of #202

PR #208, opened 2024-02-22
Updated ml-kem C extraction

franziskuskiefer:
This updates the C extraction of ml-kem.
It also fixes the tests for the extraction and removes a panic in the C code that was not necessary.


karthikbhargavan (approved):
Looks good. Did not look into the tests, I don't know how they are related.

PR #211, opened 2024-02-29
Adapt Kyber to use Init-Absorb-Squeeze* API

karthikbhargavan:
This PR adds the incremental SHA-3 API to libcrux and adapts Kyber to use this incremental API. In particular, `sampling` uses `sample_from_xof` which first samples three SHAKE128 blocks, and then, if needed, it squeezes more blocks from the SHAKE128 state. 

The code has been adapted to work with both Scalar and Simd256 versions of SHA-3.

The code also uses the `opaque_type` attribute to hide the internal implementation of SHAKE-128 state in `digest.rs`


dconnolly:
Neat

franziskuskiefer:
We merged #220 instead

PR #213, opened 2024-03-06
fix(f*/kyber): update for https://github.com/hacspec/hax/pull/546

W95Psp:
Add `f_pre` and `f_post` to traits, following https://github.com/hacspec/hax/pull/546


franziskuskiefer:
Looks like the patch doesn't apply on ci.

franziskuskiefer (approved):
:shipit:

PR #216, opened 2024-03-10
Add X-Wing combiner

raphaelrobert:
Adds [draft-connolly-cfrg-xwing-kem-01](https://www.ietf.org/archive/id/draft-connolly-cfrg-xwing-kem-01.html).


src/kem.rs
```diff
@@ -147,13 +153,44 @@ impl X25519MlKem768Draft00PrivateKey {
     }
 }
 
+/// An X-Wing private key.
+pub struct XWingKemDraft01PrivateKey {
```
franziskuskiefer, line 157: There's a small diff to -02. Can you update to that?

src/kem.rs
```diff
@@ -183,13 +220,41 @@ impl X25519MlKem768Draft00PublicKey {
     }
 }
 
+/// An X-Wing public key.
+pub struct XWingKemDraft01PublicKey {
+    pub pk_m: MlKem768PublicKey,
+    pub pk_x: x25519::PublicKey,
+}
+
+// TODO: Check serialization
```
franziskuskiefer, line 229: Is this something you still want to do?

src/kem.rs
```diff
@@ -445,6 +565,18 @@ impl Ss {
                 out.extend_from_slice(kk.as_ref());
                 out
             }
+            Ss::XWingKemDraft01(ss_m, ss_x, ct_x, pk_x) => {
+                let x1 = r"\./";
```
franziskuskiefer, line 569: This can be updated "In hex XWingLabel is given by 5c2e2f2f5e5c."

franziskuskiefer (changes requested):
Thanks, generally looks good with a few nits.

src/kem.rs
raphaelrobert: This was already addressed

raphaelrobert: This was already addressed

PR #219, opened 2024-03-12
X25519 benchmarks

franziskuskiefer:
More x25519 benchmarks, including more comparisons.


karthikbhargavan (approved):
Lib25519 now builds on mac as well, so we should add instructions for that.
https://lib25519.cr.yp.to/install.html

I will test this PR on Linux and Mac, but overall it looks good.

PR #220, opened 2024-03-12
Use Init-Absorb-Squeeze* API in ML-KEM

franziskuskiefer:
This replaces #211 to avoid unreachable code and panics in the extracted C code.

See https://github.com/cryspen/hacl-packages/pull/453 for the corresponding C code.


karthikbhargavan (approved):
Works well on all my machines. 
Thanks!

PR #224, opened 2024-03-18
Update public key validation

franziskuskiefer:


src/kem/kyber/serialize.rs
```diff
@@ -101,6 +101,53 @@ pub(super) fn deserialize_to_uncompressed_ring_element(serialized: &[u8]) -> Pol
     re
 }
 
+#[inline(always)]
```
karthikbhargavan, line 104: Changes look correct.

Let us add a comment here saying that this function should only be called with "Public" arguments and should never be use with secrets.

karthikbhargavan (approved):

PR #226, opened 2024-04-03
Update ml-kem extraction for NSS

franziskuskiefer:
Update extraction for NSS and some more tests.

* Add early wycheproof tests
* While public key validation is recommended, it is not part of encaps. Hence I made encaps reduce the public key.


karthikbhargavan (approved):
Changes look good. 

Is adding the reduction to encaps impacting performance?

PR #233, opened 2024-04-23
Refactored ML-KEM SIMD scaffolding

xvzcf:
- Broke out SIMD-specific code into a separate `simd` directory
- Created a trait called `Operations` to provide a unified API over the different SIMD bit-widths and architectures
- Removed the `*_int_vec_*` prefixes, suffixes, and affixes and renamed constants


libcrux-ml-kem/src/simd/fallback.rs
```diff
@@ -0,0 +1,437 @@
+//use crate::hax_utils::hax_debug_assert;
+use crate::{
+    arithmetic::*,
+    compress::{compress_ciphertext_coefficient, compress_message_coefficient},
+};
+//use crate::cloop;
```
franziskuskiefer, line 6:
```suggestion
```

libcrux-ml-kem/src/simd/fallback.rs
```diff
@@ -0,0 +1,437 @@
+//use crate::hax_utils::hax_debug_assert;
```
franziskuskiefer, line 1:
```suggestion
```

libcrux-ml-kem/src/simd/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+        decompressed =
+            Self::add_constant(Self::shift_left(decompressed, 1), 1i32 << coefficient_bits);
+        decompressed = Self::shift_right(decompressed, coefficient_bits + 1);
+
+        decompressed
+    }
+}
+
+#[cfg(feature = "simd128")]
+mod fallback;
```
franziskuskiefer, line 99: This is not a fallback, but portable. Let's call it something like that.

libcrux-ml-kem/src/lib.rs
```diff
@@ -12,21 +12,17 @@ pub(crate) mod constants;
...
 mod hash_functions;
 mod ind_cpa;
-mod intvec;
-mod intvec32;
 mod matrix;
 mod ntt;
 mod polynomial;
 mod sampling;
 mod serialize;
+mod simd;
```
franziskuskiefer, line 25: This is not only simd, but polynomial really. It's fine to keep like this because it will need a refactor for runtime detection anyway. But the naming is a little off.

franziskuskiefer (commented):
We should update some names but we can do that later. The F* is the only real issue we should fix.

franziskuskiefer:
This will need more hax changes. We will work on `dev` for now and do this refactoring later.

karthikbhargavan (approved):
A lot of work in this PR, future PRs should be much smaller.

libcrux-ml-kem/src/simd/mod.rs
```diff
@@ -0,0 +1,20 @@
+pub(crate) mod simd_trait;
+
+#[cfg(all(not(feature = "simd128"), not(feature = "simd256")))]
+mod portable;
```
franziskuskiefer, line 4: Portable will always have to be built.
It's ok for now. I'll change that later.

sys/hacl/c/src/Hacl_Hash_SHA3_Scalar.c
```diff
@@ -1158,14 +1158,12 @@ void Hacl_Hash_SHA3_Scalar_sha3_224(uint8_t *output, uint8_t *input, uint32_t in
     uint64_t c = Hacl_Impl_SHA3_Vec_keccak_rndc[i0];
     s[0U] = s[0U] ^ c;
   }
+  uint8_t hbuf[200U] = { 0U };
```
franziskuskiefer: What are these changes? They shouldn't be here.

franziskuskiefer (approved):
There's a lot here that needs cleanup. But we'll do that in follow-ups.
I only reverted the sha3 changes. If that's something you want @karthikbhargavan, do them separately.

karthikbhargavan:
ugh, was a local experiment. should not have been pushed.

PR #243, opened 2024-04-30
simd ml kem generic

### LINKED ISSUE #234: Generic ml-kem implementation
Make the implementation generic over the polynomial state.

franziskuskiefer:
This PR makes everything generic over the vector type.
There's no runtime detection or anything else that's more intelligent. It is only one step towards making the code flexible over the vector implementation.

Fixes #234


franziskuskiefer:
We'll pull out crates instead in #248

PR #244, opened 2024-04-30
First shot at AVX2 implementations for ciphertext coefficient compression and the remaining NTT functions.

xvzcf:


franziskuskiefer:
There's a fmt issue. Otherwise I'd suggest to get this in.

franziskuskiefer (approved):
Let's get this in

PR #248, opened 2024-05-03
Separate simd crates for ml-kem vectors

franziskuskiefer:
This PR pulls out the vector implementations for avx2 and aarch64 from ml-kem.
This needs more cleanup and we may want to abstract more. But this sets up the structure and adds runtime detection.


karthikbhargavan (approved):
Tested on my mac. I also moved ntt code into polynomial, eliminating yet another module.

Issue #250, opened 2024-05-06
16-bit version of ml-kem


franziskuskiefer:
- [x] Move code from #251 into the new structure
- [x] Use 16 field elements instead of 8
- [x] Implement avx2 functions

franziskuskiefer:
This has been merged to dev

PR #251, opened 2024-05-06
16-bit version of ml-kem

karthikbhargavan:
This PR switches the polynomial representation from i32 to i16.
The changes have been propagated to the portable and simd128 version but not to simd256, which pass tests.
Performance for portable is unchanged, and perf for simd128 improves by 15-20%

The only place where the switch required changing the expected range constraints was in inv-ntt.
The new code now carefully documents the expected ranges.

This PR addresses #250


franziskuskiefer:
This was done as part of #256

PR #256, opened 2024-05-06
Use 16 bit representatives for Kyber SIMD instead of 32 bit representatives

xvzcf:


karthikbhargavan (commented):
The portable and aarch64 versions work on my machine and appear to be good. 
Is the AVX2 code going to appear in a subsequent PR?

franziskuskiefer:
So this was working but broke with the latest commit. Let's get that out again and merge this PR.

karthikbhargavan:
So what happened here was that it was working for 8-elements-per-vector for portable and simd128, but not for simd256. 
So the choice was between implementing 8-per-vector for simd256 (which seemed to require a non-trivial amount of work) and then switching to 16-per-vector, or bypassing this redundant step and going straight for 16-per-vector. This PR is now taking the latter path, so Goutam put this PR back into a draft (not ready for review) while he switches to 16-per-vector.

franziskuskiefer (approved):

PR #257, opened 2024-05-07
Libcrux optimize decap

protz:
In addition to Karthik's API improvements, this PR contains a few more goodies on top, notably:
- generating code for all three variants of kyber, but in separate files
- putting all of the shared code in a header where everything is marked as static inline
- updated config.yaml for all of the above
- addition of a few well-chosen type declarations to generate prettier C code

I think this would be worth merging now.

PR on the hacl-packages side forthcoming


protz:
This required quite a few changes in upstream Eurydice to support generating all variants all at once, but these have now been merged.

franziskuskiefer:
Looks like CI is failing all over the place.

franziskuskiefer (approved):
Thanks, let's get this in.

PR #258, opened 2024-05-08
SIMD rejection sampling

karthikbhargavan:
This PR restores the SIMD version of rejection sampling.
Currently this works for portable and AArch64, and is ready for an AVX2 implementation.
(It gives a 10% improvement for AArch64.)


libcrux-ml-kem/src/sampling.rs
```diff
@@ -46,37 +43,39 @@ use crate::{
...
 /// <https://csrc.nist.gov/pubs/fips/203/ipd>.
-fn sample_from_uniform_distribution_next<const K: usize, const N: usize>(
+#[inline(always)]
+fn sample_from_uniform_distribution_next<Vector: Operations, const K: usize, const N: usize>(
     randomness: [[u8; N]; K],
     sampled_coefficients: &mut [usize; K],
     out: &mut [[i16; 256]; K],
 ) -> bool {
-    let mut done = true;
+    // Would be great to trigger auto-vectorization or at least loop unrolling here
```
franziskuskiefer, line 52: In this case, this would need to the platform dependent crates. Do you want to do that here?

franziskuskiefer: Right, but for the compiler to do it with all the hardware instructions, it needs to move. Leaving this for later.

franziskuskiefer (approved):

PR #259, opened 2024-05-08
ml-kem aarch64 move all unsafe code into neon mod

franziskuskiefer:
This PR pulls out all the unsafe code in the aarch ml-kem crate into a single module.


karthikbhargavan (approved):
Looks good. Would be nice to evolve these low-level APIs over time  to something like:
https://doc.rust-lang.org/std/simd/index.html

PR #260, opened 2024-05-08
Implementations of some AVX2 functions on 16-bit representatives

xvzcf:


franziskuskiefer (approved):
Thanks. Let's get this in.

PR #270, opened 2024-05-13
AVX2 implementations of Barrett reduction, the NTT, and deserialization for some parameters.

xvzcf:


franziskuskiefer (approved):
Thanks, let's get this in.

PR #276, opened 2024-05-14
AVX2 implementation of Kyber rejection sampling.

xvzcf:


xvzcf:
Subsumed by https://github.com/cryspen/libcrux/pull/278

PR #279, opened 2024-05-16
Documentation and cleanup of the Kyber avx2 code.

xvzcf:


karthikbhargavan (changes requested):
I expect that all modules will now use "intrinsics" instead of directly using core::arch::x86_64?

xvzcf:
> I expect that all modules will now use "intrinsics" instead of directly using core::arch::x86_64?

Yup.

PR #280, opened 2024-05-16
Feature detection and cleanup (Ml-KEM/SHA3)

franziskuskiefer:
This PR
- adds runtime detection for hash functions
- cleans up the sha3 implementation


karthikbhargavan (approved):
Changes look good. I wish we could have hidden the different implementations inside libcrux-sha3 rather than polluting hash_functions.rs but I don't think that's possible with our structure and optimization constraints.

franziskuskiefer:
> I wish we could have hidden the different implementations inside libcrux-sha3 rather than polluting hash_functions.rs but I don't think that's possible with our structure and optimization constraints.

We can, I'm not sure it's the right structure though. It's pushing that logic into the sha3 crate. I don't have a strong opinion, both is doable and has pros and cons.

PR #283, opened 2024-05-16
More safe intrinsic wrappers

xvzcf:


polynomials-avx2/src/intrinsics.rs
```diff
@@ -67,6 +248,12 @@ pub(crate) fn mm256_slli_epi32<const SHIFT_BY: i32>(vector: __m256i) -> __m256i
     unsafe { _mm256_slli_epi32(vector, SHIFT_BY) }
 }
 
+pub(crate) fn mm_shuffle_epi8(vector: __m128i, control: __m128i) -> __m128i {
```
prosecco, line 251: Are there any preconditions on `control` we should be debug-asserting here?

prosecco (approved):

PR #288, opened 2024-05-22
Refresh patches, followup for hax PR hacspec/hax#676

W95Psp:
This PR regenerates patches following hacspec/hax#676.
Once again, either this CI or Hax' CI should be bypassed.

### DIFF
--- a/proofs/fstar/patches.sh
+++ b/proofs/fstar/patches.sh
@@ -71,8 +71,12 @@ case $1 in
 
             
         )
-        mv "$TEMPDIR/extraction-edited.patch" extraction-edited.patch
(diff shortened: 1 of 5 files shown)


proofs/fstar/patches.sh
```diff
@@ -71,8 +71,12 @@ case $1 in
 
             
         )
-        mv "$TEMPDIR/extraction-edited.patch" extraction-edited.patch
-        mv "$TEMPDIR/extraction-secret-independent.patch" extraction-secret-independent.patch
+        for patch in "extraction-edited.patch" "extraction-secret-independent.patch"; do
```
franziskuskiefer, line 74: 😬  We should really get rid of the patch files ...

franziskuskiefer (approved):
Thanks. I'll merge this.

PR #289, opened 2024-05-23
Added scaffolding to get work started on ML-DSA.

xvzcf:


ml-dsa/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "ml-dsa"
```
franziskuskiefer, line 2: That's squatted by Tony already. Let's always prefix with `libcrux`.

```suggestion
name = "libcrux-ml-dsa"
```

franziskuskiefer (approved):
Let's get started :)

PR #291, opened 2024-05-27
ML-DSA Keygen: Some functions related to the vector t.

xvzcf:


franziskuskiefer (approved):
Thanks

.github/workflows/mlkem.yml
```diff
@@ -36,7 +36,7 @@ jobs:
     defaults:
       run:
         shell: bash
-        working-directory: libcrux-ml-kem
+        working-directory: libcrux-ml-dsa
```
franziskuskiefer: This looks wrong. You probably wanted to do this in `mldsa.yml` 🙂

franziskuskiefer (approved):
lgtm with the comment addressed.

.github/workflows/mlkem.yml
xvzcf: Ugh, sorry about that. Fixing

PR #293, opened 2024-05-31
More Kyber AVX2-serialization documentation.

xvzcf:


franziskuskiefer (changes requested):
Can you rebase this onto #296 to avoid conflicts?

xvzcf:
> Can you rebase this onto #296 to avoid conflicts?

Done.

franziskuskiefer (approved):
Thanks.
Let's not move code around in future. We're starting with the proofs.

PR #300, opened 2024-06-05
Add Nix files and commit `Cargo.lock`

pnmadelaine:
Add a Nix expression to build `ml-kem`


pnmadelaine:
I disabled LTO in the Nix build as it causes the build to fail. Am I missing something, is there a specific toolchain to use? I am using `gcc` and `ld`.

pnmadelaine:
using mold instead of ld seems to fix the issue!

libcrux-ml-kem/c/CMakeLists.txt
```diff
@@ -108,13 +108,7 @@ target_link_libraries(sha3_test PRIVATE
 
 # --- Benchmarks
 
-FetchContent_Populate(benchmark
```
franziskuskiefer, line 111: Where does this change come from? It doesn't look right.

franziskuskiefer (changes requested):
Thanks, generally looks good except for that one change.

pnmadelaine:
@franziskuskiefer I'll just continue the conversation here as github seems to double all my comments on the code (?)

I think you meant using `fetch_package` was not right because it required installing `benchmark` manually? I used `FetchContent_Declare` instead, is that okay with you?

W95Psp:
(quoting myself from https://github.com/cryspen/bertie/pull/118, since it's the same concern)
> I see no point in not committing flake.lock, of we don't commit flake.lock, we should drop the nix expressions altogether. The nix job you added will check the nix expressions are up to date as code changes, that's all we need imo. Locked flakes are especially useful to make sure we'll be able to build the project in the future.

pnmadelaine:
converting to draft as this is currently broken after updating

protz (approved):

franziskuskiefer (approved):

PR #305, opened 2024-06-10
Pull out `hkdf` and `hmac` modules as standalone crates

jschneider-bensch:
This PR pulls out the `libcrux::hkdf` and `libcrux::hmac` modules, as well as their dependencies in the `libcrux::hacl` module into their own crates:
-  `libcrux-hkdf` which only depends on `sys/hacl` and 
-  `libcrux-hmac`, which also depends on the new `libcrux-hkdf` crate

I've also replaced these modules in `libcrux` with a re-export of the standalone crates' APIs and pruned the `libcrux::hacl` module of the dead code.


franziskuskiefer (approved):
Thanks, looks good to me.
We should add the wycheproofs tests to the crates. But looks like they're not there yet. Can you file a follow up?

jschneider-bensch:
> Thanks, looks good to me. We should add the wycheproofs tests to the crates. But looks like they're not there yet. Can you file a follow up?

I've filed a follow up here: #306

PR #308, opened 2024-06-11
Making ML-KEM Extract Again

karthikbhargavan:
This PR does some changes to make the new ML-KEM code extract to F*.
This does not lax typecheck yet.


franziskuskiefer:
I checked performance and couldn't measure a difference in Rust or C.
I updated the C extraction with AeneasVerif/eurydice#25 and added a small helper for build after getting tired of calling cmake directly all the time.

The hax CI needs to get fixed and we need to re-enable the extraction of this crate on CI.

franziskuskiefer (approved):

PR #309, opened 2024-06-11
Laying groundwork for implementing ML-DSA signing.

xvzcf:


libcrux-ml-dsa/src/deserialize.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_deserialize_to_mask_when_gamma1_is_2_pow_17() {
+        let bytes = [
```
franziskuskiefer, line 81: Please add a description where this is coming from or instructions how to generate it.

franziskuskiefer (approved):
lgtm with one nit

PR #315, opened 2024-06-12
CI: update `nix.yml`

pnmadelaine:
- run `nix.yml` on pull requests
- use nix cache


pnmadelaine:
I think it is better to merge this first.

The Nix job is going to fail anyway until the interaction with Eurydice is fixed.
With this at least it is going to fail faster! :)

franziskuskiefer (approved):

PR #320, opened 2024-06-17
ML-DSA: Signing works across all parameter sets.

xvzcf:
- The contents of the two files `serialize.rs` and `deserialize.rs` have been moved into a module called `encoding`, serialization and deserialization functions are now grouped together according to what they act upon.
- The KATs have been updating to also store a `signing_randomness` value that is used during signing
- I now use tuple structs to hold the signing key, verification key, and signature, as well as another struct that groups the signing and verification key tuple structs


libcrux-ml-dsa/src/arithmetic.rs
```diff
@@ -10,17 +10,60 @@ impl PolynomialRingElement {
         // FIXME: hax issue, 256 is COEFFICIENTS_IN_RING_ELEMENT
         coefficients: [0i32; 256],
     };
+
+    pub(crate) fn add(&self, rhs: &Self) -> Self {
```
franziskuskiefer, line 14: Can all these small functions get `inline`s? Just to make sure the compiler is doing the right thing.

libcrux-ml-dsa/src/arithmetic.rs
```diff
@@ -10,17 +10,60 @@ impl PolynomialRingElement {
...
+        // It is ok to leak which coefficient violates the bound since
+        // the probability for each coefficient is independent of secret
+        // data but we must not leak the sign of the centralized representative.
+        for coefficient in self.coefficients.iter() {
+            // Normalize the coefficient
+            let sign = coefficient >> 31;
+            let normalized = coefficient - (sign & (2 * coefficient));
+
+            if normalized >= value {
+                return true;
```
franziskuskiefer, line 48: I don't think this will go through hax.

libcrux-ml-dsa/src/arithmetic.rs
```diff
@@ -10,17 +10,60 @@ impl PolynomialRingElement {
...
-) -> PolynomialRingElement {
-    for i in 0..lhs.coefficients.len() {
-        lhs.coefficients[i] += rhs.coefficients[i];
+pub(crate) fn vector_infinity_norm_exceeds<const DIMENSION: usize>(
+    vector: [PolynomialRingElement; DIMENSION],
+    value: i32,
+) -> bool {
+    for i in 0..DIMENSION {
+        if vector[i].infinity_norm_exceeds(value) {
+            return true;
```
franziskuskiefer, line 62: Same here.

libcrux-ml-dsa/src/encoding/error.rs
franziskuskiefer, line 1: Is there a reason you call these fines encoding/decoding now instead (de)serialize?

franziskuskiefer (approved):
lgtm with a few nits

libcrux-ml-dsa/src/encoding/error.rs
xvzcf: I've now grouped the functions based on which ring element (the error, the commitment) they operate on. I could've named the top level module `serialize` instead of `encoding`, but I thought that'd be less clear because:

- The module contains both serialization and deserialization functions
- `encoding::error::serialize` looks nicer to me than `serialize::error::serialize` (not to mention `encoding::error::deserialize`)

libcrux-ml-dsa/src/arithmetic.rs
xvzcf: For both this and the other loop, I used a naive approach with a comment indicating we could revisit if performance is impacted.

franziskuskiefer (approved):
Thanks

libcrux-ml-dsa/src/encoding/error.rs
franziskuskiefer: K, I don't have a strong opinion either way. But we'll want to probably unify it at some point

Issue #302, opened 2024-06-10
Plan pqcp


franziskuskiefer:
This issue tracks the pqcp packaging and release policies.

franziskuskiefer:
- APIs (pqcp feature?)
- Update script

jschneider-bensch:
- `libcrux-ml-kem`'s `Cargo.toml` should be made independent from the workspace
- should the PQCP repo contain the necessary dependency crates (`libcrux-intrinsics`, `libcrux-sha3`, `libcrux-platform`) as well?

jschneider-bensch:
The following `Cargo.toml` e.g. has the package metadata and release profile settings pulled down from the workspace, as well as replacing local path dependencies by the `libcrux` repo link. (This works as a first step, but in the code package there probably shouldn't be a dependency on `libcrux-dev`, right?)

```toml
[package]
name = "libcrux-ml-kem"
version = "0.0.2-pre.2"
authors = ["Cryspen"]
license = "Apache-2.0"
homepage = "https://github.com/cryspen/libcrux"
edition = "2021"
repository = "https://github.com/cryspen/libcrux"
readme = "Readme.md"
exclude = ["/tests", "/implementation_notes.pdf"]

[dependencies]
rand_core = { version = "0.6" }
libcrux-platform = { git = "https://github.com/cryspen/libcrux", version = "0.0.2-pre.2", branch = "dev" }
libcrux-sha3 = { git = "https://github.com/cryspen/libcrux", version = "0.0.2-pre.2", branch = "dev" }
libcrux-intrinsics = { git = "https://github.com/cryspen/libcrux", version = "0.0.2-pre.2", branch = "dev" }

# This is only required for verification.
# The hax config is set by the hax toolchain.
[target.'cfg(hax)'.dependencies]
hax-lib = { git = "https://github.com/hacspec/hax/" }

[features]
default = ["std"]
simd128 = ["libcrux-sha3/simd128"]
simd256 = ["libcrux-sha3/simd256"]
tests = []                         # Test utilities. DO NOT USE.
std = []

[dev-dependencies]
rand = { version = "0.8" }
serde_json = { version = "1.0" }
serde = { version = "1.0", features = ["derive"] }
hex = { version = "0.4.3", features = ["serde"] }
criterion = "0.5"
libcrux-ml-kem = { path = ".", features = ["tests"] }

[[bench]]
name = "ml-kem"
harness = false

[profile.release]
lto = "fat"
codegen-units = 1
[...]

PR #323, opened 2024-06-18
PQ code package update script

jschneider-bensch:
This is part of the planned workflow for packaging `libcrux-ml-kem` in the PQ code package. (cf. #302)

The script copies files from an allow-list over to the PQCP repo submodule directory. It then runs tests using the default features and upon success creates a signed-off commit in the submodule on a separate branch named after the `libcrux` revision. Pushing needs to be done manually.
The allow-list does not include e.g. everything related to C code extraction.


jschneider-bensch:
Waiting on #327 and #329 for further changes before marking this ready for review.

jschneider-bensch:
I've updated the custom `Cargo.toml` to include the feature flags from #327, #333 and those that will probably land with #329.

libcrux-ml-kem/pqcp/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "libcrux-ml-kem"
+version = "0.0.2-pre.2"
+authors = ["Cryspen"]
+license = "Apache-2.0"
+homepage = "https://github.com/cryspen/libcrux"
+edition = "2021"
+repository = "https://github.com/cryspen/libcrux"
+readme = "Readme.md"
+exclude = ["/tests"]
```
franziskuskiefer, line 10: What about the other excludes?

libcrux-ml-kem/pqcp/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "libcrux-ml-kem"
+version = "0.0.2-pre.2"
```
franziskuskiefer, line 3: Would it be easier to copy the existing one and `sed` these things?

libcrux-ml-kem/update-pqcp.sh
```diff
@@ -0,0 +1,31 @@
...
+while read -r item; do
+    cp -r "$item" "$TARGET_DIRECTORY"
+done < "$ALLOWLIST"
+
+cd "$TARGET_DIRECTORY"
+cargo test
+cargo clean
+
+git add .
+git commit -sm "mlkem-rust-libcrux PQ code packages update ($(date))"
```
franziskuskiefer, line 31: This should probably get the git revision from the libcrux repo as well.

franziskuskiefer (dismissed):
Thanks, generally looks good with some comments and questions.

I'm also wondering if we should tie the two a little closer togther. We could make libcrux a git submodule for the pqcp repo and then pull it in on that side? Just an idea, trying to see how we can make this as seamless and hassle-free as possible.

libcrux-ml-kem/pqcp/Cargo.toml
jschneider-bensch: The other excludes I don't include in `pqcp/allowlist.txt`, so they are not copied over to the PQCA repository. I was wondering about the value of this field, though, in the context of the PQ code package, since we will not re-publish from there anyway, right?

franziskuskiefer: Good point. I was thinking about keeping things in sync. It shouldn't matter indeed.

jschneider-bensch: The new, generated `Cargo.toml` does not include this field.

libcrux-ml-kem/pqcp/Cargo.toml
jschneider-bensch: I've included a Python script that builds a `Cargo.toml` from the workspace and crate `Cargo.toml`s. This way, these things should be in sync always.

libcrux-ml-kem/pqcp/update-pqcp.sh
```diff
@@ -0,0 +1,47 @@
...
+libcrux_ml_kem_crate="$libcrux_root/libcrux-ml-kem"
+allowlist="$libcrux_ml_kem_crate/pqcp/allowlist.txt"
+pqcp_repo="$libcrux_ml_kem_crate/pqcp/repo"
+
+cd "$libcrux_ml_kem_crate"
+while read -r item; do
+    cp -r "$item" "$pqcp_repo"
+done < "$allowlist"
+
+python "$libcrux_ml_kem_crate/pqcp/cargo.py" "$libcrux_root"
```
franziskuskiefer, line 38: `python` doesn't exist on many systems. You probably want `python3`.

libcrux-ml-kem/pqcp/update-pqcp.sh
```diff
@@ -0,0 +1,47 @@
...
+
+cd "$pqcp_repo"
+cargo test
+cargo clean
+
+message="$(printf "mlkem-rust-libcrux PQ code packages update\n\n$(date): libcrux revision $(git -C $libcrux_root rev-parse HEAD)")"
+git switch -c "libcrux-ml-kem-$(git -C $libcrux_root rev-parse HEAD)"
```
franziskuskiefer, line 45: This moved the libcrux repo over to this revision. Something doesn't seem quite right here.

franziskuskiefer (changes requested):

franziskuskiefer:
We finally need to get back to this. This may be a little stale. But still the thing to do.

PR #324, opened 2024-06-18
mlkem release prep

franziskuskiefer:


libcrux-intrinsics/Cargo.toml
```diff
@@ -9,12 +9,6 @@ repository.workspace = true
 readme.workspace = true
```
jschneider-bensch, line 10:
```suggestion
exclude = [ "/proofs" ]

```
We want to exclude proofs from the published crates, also on the other ones.

libcrux-ml-kem/Cargo.toml
jschneider-bensch, line 1: Why do we want to exclude the `/tests` directory?

jschneider-bensch (changes requested):
This looks good to me, but we don't want to publish the `proof` directories with the crates.

libcrux-ml-kem/Cargo.toml
franziskuskiefer: It's huge, and I'm not sure people really use the tests of a published crate.

jschneider-bensch (approved):
Thanks, that resolves my questions.

PR #328, opened 2024-06-20
Restructure Vector code in ML-KEM to split functionality

karthikbhargavan:
This PR makes the structure of the 3 vector implementations uniform by using the same module names for AVX2, Neon, and Portable. It also breaks up the monolithic portable code in `vector.rs` into multiple modules.


karthikbhargavan:
I added a drive-by fix for a C compilation error (implicit conversion from 'int' to 'short' changes value from 32768 to -32768)

franziskuskiefer (changes requested):
lgtm, can you update the C code with this to make sure this is working?

karthikbhargavan:
I regenerated and tested the C code, but I would love it if someone else could also check that this works (once CI is happy)

franziskuskiefer (approved):
@karthikbhargavan there's a conflict. Can you resolve that? Then we can get this merged.

I tested the C code. It works fine on arm-based macOS and x64. CI says the same.

You committed the neon files again though. They don't do anything, but we also don't use them.

PR #331, opened 2024-06-24
ml dsa verify

### LINKED ISSUE #274: ML-DSA 65 Ref Implementation
Following the style from ML-KEM, implement ML-DSA.
This issue is about the portable implementation. But since we're doing a SIMD optimised version, it should be generic not only over the key size but also SIMD.

- #292 
- #290 
- #289 
- #291

franziskuskiefer:
Closes #274


karthikbhargavan (approved):
This all looks very clean and nice.

PR #333, opened 2024-06-24
Add features for ml-kem variants

franziskuskiefer:
This PR

* adds ml-kem variants features (`mlkem768` etc) to allow enabling only a single variant
* adds separate extraction for mlkem768 in a header only mode with `cg.yaml`
* makes c extraction in `c.sh` more flexible to use all these things

The Rust code changes are
* use a simple for loop in rejection sampling
* drop strings from `unimplemented` macros in sha3 to avoid extraction them to C
* disable the `multiplexing` module when extracting C


karthikbhargavan (approved):
Looks good. It works on my machine as well.
Shall we document the various features and use-cases in `libcrux-ml-kem/README.md`?

PR #335, opened 2024-06-26
Workflow to dismiss stale reviews

franziskuskiefer:
Dismiss old reviews on PRs to make GH more usable.

### DIFF
--- a/.github/workflows/reviews.yml
+++ b/.github/workflows/reviews.yml
@@ -1,0 +1,12 @@
+name: Dismiss old reviews when re-requested
+
+on:
+  pull_request:
+    types: [review_requested]
+  workflow_dispatch:
+
+jobs:
+  review_requested:
+    permissions:
+      pull-requests: write
+    uses: cryspen/actions/.github/workflows/dismiss-old-reviews.yml@main


.github/workflows/reviews.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  pull_request:
+    types: [review_requested]
+  workflow_dispatch:
+
+jobs:
+  review_requested:
+    permissions:
+      pull-requests: write
+    uses: cryspen/actions/.github/workflows/dismiss-old-reviews.yml@main
```
jschneider-bensch, line 12: Just to verify my understanding is correct, this action dismisses old reviews of exactly those reviewers from which a review has been re-requested. If there are other reviewers from which a review has not been re-requested, their reviews will not be dismissed, right?

jschneider-bensch (approved):
Thanks, this looks good to me!

.github/workflows/reviews.yml
franziskuskiefer: Yes, that's the idea. We'll need to see if it does the right thing 😄 .

PR #342, opened 2024-07-01
Simplify config by relying on the fact that the various variants are now macro-generated into their own modules

protz:


franziskuskiefer (approved):
Thanks, config looks much cleaner.
I did another refresh of the code.

PR #343, opened 2024-07-02
More benchmarks and ct

franziskuskiefer:


libcrux-ml-kem/src/constant_time_ops.rs
```diff
@@ -1,38 +1,47 @@
 use crate::constants::SHARED_SECRET_SIZE;
 use crate::hax_utils::hax_debug_assert;
 
+// These are crude attempts to prevent LLVM from optimizing away the code in this
+// module. This is not guaranteed to work but at the time of writing, achieved
+// its goals.
+// `read_volatile` could be used as well but seems unnecessary at this point in
+// time.
 // Examine the output that LLVM produces for this code from time to time to ensure
```
jschneider-bensch, line 9: This is a tangent, but what would be necessary to automate this check?

jschneider-bensch (commented):
The Rust code looks good to me, but the C build seems to fail?

franziskuskiefer:
> The Rust code looks good to me, but the C build seems to fail?

Ah, `select` is a function the stdlib that conflicts. I'll change the name ...

jschneider-bensch (approved):

PR #347, opened 2024-07-03
Drop `libcrux` dependency in `libcrux-ecdh`

jschneider-bensch:
This PR gets rid of the `libcrux` dependency for testing in `libcrux-ecdh`. To do this, I've replaced the `libcrux` Drbg with `OsRng`, that was previously only used for the `wasm32` target.


franziskuskiefer (approved):
Thanks!

PR #348, opened 2024-07-03
Remove `libcrux` dependency in `libcrux-kem`

jschneider-bensch:
Same procedure as #347, replacing `libcrux` `Drbg` with `OsRng`.

### DIFF
--- a/libcrux-kem/Cargo.toml
+++ b/libcrux-kem/Cargo.toml
@@ -25,5 +25,4 @@ pre-verification = ["libcrux-ml-kem/pre-verification"]
 
 [dev-dependencies]
 libcrux-kem = { version = "0.0.2-alpha.1", path = "./", features = ["tests"] }
-rand_core = { version = "0.6" }
 hex = { version = "0.4.3", features = ["serde"] }
--- a/libcrux-kem/Cargo.toml
+++ b/libcrux-kem/Cargo.toml
@@ -8,6 +8,7 @@ edition.workspace = true
 repository.workspace = true
 readme.workspace = true
 description = "Libcrux KEM implementation"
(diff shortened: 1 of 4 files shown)


libcrux-kem/Cargo.toml
```diff
@@ -25,5 +25,5 @@ pre-verification = ["libcrux-ml-kem/pre-verification"]
 
 [dev-dependencies]
 libcrux-kem = { version = "0.0.2-alpha.1", path = "./", features = ["tests"] }
-libcrux = { version = "0.0.2-alpha.1", path = "../", features = ["rand"] }
+rand_core = { version = "0.6" }
```
franziskuskiefer, line 28: No need for this if you use `rand::rngs::OsRng` instead.

franziskuskiefer (changes requested):

libcrux-kem/Cargo.toml
jschneider-bensch: You' right, thanks! :+1:

libcrux-kem/Cargo.toml
```diff
@@ -25,5 +26,4 @@ pre-verification = ["libcrux-ml-kem/pre-verification"]
 
 [dev-dependencies]
 libcrux-kem = { version = "0.0.2-alpha.1", path = "./", features = ["tests"] }
-libcrux = { version = "0.0.2-alpha.1", path = "../", features = ["rand"] }
 hex = { version = "0.4.3", features = ["serde"] }
```
franziskuskiefer, line 29: There shouldn't be a version here. That prevents it from being released. It should always take the current version.

franziskuskiefer (approved):
Thanks!

PR #350, opened 2024-07-03
release cleanup 0.0.2-alpha.1

franziskuskiefer:
Some more cleanup for releasing all crates in this version.
I disable some crates that we don't want to publish right now.


Cargo.toml
```diff
@@ -42,7 +42,7 @@ repository.workspace = true
 readme.workspace = true
 documentation = "https://docs.rs/libcrux/"
 description = "The Formally Verified Cryptography Library"
-exclude = ["/tests", "/specs"]
```
jschneider-bensch, line 45: Is there a need to exclude other things like the `wasm-demo`, `fuzz` and `formal_verification` directories, or things like `add-c-kyber-to-nss.py`? Or is that not included for some reason by default?

jschneider-bensch (approved):
Thanks, looks good to me!

PR #353, opened 2024-07-04
Update header only export

franziskuskiefer:


karthikbhargavan (approved):
Works on my machine. Will prioritize resolving https://github.com/cryspen/libcrux/issues/368 so that the script-aided attributes are no longer needed.

PR #354, opened 2024-07-04
initial sha3 readme

franziskuskiefer:
A readme for sha3 with some ideas for badges.


jschneider-bensch (approved):
I think this is good as a minimal starting point.
I like the use of badges for verification status! Perhaps we can add a legend of sorts to the top-level README?

franziskuskiefer:
> Perhaps we can add a legend of sorts to the top-level README?

I added that to #355

PR #367, opened 2024-07-07
Setting up scaffolding for vectorization.

xvzcf:
I moved the code for `PolynomialRingElement` into `polynomial.rs`, and in order to facilitate the transition towards a vectorized codebase, I added another struct to polynomial.rs called `VectorPolynomialRingElement`:

`VectorPolynomialRingElement` contains functions to convert to and from a `PolynomialRingElement`. This way I can incrementally make changes to the codebase; eventually `PolynomialRingElement` will be removed and `VectorPolynomialRingElement` will be take its place. An example of how this works can be seen [here](https://github.com/cryspen/libcrux/compare/main...goutam/ml-dsa-vec-setup?expand=1#diff-3d667db8aecf82b13428fd3d918d06bb0e7a8b76b8c784e8a3ae3af0873a5adcR126).


karthikbhargavan (approved):
The restructure looks good. The real test will be when we add a new implementation of the ring element.

PR #374, opened 2024-07-08
Make ml-kem modules panic-free

mamonet:
This PR is still WIP. The purpose is to have all modules in `libcrux-ml-kem/src` panic-free so the extracted F* files are verifable.

Yet some assumptions made to pass the verification process are slippery and supposed to be revised.

The modules are tested already to pass lax-check.


franziskuskiefer:
@mamonet please close this if it's not actively being worked on.

PR #375, opened 2024-07-08
ML_DSA: ntt.rs now has a vectorized structure.

xvzcf:


franziskuskiefer (approved):
lgtm, I assume you move other parts in arithmetic later? Or do you first want to vectorise what's in there now and then do more later if necessary, that's fine too.

xvzcf:
> lgtm, I assume you move other parts in arithmetic later? Or do you first want to vectorise what's in there now and then do more later if necessary, that's fine too.

I already starting moving in the arithmetic, the plan is to set up a vectorized structure for file and then start filling out the AVX2 code and what not after.

PR #384, opened 2024-07-10
Run benchmarks on merge queue, or manually

jschneider-bensch:
This PR introduces `merge_queue` triggers for all jobs that are run on PRs and makes it so that benchmarks are only run on the `merge_queue` event, or when manually triggered.

### DIFF
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -57,5 +57,5 @@ jobs:
 
       - name: 🏃🏻‍♀️ Benchmark
         run: ./build/ml_kem_bench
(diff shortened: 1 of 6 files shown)


franziskuskiefer (dismissed):
Generally lgtm, but the `ML-KEM / benchmarks` are still enabled. Let's move them to the merge queue as well.

Also, I enabled the merge queue. So we'll see if this works.

jschneider-bensch:
Sorry, I missed that somehow.

franziskuskiefer (approved):
Thanks.

I also moved everything over to the new rulesets. Let's see if that's working.

PR #386, opened 2024-07-10
Report merge queue status

jschneider-bensch:
I've also noticed that I previously didn't include the specs workflow on the queue, which I've done now.

### DIFF
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -59,3 +59,16 @@ jobs:
         run: ./build/ml_kem_bench
         if: ${{ matrix.os != 'windows-latest' && (github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch') }}
 
+  mq_status:
+    if: |
+      always() &&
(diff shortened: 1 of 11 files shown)


.github/workflows/c.yml
```diff
@@ -59,3 +59,16 @@ jobs:
         run: ./build/ml_kem_bench
         if: ${{ matrix.os != 'windows-latest' && (github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch') }}
 
+  mq_status:
```
franziskuskiefer, line 62: Can you move this out to the actions repo to make it re-usable and avoid duplicating the code?

franziskuskiefer (changes requested):
Do you want to re-run all the build on the merge queue? Looks like that's what you require now. That seems like a little much. Only the benchmarks would be enough I think. Having a build or two may be nice as well, but I'd be fine without since we need to figure out which build would make sense.

franziskuskiefer (approved):

Issue #301, opened 2024-06-06
Make libcrux_ml_kem lax-check in F*


mamonet:
Make sure the extracted F* files of libcrux_ml_kem pass lax-check.
This work is being tracked in branches `dev_ml_kem_hax` on libcrux and on hax.
* https://github.com/cryspen/libcrux/tree/dev_ml_kem_lax/libcrux-ml-kem/proofs/fstar/extraction
* https://github.com/hacspec/hax/tree/dev_ml_kem_lax/proof-libs/fstar

("*" indicates that some hax changes, not yet upstreamed, are required to make the module pass)

- [x] Libcrux_ml_kem.Constant_time_ops.fst
- [x] Libcrux_ml_kem.Constant_time_ops.fsti
- [x] Libcrux_ml_kem.Constants.fsti
- [x] Libcrux_ml_kem.Hash_functions.Avx2.fsti
- [x] Libcrux_ml_kem.Hash_functions.Neon.fsti
- [x] Libcrux_ml_kem.Hash_functions.Portable.fsti
- [x] Libcrux_ml_kem.Hash_functions.fsti
- [x] Libcrux_ml_kem.Ind_cca.Instantiations.Avx2.fst
- [x] Libcrux_ml_kem.Ind_cca.Instantiations.Avx2.fsti
- [x] Libcrux_ml_kem.Ind_cca.Instantiations.Portable.fst
- [x] Libcrux_ml_kem.Ind_cca.Instantiations.Portable.fsti
- [x] Libcrux_ml_kem.Ind_cca.Multiplexing.fst
- [x] Libcrux_ml_kem.Ind_cca.Multiplexing.fsti
- [x] Libcrux_ml_kem.Ind_cca.fst*
- [x] Libcrux_ml_kem.Ind_cca.fsti
- [x] Libcrux_ml_kem.Ind_cpa.fst*
- [x] Libcrux_ml_kem.Ind_cpa.fsti
- [x] Libcrux_ml_kem.Invert_ntt.fst*
- [x] Libcrux_ml_kem.Invert_ntt.fsti
- [x] Libcrux_ml_kem.Matrix.fst
- [x] Libcrux_ml_kem.Matrix.fsti
- [x] Libcrux_ml_kem.Mlkem1024.Avx2.fst
- [x] Libcrux_ml_kem.Mlkem1024.Avx2.fsti
- [x] Libcrux_ml_kem.Mlkem1024.Portable.fst
[...]

PR #394, opened 2024-07-14
Make libcrux-ml-kem lax check in F*

karthikbhargavan:
This PR addresses https://github.com/cryspen/libcrux/issues/301

It extracts F* code for libcrux-ml-kem and typechecks the resulting code in lax mode. Running `./hax.py extract` generates F* code for portable, avx2, and neon versions. And all the code now lax typechecks by running `make` within `proofs/fstar/extraction`.

~~This PR depends on the corresponding hax PR that adds the necessary new library functions and types: https://github.com/hacspec/hax/pull/769~~ (merged)

~~It also currently relies on one manual edit to the generated F*, which we need to resolve before merging this PR:
See https://github.com/cryspen/libcrux/commit/91b12cf43ecfe10539bc5fe29becc77736248f8c~~
It relies on helping F* with typeclass inference using the `hax-lib::fstar` macro (see https://github.com/cryspen/libcrux/commit/f77dc3bae4975174692199ae5ce3e21b0f0e556e)

In the F* extraction driven by `hax.py`, one major change is that we now extract only the interface for `hash_functions.rs` and not the code, and this allows us to break the dependency on `libcrux-sha3`, significantly simplifying our task for verification. The advantage is that the F* code generated from `libcrux-ml-kem` no longer depends on any F* code or interfaces generated in `libcrux-sha3`. The drawback is that the code in `libcrux-ml-kem/hash_functions.rs` is no longer extracted to F*, only its interface is.  In a future PR, we should separately aim to extract and check `libcrux-sha3` and link it properly to `hash_functions.rs` (https://github.com/cryspen/libcrux/issues/395)

We also edited the Rust code in the following ways:
* The first change is that we rewrote the code in `vector/portable/serialize.rs`, `vector/portable/ntt.rs`, `vector/avx2/serialize.rs`,  and`vector/neon/serialize.rs` to factor out the core functions and reduce redundancy. We primarily did this to help F* handle the generated code (it was running out of memory with the previous code) but this rewrite also reduces the size of the Rust code and we hope it makes the source code more readable. 

* Second, we had to make some slightly annoying changes to the Rust code that were needed as a workaround for this issue: https://github.com/hacspec/hax/issues/720 These changes can be reverted when the hax issue is resolved.

The generated C code (via `c.sh` and `boring.sh`) compiles and passes tests.


libcrux-ml-kem/src/vector/portable/ntt.rs
```diff
@@ -1,6 +1,18 @@
...
+pub(crate) fn ntt_step(
+    v: &mut PortableVector,
+    zeta: i16,
+    i: usize,
+    j:usize
+) {
+    let t = montgomery_multiply_fe_by_fer(v.elements[j], zeta);
+    v.elements[j] = v.elements[i] - t;
+    v.elements[i] = v.elements[i] + t;
+}
```
franziskuskiefer, line 14: `i` and `j` could be `const` here. Not sure if it makes a difference though.

franziskuskiefer (changes requested):

libcrux-ml-kem/src/vector/portable/serialize.rs
```diff
@@ -4,321 +4,403 @@ use crate::vector::traits::FIELD_ELEMENTS_IN_VECTOR;
...
+    let r7 = ((v[5] >> 1) & 0xFF) as u8;
+    let r8 = ((v[6] & 0x3F) as u8) << 2 | (v[5] >> 9) as u8;
+    let r9 = ((v[7] & 0x7) as u8) << 5 | (v[6] >> 6) as u8;
+    let r10 = (v[7] >> 3) as u8;
+    (r0,r1,r2,r3,r4,r5,r6,r7,r8,r9,r10)
 }
 
 #[inline(always)]
 pub(crate) fn serialize_11(v: PortableVector) -> [u8; 22] {
+    let r0_10 = serialize_11_int(&v.elements[0..8]);
```
xvzcf, line 252: Is there a particular reason this style of splitting the serialize functions was adopted?

libcrux-ml-kem/hax.py
```diff
@@ -30,36 +30,6 @@ def shell(command, expect=0, cwd=None, env={}):
 class extractAction(argparse.Action):
 
     def __call__(self, parser, args, values, option_string=None) -> None:
-        # Extract sha3 interfaces
```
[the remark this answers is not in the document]
karthikbhargavan: In the F* extracted from this crate, we no longer rely on any SHA-3 extracted interfaces. This is because we only extract and check the _interface_ of `hash_functions.rs` not its implementation. I have added a comment to this effect to the top of `hash_functions.rs`. In a future PR, we should extract and lax-check libcrux-sha3 and then lax-check the _implementation_ of `hash_function.rs` against the `libcrux-sha3` interfaces. The general issue for tracking progress on libcrux-sha3 verification is https://github.com/cryspen/libcrux/issues/395

libcrux-ml-kem/Cargo.toml
```diff
@@ -28,6 +28,7 @@ libcrux-intrinsics = { version = "0.0.2-alpha.1", path = "../libcrux-intrinsics"
 # The hax config is set by the hax toolchain.
 [target.'cfg(hax)'.dependencies]
 hax-lib = { version = "0.1.0-alpha.1", git = "https://github.com/hacspec/hax/" }
+hax-lib-macros = { version = "0.1.0-alpha.1", git = "https://github.com/hacspec/hax/" }
```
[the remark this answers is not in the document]
karthikbhargavan: Removed.

libcrux-ml-kem/src/vector/neon/serialize.rs
```diff
@@ -43,81 +43,28 @@ pub(crate) fn serialize_4(v: SIMD128Vector) -> [u8; 8] {
...
-    high[0] = (high64 & 0x1F) as i16;
-    high[1] = ((high64 >> 5) & 0x1F) as i16;
-    high[2] = ((high64 >> 10) & 0x1F) as i16;
-    high[3] = ((high64 >> 15) & 0x1F) as i16;
-    high[4] = ((high64 >> 20) & 0x1F) as i16;
-    high[5] = ((high64 >> 25) & 0x1F) as i16;
-    high[6] = ((high64 >> 30) & 0x1F) as i16;
-    high[7] = ((high64 >> 35) & 0x1F) as i16;
-
+    let output = crate::vector::portable::PortableVector::deserialize_5(v);
```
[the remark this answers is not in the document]
karthikbhargavan: Done.

libcrux-ml-kem/src/vector/avx2/serialize.rs
```diff
@@ -514,17 +514,17 @@ pub(crate) fn deserialize_10(bytes: &[u8]) -> Vec256 {
...
 #[inline(always)]
 pub(crate) fn serialize_11(vector: Vec256) -> [u8; 22] {
-    let array = to_i16_array(SIMD256Vector { elements: vector });
-    let input = portable::from_i16_array(array);
-
-    portable::serialize_11(input)
+    let mut array = [0i16; 16];
+    mm256_storeu_si256_i16(&mut array, vector);
+    let input = crate::vector::portable::PortableVector::from_i16_array(&array);
+    crate::vector::portable::PortableVector::serialize_11(input)
```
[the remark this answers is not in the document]
karthikbhargavan: The c code extracts without cycles.

libcrux-ml-kem/src/hash_functions.rs
```diff
@@ -54,76 +54,112 @@ pub(crate) mod portable {
...
+        out
+    }
+
     impl<const K: usize> Hash<K> for PortableHash<K> {
         #[inline(always)]
         fn G(input: &[u8]) -> [u8; G_DIGEST_SIZE] {
-            let mut digest = [0u8; G_DIGEST_SIZE];
-            portable::sha512(&mut digest, input);
-            digest
+            G(input)
```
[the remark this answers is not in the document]
karthikbhargavan: Added a link to Issue https://github.com/hacspec/hax/issues/616  at the top of the file.
Extracting the functions out of the "impl" is needed both for abstraction in F* (interface hiding implementation details) and to make the F* typechecker able to handle this code.
However, this could be done automatically in hax without needing to uglify the Rust, hence the issue above.

libcrux-ml-kem/src/mlkem1024.rs
```diff
@@ -229,6 +229,13 @@ macro_rules! instantiate {
             /// Generates an ([`MlKem1024Ciphertext`], [`MlKemSharedSecret`]) tuple.
             /// The input is a reference to an unpacked public key of type [`MlKem1024PublicKeyUnpacked`],
             /// the SHA3-256 hash of this public key, and [`SHARED_SECRET_SIZE`] bytes of `randomness`.
+            #[cfg_attr(hax,hax_lib::fstar::before("
```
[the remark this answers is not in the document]
karthikbhargavan: Added a note linking it to https://github.com/hacspec/hax/issues/770

libcrux-ml-kem/hax.py
```diff
@@ -102,15 +76,15 @@ def __call__(self, parser, args, values, option_string=None) -> None:
...
+            "+** -libcrux_ml_kem::types::index_impls::** -libcrux_ml_kem::kem::** -libcrux_ml_kem::hash_functions::portable::* -libcrux_ml_kem::hash_functions::avx2::* -libcrux_ml_kem::hash_functions::neon::* +:libcrux_ml_kem::hash_functions::*::*"
         )
         interface_include = "+**"
         cargo_hax_into = [
             "cargo",
             "hax",
             "-C",
             "--features",
-            "pre-verification",
+            "simd128,simd256,pre-verification",
```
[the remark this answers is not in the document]
karthikbhargavan: Could you fix this? I don't know the desired format.

libcrux-ml-kem/src/vector/portable/ntt.rs
karthikbhargavan: Perf does not seem to be affected, probably because this is `inline(always)`.

libcrux-ml-kem/src/vector/portable/serialize.rs
karthikbhargavan: I added a comment on top of this file. 
```
//! A general style adopted here is to first define an internal function
//! called serialize_N_int or deserialize_N_int that (de)serializes
//! the minimal number of inputs K such that N*K is a multiple of 8.
//! These functions are then called multiple times in the main function,
//! called serialize_N or deserialize_N.
//! This factoring reduces redundancy, and also makes he code easier for
//! F* to handle. As a general rule, any function that modifes an array
//! more than 8 times with complex expressions starts to strain F*, so
//! we separate out the code that does the computation (in _int functions)
//! and code that updates arrays (in the outer functions).
```

karthikbhargavan: More specifically, F* works very well if we do the main computation in functions that return tuples and then only do simple array updates in the outer functions, hence this style choice. Many of these functions would also work if we used arrays everywhere, but they would be slow to even lax-check, which would then make full verification even slower.

libcrux-ml-kem/src/vector/portable/serialize.rs
```diff
@@ -1,324 +1,416 @@
...
+    result[4] = result4_7.0;
+    result[5] = result4_7.1;
+    result[6] = result4_7.2;
+    result[7] = result4_7.3;
     result
 }
 
+#[inline(always)]
+pub(crate) fn deserialize_4_int(bytes: &[u8]) -> (i16, i16, i16, i16, i16, i16, i16, i16) {
+    let input = u32::from_le_bytes(bytes.try_into().unwrap());
```
franziskuskiefer, line 69: This generates noise in the C code because it introduces an explicit panic. It's also more code. So I don't think this is a great style. And accessing the `bytes` directly is still not more than 8 operations and it's what's being done in the other functions too.

libcrux-ml-kem/src/vector/portable/serialize.rs
franziskuskiefer: I understand the reason. But it's pretty ugly. We should try to find a better style later.

franziskuskiefer (commented):
I made a couple changes to clean this up. We should change `deserialize_4_int`, then we can get this in.

libcrux-ml-kem/src/vector/portable/serialize.rs
karthikbhargavan: We actually had two versions of this function, one in `portable/serialize.rs` and another in `neon/serialize.rs`. This version was in `neon` and in my benchmarks seemed to be faster, so I picked that one. But benchmarks are noisy, and I have now reverted to using byte arrays here. In the future, if (de)serialization becomes a bottleneck we can look at it again.

libcrux-ml-kem/src/vector/portable/serialize.rs
karthikbhargavan: There are two things here.

The first is breaking down the functions into bits that repeat. This is a "good thing" for both code clarity and verification. Indeed, our original code in `src/kyber` was written this way. 

E.g. see the loop body in https://github.com/cryspen/libcrux/blob/02472265ddcd5af5cb1a9778855e081cb7f083b5/libcrux-ml-kem/src/kem/kyber/serialize.rs#L195 

It also breaks the input into chunks of 8, and here we are doing the same for a vector with 16 chunks.
The operations on the two chunks of 8 is the same.
I do not think this is an uglification, and it should not be changed.

The second bit is the use of tuples in the internal functions, and this is indeed an artifact of making it work in F*.
If we can, we should switch this to maybe working on arrays or mutable slices or something more idiomatic in Rust.

karthikbhargavan: I also added a comment to the code. There is another thing I tried which is to return an array expression instead of copying the tuple to the array in a long sequence of array updates. This was slowing down F* but still terminating, but I think when we start doing proofs, this slowdown will hurt more. So I leave it for a future pass where we prettify the code without killing F*.

franziskuskiefer (approved):
Let's get this merged and then incrementally get the modules to typecheck.

PR #401, opened 2024-07-15
Modularized PSQ

jschneider-bensch:
This PR modularizes the PSQ protocol into 
* a PSQ-component generation module: A minimal version of the previous PSQ protocol, deriving a "Pre-PSK" by encapsulating towards the receiver using PQ-KEM and then working in outer protocol context into the shared secret.
*  A protocol binder that makes use of the PSQ-component generation to bind to incorporate outer protocol requirements, e.g. a PSK time to live or binding / authenticating other material. This is exemplified here by a binder for ECDH based-protocols with PSK-TTL.


libcrux-psq/Cargo.toml
```diff
@@ -10,18 +10,19 @@ readme.workspace = true
...
+] }
 libcrux-hkdf = { version = "=0.0.2-alpha.1", path = "../libcrux-hkdf" }
 libcrux-hmac = { version = "=0.0.2-alpha.1", path = "../libcrux-hmac" }
 classic-mceliece-rust = { version = "2.0.0", features = [
     "mceliece460896f",
     "zeroize",
 ] }
 rand = { version = "0.8" }
+libcrux-ecdh = { version = "0.0.2-alpha.1", path = "../libcrux-ecdh" }
+libcrux = { version = "0.0.2-alpha.1", path = ".." }
```
franziskuskiefer, line 25: I guess we should expose everything through the `libcrux` crate eventually so that we only need this one dependency. (Not for now or this PR.)

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+const AEAD_KEY_NONCE: usize = Algorithm::key_size(Algorithm::Chacha20Poly1305)
+    + Algorithm::nonce_size(Algorithm::Chacha20Poly1305);
+
+const AEAD_KEY_LENGTH: usize = Algorithm::key_size(Algorithm::Chacha20Poly1305);
+
+/// An ECDH-bound PSQ encapsulation.
+pub struct ECDHPsk {
+    encapsulation: crate::psq::Ciphertext,
+    initiator_dh_pk: Vec<u8>,
+    aead_mac: (libcrux::aead::Tag, Vec<u8>),
```
franziskuskiefer, line 31: Using a struct here makes the code a little nicer. Not too important though.

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
+//! # ECDH Binder
+//!
+//! This module implements a protocol binder that binds a post-quantum
+//! pre-shared key component to an outer ECDH based key agreement.
```
franziskuskiefer, line 4: The comments read like this could be any ECDH key, but it is implement for x25519 only.
That's fine, but we should mention it at least once up here.

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+    receiver_dh_pk: &X25519PublicKey,
+    initiator_dh_sk: &X25519PrivateKey,
+    psk_ttl: Duration,
+    sctx: &[u8],
+    rng: &mut (impl CryptoRng + Rng),
+) -> Result<(Psk, ECDHPsk), Error> {
+    let now = SystemTime::now();
+    let ts = now
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("now cannot be before UNIX_EPOCH");
```
franziskuskiefer, line 54: We should try to generally write non-panicking code. That's easy here since we return a `Result` anyway. You can create a new `SystemError` or something like that.
It's a pointless error because the applicaiton probably can't do anything about it. But at least we don't crash.

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+    let ts_seconds = ts.as_secs();
+    let ts_subsec_millis = ts.subsec_millis();
+    let mut ts_ttl = ts_seconds.to_be_bytes().to_vec();
```
franziskuskiefer, line 57: This has a known size. So you could put it into a fixed sized array instead of a heap allocation (would allow `no_std`).

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+    ts_ttl.extend_from_slice(&ts_subsec_millis.to_be_bytes());
+    ts_ttl.extend_from_slice(&psk_ttl.as_millis().to_be_bytes());
+
+    let (ss_q, encapsulation) = psq_pk.gen_pq_psk(sctx, rng)?;
+    let ss_dh = libcrux_ecdh::x25519_derive(receiver_dh_pk, initiator_dh_sk)
+        .map_err(|_| Error::CryptoError)?;
+
+    // ikm = ss_q || ss_dh
+    let mut ikm = Vec::from(ss_q);
```
franziskuskiefer, line 66: Same here (fixed sized vs vec).

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+    ikm.extend_from_slice(&ss_dh.0);
+
+    let prk = libcrux_hkdf::extract(libcrux_hkdf::Algorithm::Sha256, b"", ikm);
```
franziskuskiefer, line 69: Can you file an issue to add a fixed sized APIs for HKDF, or something where we pass in the output?

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    let psk: [u8; DH_PSK_LENGTH] = libcrux_hkdf::expand(
+        libcrux_hkdf::Algorithm::Sha256,
+        prk,
+        DH_PSK_CONTEXT,
+        DH_PSK_LENGTH,
+    )
+    .map_err(|_| Error::CryptoError)?
+    .try_into()
+    .expect("should receive the correct number of bytes from HKDF");
```
franziskuskiefer, line 79: We could do without a panic here again.

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+) -> Result<Psk, Error> {
+    let ECDHPsk {
+        encapsulation,
+        initiator_dh_pk,
+        aead_mac,
+        psk_ttl,
+        ts,
+    } = ecdh_psk_message;
+    let ss_q = receiver_pqsk.derive_pq_psk(receiver_pqpk, encapsulation, sctx)?;
+    let initiator_dh_pk_bytes: [u8; 32] = initiator_dh_pk[0..32]
```
franziskuskiefer, line 122: The `[0..32]` is an implicit panic. The `try_into` won't fail, but the indexing may panic. So you should check the length before. Or even better, not use a `Vec` in the `ECDHPsk`.

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+
+    // validate TTL
+    let now = SystemTime::now();
+    let ts_since_epoch =
+        Duration::from_secs(ts_seconds) + Duration::from_millis((ts_subsec_millis).into());
+    if initiator_dh_pk_decrypted != *initiator_dh_pk
+        || now
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("now cannot be before UNIX_EPOCH")
+            - ts_since_epoch
```
franziskuskiefer, line 167: This may underflow. You should check that this can't happen in the condition before.

franziskuskiefer (changes requested):
Thanks! lgtm in general. I left some comments on the code (not protocol, that looks fine).

libcrux-psq/src/ecdh_binder.rs
```diff
@@ -0,0 +1,310 @@
...
+    psk: [u8; 32],
+    info: &[u8],
+) -> Result<(libcrux::aead::Iv, libcrux::aead::Key), Error> {
+    let key_iv_bytes =
+        libcrux_hkdf::expand(libcrux_hkdf::Algorithm::Sha256, psk, info, AEAD_KEY_NONCE)
+            .map_err(|_| Error::CryptoError)?;
+    let (key_bytes, iv_bytes) = key_iv_bytes.split_at(AEAD_KEY_LENGTH);
+    let key = libcrux::aead::Key::from_bytes(
+        libcrux::aead::Algorithm::Chacha20Poly1305,
+        key_bytes.to_vec(),
```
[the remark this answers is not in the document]
jschneider-bensch: There is `from_slice` for `Key` already, I found :+1:

jschneider-bensch:
We know the content of the AEAD ciphertext as well, so I could use the non-allocating API there, too.
But I found that a bit weird to use, since you have to allocate a buffer for ciphertext _and_ tag, and it returns the tag as well.

franziskuskiefer (approved):
lgtm, let's get this in and then iterate.

It would be great if you would add a usage example in the lib docs os there's an easy starting point.

PR #404, opened 2024-07-16
Make constant_time_ops.rs and utils.rs panic-free

mamonet:
This is an entry PR that makes sure particular modules of ML-KEM are panic-free, it adds assumptions and pre-conditions using `hax_lib::requires` to functions that have a potential to panic to get the extracted F* filed verifiable. For now, two modules have been added and others are yet to be followed.


mamonet:
I didn't get to add serialize.rs module today, hopefully will include it tomorrow.

franziskuskiefer:
Let's go one module at a time. So just drop serialize from this PR and we have something to get in.
It looks like the hax job failed though. Also, please run the typechecking on CI for the modules that tc.

mamonet:
Ok. Let me see where it goes wrong.

mamonet:
Apparently, the process of lax-checking `Libcrux_ml_kem.Ind_cca.fst` has been killed in hax CI job. Maybe certain functions need to be divided up there, I will see which functions in that module take the longest to lax-check.

libcrux-ml-kem/src/constant_time_ops.rs
```diff
@@ -78,6 +70,10 @@ pub(crate) fn compare_ciphertexts_in_constant_time(lhs: &[u8], rhs: &[u8]) -> u8
 }
 
 #[inline(never)] // Don't inline this to avoid that the compiler optimizes this out.
+#[cfg_attr(hax, hax_lib::requires(
+    lhs.len() == rhs.len() &&
+    lhs.len() == SHARED_SECRET_SIZE
+))]
```
mamonet, line 76: Here's my observation, with this pre-condition applied, `decapsulate` function in Libcrux_ml_kem.Ind_cca.fst takes 2.5 mins to lax-check. Without the pre-condition, that function takes ~1 min to lax-check. `encapsulate` function lax-checks right away in both cases.

mamonet: It seems F* has difficulty to have decapsulate function, is splitting it up an option?

mamonet:
@franziskuskiefer It seems the last commit makes hax CI job green, could you review it please or assign someone who is aware of `decapsulate` logic to look at?

libcrux-ml-kem/src/ind_cca.rs
```diff
@@ -261,16 +262,15 @@ pub(crate) fn decapsulate<
...
     >(ind_cpa_public_key, decrypted, pseudorandomness);
 
-    let selector = compare_ciphertexts_in_constant_time(ciphertext.as_ref(), &expected_ciphertext);
-
     let implicit_rejection_shared_secret =
         Scheme::kdf::<K, CIPHERTEXT_SIZE, Hasher>(&implicit_rejection_shared_secret, ciphertext);
     let shared_secret = Scheme::kdf::<K, CIPHERTEXT_SIZE, Hasher>(shared_secret, ciphertext);
 
-    select_shared_secret_in_constant_time(
+    compare_ciphertexts_select_shared_secret_in_constant_time(
```
franziskuskiefer, line 269: I think this should be fine. But I wonder why F* has trouble here. @karthikbhargavan do you understand what's going on here?

karthikbhargavan (approved):
I will debug the lax thing, seems weird. But meanwhile, let's get the Makefile updated and get this in so the pipeline is in motion.

jschneider-bensch:
SHA-3 C extraction should be fixed with the changes in #412.

PR #405, opened 2024-07-16
ML-DSA: Propagated SIMDUnit type throughout the code.

xvzcf:
- Removed the non-SIMD `PolynomialRingElement` struct and renamed `SIMDPolynomialRingElement -> PolynomialRingElement`

- Moved the code for serializing and de-serializing the `Signature` struct into a new `encoding/signature.rs` file.


franziskuskiefer (approved):
lgtm, looks like we're ready to get on the avx2 implementation?

xvzcf:
> lgtm, looks like we're ready to get on the avx2 implementation?

Yeah, in the `ml_dsa_{44|65|87}` top level files I'll just add in simple feature gates to pass in the right `SIMDUnit` type, and take it from there.

PR #410, opened 2024-07-18
SHA3 NIST test vectors

franziskuskiefer:
Adding NIST's CAVP test vectors for SHA3.


cavp/src/lib.rs
```diff
@@ -0,0 +1,485 @@
...
+
+            // Read a section header
+            //
+            // A section header has the form `[ ... ]`
+            if line.starts_with("[") && line.ends_with("]") {
+                line = line.strip_suffix("]").ok_or(Error::Read)?;
+                line = line.strip_prefix("[").ok_or(Error::Read)?;
+                log::debug!("line {line}");
+                T::H::read_header::<T>(line, &mut header)?;
+                // section.push(line.to_string());
```
jschneider-bensch, line 65: Was the idea here to group test vectors by section?

cavp/src/lib.rs
jschneider-bensch, line 1: Nice! This could be useful for reading other CAVP vectors as well, I think. It also works if there are no sections defined in the `.rsp` file, right?  This is the case e.g. in the Kyber submission KATs.

Issue #145, opened 2023-12-04
CI for C Extraction


karthikbhargavan:
Add C extraction to libcrux CI.

- [x] #408 
- [x] #411 
- [x] Run build and tests on extracted version
- [x] Fail when there's a diff between checked-in C code and generated C code

PR #411, opened 2024-07-18
Add extraction job to C workflow

jschneider-bensch:


franziskuskiefer (approved):
Great! Looks like this is working.
I updated #145 to track what we need here.

PR #413, opened 2024-07-18
better header for ml-kem C extraction

franziskuskiefer:
Add libcrux revision, license and copyright to the C headers.

This requires aeneasverif/eurydice#43


jschneider-bensch:
For CI to work we need to update the docker image to the latest version of Eurydice.

PR #420, opened 2024-07-18
c extraction update

franziskuskiefer:
New C extraction with changes from #413 and #410.


jschneider-bensch (approved):
Looks good to me :+1:  
#404 does not affect C extraction, right?

PR #421, opened 2024-07-18
Make ML-KEM serialize.rs panic-free

mamonet:
This patch adds pre-conditions and workaround assumptions to serialize.rs to make Libcrux_ml_kem.Serialize.fst verifiable.
It also adds Libcrux_ml_kem.Serialize.fst to PANIC_FREE category in ML-KEM Makefile.

### DIFF
--- a/libcrux-ml-kem/src/serialize.rs
+++ b/libcrux-ml-kem/src/serialize.rs
@@ -1,20 +1,27 @@
 use crate::{
-    constants::{BYTES_PER_RING_ELEMENT, SHARED_SECRET_SIZE},
-    hax_utils::hax_debug_assert,
+    constants::{COEFFICIENTS_IN_RING_ELEMENT, BYTES_PER_RING_ELEMENT, SHARED_SECRET_SIZE},
     helper::cloop,
     polynomial::{PolynomialRingElement, VECTORS_IN_RING_ELEMENT},
     vector::{decompress_1, to_unsigned_representative, Operations},
 };
 
+use hax_lib::*;
+use hax_lib::int::*;
+
 #[inline(always)]
+#[cfg_attr(hax, requires(
+    32 + 2 <= SHARED_SECRET_SIZE
+))]
 pub(super) fn compress_then_serialize_message<Vector: Operations>(
     re: PolynomialRingElement<Vector>,
 ) -> [u8; SHARED_SECRET_SIZE] {
     let mut serialized = [0u8; SHARED_SECRET_SIZE];
     for i in 0..16 {
         let coefficient = to_unsigned_representative::<Vector>(re.coefficients[i]);
+        fstar!("assume (Libcrux_ml_kem.Vector.Traits.f_compress_1_pre #v_Vector coefficient)");
         let coefficient_compressed = Vector::compress_1(coefficient);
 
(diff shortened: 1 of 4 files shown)


libcrux-ml-kem/src/serialize.rs
```diff
@@ -102,94 +131,156 @@ pub(super) fn deserialize_ring_elements_reduced<
...
-
     match COMPRESSION_FACTOR as u32 {
         10 => compress_then_serialize_10(re),
-        11 => compress_then_serialize_11(re),
-        _ => unreachable!(),
+        // Why it needs assumption here?
+        11 => {assume!(OUT_LEN == 352); compress_then_serialize_11(re)},
+        // unreachable!() does't verify, Option enum works but we have
+        // to edit refrences!!?
+        _ => [0u8; OUT_LEN],
```
mamonet, line 213: With Option enum path, we can the following
```pub(super) fn compress_then_serialize_ring_element_u<
    const COMPRESSION_FACTOR: usize,
    const OUT_LEN: usize,
    Vector: Operations,
>(
    re: &PolynomialRingElement<Vector>,
) -> Option<[u8; OUT_LEN]> {
    match COMPRESSION_FACTOR as u32 {
        10 => Some(compress_then_serialize_10(re)),
        11 => Some(compress_then_serialize_11(re)),
        _ => None,
    }
}```
This would work and verify but also need to modify the references!!

Comment (franziskuskiefer) [reply libcrux-ml-kem/src/serialize.rs:213]: This should stay `unreachable`. That's been the style used so far and worked well. Everything else is not correct.

File: libcrux-ml-kem/src/serialize.rs [hunk 2]
```diff
@@ -1,20 +1,23 @@
 use crate::{
-    constants::{BYTES_PER_RING_ELEMENT, SHARED_SECRET_SIZE},
-    hax_utils::hax_debug_assert,
+    constants::{COEFFICIENTS_IN_RING_ELEMENT, BYTES_PER_RING_ELEMENT, SHARED_SECRET_SIZE},
     helper::cloop,
     polynomial::{PolynomialRingElement, VECTORS_IN_RING_ELEMENT},
     vector::{decompress_1, to_unsigned_representative, Operations},
 };
 
+use hax_lib::*;
```
franziskuskiefer, line 8: This needs a `cfg` guard.

libcrux-ml-kem/src/serialize.rs
mamonet: I am trying to verify `unreachable` avenue, if I couldn't I will seek help from Karthik.

mamonet: I replaced with `unreachable` expression and used proper asserts to complement that.

libcrux-ml-kem/src/serialize.rs
mamonet: What is supposed to be inside `cfg` expression?

mamonet:
I turn this PR to draft until we get rid of all the assumes.

franziskuskiefer:
@mamonet What's the state here? This looks stale. Do you want to close it?

mamonet:
The changes are still relative. However, this PR depends on https://github.com/hacspec/hax/issues/605 to remove assumes.

mamonet:
I will close this PR because it's now part of https://github.com/cryspen/libcrux/pull/488

PR #422, opened 2024-07-18
Refresh C ml-kem code

franziskuskiefer:
Drop unused functions.

### DIFF
--- a/libcrux-ml-kem/cg/libcrux_core.h
+++ b/libcrux-ml-kem/cg/libcrux_core.h
@@ -4,11 +4,11 @@
  * SPDX-License-Identifier: MIT or Apache-2.0
  *
  * This code was generated with the following revisions:
- * Charon: 0f1b5e50fa4e96ed9e650d8334e1afbf2bf319b1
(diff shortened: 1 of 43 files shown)


libcrux-ml-kem/cg/libcrux_core.h
```diff
@@ -25,10 +25,6 @@ typedef struct core_ops_range_Range__size_t_s {
   size_t end;
 } core_ops_range_Range__size_t;
 
-static inline uint8_t Eurydice_bitand_pv_u8(uint8_t *x, uint8_t y);
```
jschneider-bensch, line 28: I guess these were some kind of Eurydice helpers that are no longer generated?

jschneider-bensch (approved):

libcrux-ml-kem/cg/libcrux_core.h
franziskuskiefer: Yes, they are built-ins, generated from eurydice. But they were not caught by the phase that removed unused code. Now they are and hence are removed. See https://github.com/AeneasVerif/eurydice/issues/45

PR #438, opened 2024-07-25
use pre-verification feature for ml-kem on CI runs

franziskuskiefer:
We don't need to test the old version, let's use the pre-verification version that we actually use and work on.

### DIFF
--- a/.github/workflows/mlkem.yml
+++ b/.github/workflows/mlkem.yml
@@ -86,14 +86,14 @@ jobs:
       - name: 🔨 Build
         run: |
           rustc --print=cfg
-          cargo build --verbose $RUST_TARGET_FLAG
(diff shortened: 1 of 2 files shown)


jschneider-bensch (approved):

jschneider-bensch:
Good idea to test this, we found an issue already! :+1:  Tracking this in #439.

franziskuskiefer:
Yes, some other things were committed as well in #404 that broke the code.

franziskuskiefer:
> Good idea to test this, we found an issue already! 👍 Tracking this in #439.

This should also be fixed in [dcd512b](https://github.com/cryspen/libcrux/pull/438/commits/dcd512bb8722ffba93eef9f9c4de5f28649d19f2). But we also need to fix the intrinsics file.

PR #440, opened 2024-07-25
ML-DSA: Full AVX2 implementation of uniform rejection sampling and partial AVX2 implementation of gamma1 serialization.

xvzcf:


libcrux-intrinsics/src/avx2.rs
```diff
@@ -26,7 +26,13 @@ pub fn mm256_storeu_si256_i32(output: &mut [i32], vector: Vec256) {
 }
 
 pub fn mm_storeu_si128(output: &mut [i16], vector: Vec128) {
-    // debug_assert_eq!(output.len(), 8);
+    //debug_assert_eq!(output.len(), 8);
```
karthikbhargavan, line 29: Do you want to restore this assert?

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
```diff
@@ -0,0 +1,89 @@
+use libcrux_intrinsics::avx2::*;
+
+#[inline(always)]
+fn serialize_when_gamma1_is_2_pow_17<const OUTPUT_SIZE: usize>(
```
karthikbhargavan, line 4: Why replicate the code from portable here? Wouldn't it be better to convert to an array of integers and call the portable serialize?

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        serialized[9 * i + 7] = (coefficient3 >> 2) as u8;
+        serialized[9 * i + 8] = (coefficient3 >> 10) as u8;
+    }
+
+    serialized
+}
+
+#[inline(always)]
+fn serialize_when_gamma1_is_2_pow_19<const OUTPUT_SIZE: usize>(
```
karthikbhargavan, line 49: This appears to work for any `OUTPUT_SIZE <= 20` ? Maybe add that as a precondition?

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
```diff
@@ -0,0 +1,89 @@
...
+    // ... and the second 80 bits at position 0 in the upper 128-bit lane
+    let upper_4 = mm256_extracti128_si256::<1>(adjacent_4_combined);
+    mm_storeu_bytes_si128(&mut serialized[10..26], upper_4);
+
+    serialized[0..20].try_into().unwrap()
+}
+
+#[inline(always)]
+pub fn serialize<const OUTPUT_SIZE: usize>(simd_unit: Vec256) -> [u8; OUTPUT_SIZE] {
+    match OUTPUT_SIZE {
```
karthikbhargavan, line 84: This could really just be a generic function, unless you were planning on specializing the code for 18/20?

libcrux-ml-dsa/src/simd/avx2/generate_uniform_rej_sample_table.py
```diff
@@ -0,0 +1,36 @@
+#! /usr/bin/env python3
+
+
+def trailing_zeros(num):
+    return (num & -num).bit_length() - 1
+
+
+with open("uniform_rej_sample_table.rs", "w+") as f:
```
karthikbhargavan, line 8: Could you write this code in Rust? If you do, we could verify it....

karthikbhargavan (approved):
Look at my comments, but generally looks on the right track.

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
xvzcf: I'm not sure I follow, the implementations will be faster if specialized to 18/20. Could you elaborate?

libcrux-intrinsics/src/avx2.rs
xvzcf: I tried, but I seem to be getting Eurydice-related errors?

xvzcf: Removed the assert again for now.

libcrux-ml-dsa/src/simd/avx2/generate_uniform_rej_sample_table.py
xvzcf: I added an executable specification to `libcrux-ml-dsa/src/simd/avx2/uniform_rej_sample_table.rs` and a unit-test in there. Could you take a look and let me know?

libcrux-intrinsics/src/avx2.rs
franziskuskiefer: You changed the code, hence the C code changed. There's no error in that though.
However, this assert isn't good. We should allo outputs that are longer to avoid unnecessary code on the caller. I did that.

jschneider-bensch: Does that also go for the assert in `mm_storeu_si128_i32`, then, @franziskuskiefer?

libcrux-ml-dsa/src/encoding/gamma1.rs
jschneider-bensch, line 1: Could you add a pointer to where the test values here come from? Perhaps it's already documented somewhere else and I missed it.

libcrux-intrinsics/src/avx2.rs
franziskuskiefer: It's a bit of a style question I think. Having clear bounds can be nice (then we can have an assert here and use it in the proofs). But it also requires calling the functions with a slice of the right size (`&x[a...b]`), which in turn generates ugly slices in C. The Rust side is ok, the C side isn't great.
Overall I feel like handling slice only with the required bound (not the exact size) makes everything a little nicer.

libcrux-ml-dsa/src/encoding/gamma1.rs
xvzcf: There's a file called `notes.md` in the `libcrux-ml-dsa/src` that tells a bit about how, maybe I should use all caps for that file name to make it stand out more.

jschneider-bensch (approved):

Issue #275, opened 2024-05-14
ML-DSA 65 AVX2 Implementation


franziskuskiefer:
When the portable implementation #274 is working, implement an AVX2 variant.

- [x] #571
- [ ] assess performance after the NTT changes

---

- [x] https://github.com/cryspen/libcrux/pull/375
- [x] #389 
- [x] #341
- [x] #396
- [x] #423
- [x] #440 
- [x] #441
- [x] #484

franziskuskiefer:
Pushing the deadline to the end of September to
- make it extract
- improve performance

jschneider-bensch:
Moving the deadline on this to mid-November since #632 threw a wrench in the works for this.

karthikbhargavan:
@jschneider-bensch  this seems done now?

jschneider-bensch:
Yes, we are done optimizing AVX2 performance, closing this.

PR #441, opened 2024-07-26
mldsa parallel shake

franziskuskiefer:
Here an option for parallel shake in ML-DSA. There's more room for improvement but this gives us some of it.
Performance numbers are all over the place. We need to write better benchmarks.

~~Based on top of #440~~

cc #275


libcrux-ml-dsa/src/hash_functions.rs
```diff
@@ -20,12 +40,35 @@ pub(crate) mod shake128 {
 
     pub(crate) trait Xof {
         fn init_absorb(input: &[u8]) -> Self;
+
         // TODO:
         // - There should only be a `squeeze_block` and `squeeze_five_blocks`
         // - Use mutable out instead if performance is better
         fn squeeze_first_five_blocks(&mut self) -> [u8; FIVE_BLOCKS_SIZE];
```
karthikbhargavan, line 47: I think it is worth checking if mutable out for 5 blocks may provide better perf.

libcrux-ml-dsa/src/ml_dsa_65.rs
```diff
@@ -68,22 +68,33 @@ type SIMDUnit = crate::simd::avx2::AVX2SIMDUnit;
...
+type Shake256 = crate::hash_functions::portable::Shake256;
 #[cfg(not(feature = "simd256"))]
-type Shake256 = crate::hash_functions::portable::PortableShake256;
+type Shake256 = crate::hash_functions::portable::Shake256;
 
 /// Generate an ML-DSA-65 Key Pair
 pub fn generate_key_pair(randomness: [u8; KEY_GENERATION_RANDOMNESS_SIZE]) -> MLDSA65KeyPair {
     let (signing_key, verification_key) = crate::ml_dsa_generic::generate_key_pair::<
         SIMDUnit,
         Shake128,
```
karthikbhargavan, line 94: I guess there is no nice way of packaging all of these Shake parameters into a single parameter?

karthikbhargavan (approved):
Changes look good. Haven't had a chance to try the benchmarks yet.

libcrux-ml-dsa/src/hash_functions.rs
franziskuskiefer: Yeah I think that's worth doing eventually because it's big enough again.

libcrux-ml-dsa/src/ml_dsa_65.rs
franziskuskiefer: I don't think we want this in on e parameter. We can drop some of them again when everything is cleaned up. But there are different states and ideally we only pass through the types we actually need.

libcrux-ml-dsa/src/polynomial.rs
```diff
@@ -31,7 +31,7 @@ impl<SIMDUnit: Operations> PolynomialRingElement<SIMDUnit> {
     // This is useful for debugging.
     #[allow(dead_code)]
     pub(crate) fn from_i32_array(array: &[i32]) -> Self {
-        debug_assert!(array.len() == 256);
+        debug_assert!(array.len() >= 256);
```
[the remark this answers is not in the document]
franziskuskiefer: This is actually mostly larger but the callers where doing a `&[0..256]`.

franziskuskiefer:
I parallelised a couple more and moved to some more mutable outputs. We should see performance of the arithmetic better now.

libcrux-ml-dsa/src/matrix.rs
```diff
@@ -21,9 +22,25 @@ pub(crate) fn expand_to_A<
...
     #[allow(clippy::needless_range_loop)]
     for i in 0..ROWS_IN_A {
-        for j in 0..COLUMNS_IN_A {
+        let samples = sample_four_ring_element_uniform::<SIMDUnit, Shake128X4>(seed, i);
+        A[i][0] = samples.0;
+        A[i][1] = samples.1;
+        A[i][2] = samples.2;
+        A[i][3] = samples.3;
+
+        // TODO: We could do more in parallel if we think that may speed
```
jschneider-bensch, line 41: The 4x4 and 8x7 cases could be fully parallelized in groups of 4, 5x6 is the odd one out, with at least to individual SHAKE calls. Perhaps in this case we could also do everything in groups of 4 and swallow a useless computation of two hashes, in case it's still faster than two individual calls? I'll make a note to check it out later.

jschneider-bensch (commented):
Looks like the ML-KEM C extractions also need to be updated to reflect the changes in SHA3.

libcrux-ml-dsa/src/matrix.rs
franziskuskiefer: Yes, we can definitely do a little more in parallel here! Let's get a first round in and then do more.

PR #448, opened 2024-07-29
Update Readmes

jschneider-bensch:
This PR adds some basic Readme files to subcrates, indicating their verification status.
In the description of the status on the top-level, I indicate that each crate should contain a more detailed note describing its verification status, beyond the binary `verified/pre-verification` distinction. These notes are not yet part of the PR.


franziskuskiefer:
What needs to be done to get this in @jschneider-bensch ?

jschneider-bensch:
@karthikbhargavan suggested a further distinction between verified algorithms and verified APIs on top, which would clarify the situation in cases like `libcrux-hkdf`, where we have a thin unverified API on top of verified algorithms.

franziskuskiefer:
Ok, if you think that's worth doing, let's do it. Should be pretty quick. Otherwise we should get this merged.

jschneider-bensch:
I'll file a follow-up to provide more detailed verification status.

PR #450, opened 2024-07-30
ML-DSA: AVX2 implementation of error ring element sampling.

xvzcf:


franziskuskiefer:
This seems to stabilise signing performance a little, but at a slower speed. Let's try to figure out what's going on here.

jschneider-bensch (approved):
I think this looks good!

PR #451, opened 2024-07-30
Clean up SHA3 XOF

jschneider-bensch:
I did a bit of clean up around the main work, done by @franziskuskiefer:
- Some changes around the XOF API: Giving SHAKE128 and SHAKE256 as implementations of a sealed XOF trait, so library users can't instantiate non-standard XOFs; also, I introduced a distinction between absorb and squeeze states, so it shouldn't be possible to squeeze before final absorption.
- Small things: renaming, taking care of most warnings. I'm unsure about the status of `generic_keccak::keccak_xof`. I made sure it works if used in lieu of `generic_keccak::keccak` in `portable::keccakx1`, but I don't know which to prefer then.
- Made sure the ML-KEM C extraction was working.

It's just the portable implementation for now, I plan to address the SIMD implementations in a follow up.


libcrux-sha3/src/traits.rs
```diff
@@ -15,11 +15,12 @@ pub(crate) mod internal {
         fn and_not_xor(a: Self, b: Self, c: Self) -> Self;
         fn xor_constant(a: Self, c: u64) -> Self;
         fn xor(a: Self, b: Self) -> Self;
-        fn load_block<const BLOCKSIZE: usize>(a: &mut [[Self; 5]; 5], b: [&[u8]; N]);
-        fn store_block<const BLOCKSIZE: usize>(a: &[[Self; 5]; 5], b: [&mut [u8]; N]);
-        fn load_block_full<const BLOCKSIZE: usize>(a: &mut [[Self; 5]; 5], b: [[u8; 200]; N]);
-        fn store_block_full<const BLOCKSIZE: usize>(a: &[[Self; 5]; 5]) -> [[u8; 200]; N];
+        fn load_block<const RATE: usize>(a: &mut [[Self; 5]; 5], b: [&[u8]; N]);
```
xvzcf, line 18: This is nitpicky, but would renaming this to BITRATE be better? since I believe that's the term the spec uses?

xvzcf (commented):
Generally looks good, but what's the plan for the remaining warnings? I'm getting a bunch of them when compiling and testing in libcrux-ml-dsa

jschneider-bensch:
Right, I'm unsure about `keccak_xof`, maybe it's not needed at all. That being unused is the only warning I get, but you're right, I see more in the CI runs, e.g. [here](https://github.com/cryspen/libcrux/actions/runs/10199804571/job/28217801552?pr=451). AFAICT, the issue there is because parts of the `x4` API are not implemented for the portable version. Do you get even more other warnings?

libcrux-sha3/src/traits.rs
xvzcf: Ah, my bad, I was looking at this: https://keccak.team/files/Keccak-reference-3.0.pdf

xvzcf:
>  AFAICT, the issue there is because parts of the `x4` API are not implemented for the portable version. Do you get even more other warnings?

I see, I don't get more warnings than those produced by CI. Perhaps we could prefix these variables with underscores for now and put a link to a Github issue next to them?

libcrux-sha3/src/simd/avx2.rs
```diff
@@ -241,4 +241,8 @@ impl KeccakItem<4> for Vec256 {
     fn split_at_mut_n(a: [&mut [u8]; 4], mid: usize) -> ([&mut [u8]; 4], [&mut [u8]; 4]) {
         split_at_mut_4(a, mid)
     }
+
+    fn store<const RATE: usize>(_state: &[[Self; 5]; 5], _out: [&mut [u8]; 4]) {
```
karthikbhargavan, line 245: These todos need to be implemented in this round right?

jschneider-bensch: I don't know if we have a need for an x2 or x4 incremental API? In ML-DSA we only need the incremental API when taking in the message. In any case I've added an issue to track this for later: #482

franziskuskiefer (approved):
Thanks!

PR #457, opened 2024-08-01
Extract Kyber Round 3 for header only C

franziskuskiefer:


libcrux-ml-kem/src/mlkem768.rs
jschneider-bensch, line 1: Why did you push the `cfg` down to the items here? Just curious.

jschneider-bensch (approved):

libcrux-ml-kem/src/mlkem768.rs
franziskuskiefer: Good point. I played with adding some things in there but didn't end up using it. This is a leftover of that experiment. I'll revert this change.

PR #465, opened 2024-08-05
Implement `Clone` and `Send` for SHA2 state and hasher types

keks:
We need these implementations for implementing a provider that requires these features.

`Send` should be safe, since there is no other thread in C-land that operates on the data in the hash state. We already did the same for Drbg: #205


mamonet:
The changes look good. Since there's no race conditions coming from C-land, I think it's safe to send the hash state to other thread.

mamonet (approved):

PR #468, opened 2024-08-06
feat(ci/nix): always update dependencies (and print them!)

W95Psp:
This PR runs `nix flake update charon eurydice fstar karamel` before building, which:
 - updates dependencies to the latest
 - print the exact commit used

IMO that's not great for reproducibility (we should run `nix flake update charon eurydice fstar karamel` and commit that information), but at least CI logs will show for which commit of every dependency the job succeeded.


franziskuskiefer:
We should at least run the job once before merging this.
But I think more importantly, @pnmadelaine needs to add documentation where this is used and how.

PR #473, opened 2024-08-09
ML-DSA: Unroll sampling of matrix A and error ring vectors.

xvzcf:


franziskuskiefer (approved):
Thanks, let's get this in.

PR #475, opened 2024-08-11
A more complete spec of ML-KEM

karthikbhargavan:
This PR fills out some of the functions of the ML-KEM spec that had not been previously specified and does a general clean up of the spec in prep for the proofs.

PR #485, opened 2024-08-13
Proofs misc setup improvements

W95Psp:
This PR has a bunch of improvements for the F* setup.
 - It's using the new Makefile I wrote for the tutorial, and that I improved yesterday evening for sandbox. Now F* caching is project-wide and should work.
 - Use only one dependency, workspace-wide, to hax-lib, so that we can actually run `cargo update hax-lib`
 - put the makefile at one place, and have proxy makefiles that just include them


mamonet:
The output layout is much better, and rebuilding now starts where it left off.

mamonet (approved):

W95Psp:
Seems like this was merged into https://github.com/cryspen/libcrux/pull/475

karthikbhargavan:
https://github.com/cryspen/libcrux/pull/475 is now merged.
Could you check if there were some changes here that did not make it into that PR?
Otherwise, let's close this.

W95Psp:
Nice, let's close this, there are a few thing, gonna make another PR

PR #486, opened 2024-08-13
Manual benchmarks for ML-DSA

jschneider-bensch:
This extends the manual benchmark from Franziskus' [NTT unrolling branch](https://github.com/cryspen/libcrux/blob/franziskus/mldsa-unroll-ntt-wip/libcrux-ml-dsa/examples/sign_65.rs) to cover the key generation, signing and verification across the three ML-DSA variants.

Can be run with `cargo bench manual`.


libcrux-ml-dsa/benches/manual.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+macro_rules! bench_group {
+    ($variant:literal, $mod:ident, $keypair_t:ident, $signature_t:ident) => {{
+        bench!(
+            "KeyGen",
+            $variant,
+            (),
+            |()| {
+                // FIXME: expose the constants
```
franziskuskiefer, line 77: You expose them, so you can drop the FIXMEs :)

franziskuskiefer (approved):
lgtm with one nit.

We should add pqclean benchmarks as well to have a fair comparison. But we can do that in a follow up.

libcrux-ml-dsa/benches/manual.rs
```diff
@@ -0,0 +1,131 @@
...
+            }
+        );
+
+        println!("");
+    }};
+}
+
+fn main() {
+    bench_group!("65", ml_dsa_65, MLDSA65KeyPair, MLDSA65Signature);
+    bench_group!("44", ml_dsa_44, MLDSA44KeyPair, MLDSA44Signature);
```
franziskuskiefer, line 129: Maybe swap the two here.

libcrux-ml-dsa/benches/manual.rs
```diff
@@ -0,0 +1,128 @@
...
+        format!("{} μs", micros / ITERATIONS as u128)
+    } else if micros < (1_000_000 * ITERATIONS as u128) {
+        format!(
+            "{:.2} ms",
+            (micros as f64 / (1_000_f64 * ITERATIONS as f64))
+        )
+    } else {
+        format!(
+            "{:.2}s",
+            (micros as f64 / (1_000_000_f64 * ITERATIONS as f64))
```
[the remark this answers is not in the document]
jschneider-bensch: Will do that in the next iteration :+1:

Issue #454, opened 2024-07-30
Formally Verify ML-KEM (Portable + AVX2)


karthikbhargavan:
Some dependencies:

* ML-KEM Spec: [https://github.com/cryspen/libcrux/issues/144](https://github.com/cryspen/libcrux/issues/144)   
* Intrinsics Spec: [https://github.com/cryspen/libcrux/issues/452](https://github.com/cryspen/libcrux/issues/452)   
* Secret Independence: [https://github.com/cryspen/libcrux/issues/453](https://github.com/cryspen/libcrux/issues/453)   
* ~~Bounded Integers: [https://github.com/hacspec/hax/issues/798](https://github.com/hacspec/hax/issues/798)~~ (we don't use bounded integers)
* Iterators:~~[https://github.com/hacspec/hax/issues/394](https://github.com/hacspec/hax/issues/394),  [https://github.com/hacspec/hax/issues/605](https://github.com/hacspec/hax/issues/605)~~ (we use https://github.com/hacspec/hax/pull/857)
* https://github.com/cryspen/libcrux/issues/566
* https://github.com/cryspen/libcrux/issues/565

ML-KEM Verification tasks:

* vector/portable  
  - [x] vector\_type.rs  
        * status: verified  
  - [ ] arithmetic.rs  
        * status: cleanup 
        * proof dependencies: bounded integers  
        * note: previous proofs in Libcrux.Kem.Kyber.Arithmetic.fst (8/12)  
        * estimate changes: small  
  - [ ] compress.rs  
        * status: cleanup 
        * proof dependencies: bounded integers  
        * note: previous proofs in Libcrux.Kem.Kyber.Compress.fst (3/5)  
        * estimate changes: small  
  - [ ] ntt.rs  
        * status: cleanup  
        * proof dependencies: bounded integers  
        * note: helpful proofs in Libcrux.Kem.Kyber.Ntt.fst (5/11)  
        * estimate changes: medium  
  - [x] sampling.rs  
        * status: panic free 
        * note: previous proofs in Libcrux.Kem.Kyber.Sampling.fst (1/1)  
        * estimate changes: small  
  - [ ] serialize.rs  
        * status: verified  
        * note: helpful proofs in Libcrux.Kem.Kyber.Serialize.fst (?/22)  
        * estimate changes: large  
* vector/avx2  
  - [ ] arithmetic.rs  
        * status: cleanup 
        * proof dependencies: libcrux-intrinsics, portable/arithmetic.rs  
[...]

PR #488, opened 2024-08-14
Add pre/post-conditions for Ind_cpa and matrix functions

mamonet:
This PR adds the necessary pre-conditions for the functions in ind_cpa.rs and matrix.rs. And makes sure that the extracted interface passes lax-check.

Part of #454


karthikbhargavan:
Let's follow the folliowng procedure for this and subsequent PRs:
- [ ] Have the preconditions for const-generics been added?
- [ ] Have the postconditions relating the code to the spec been added?
- [ ] Have all the functions been labeled as lax/panic-free (default is that they are fully verified)?
- [ ] Have the proofs from the old Kyber proof been backported yet?

Even if only one of these check boxes is ticked, it will be good to have documentation of the status of this module.

mamonet:
For now only three functions of ind_cpa have post-conditions (encrypt/decrypt/generate_keypair), these functions are utilized by ind_cca. But none of the old Kyber proofs have been backported to these functions yet.

mamonet:
I pushed a commit that set post-cinditions for Ind_cpa functions but the only one missing is for `sample_ring_element_cbd` because it doesn't have a spec function right now. It also marks which functions are lax or panic-free for that module.

karthikbhargavan (approved):

PR #489, opened 2024-08-14
fix(mlkem/fstar): fix regression introduced by hacspec/hax#864

W95Psp:
https://github.com/hacspec/hax/pull/846 makes `fstar::before` and `fstar::after` chunks of code default to implementation files (fst).

This commit backports that change and makes some code go explicitly to interfaces.


franziskuskiefer (approved):
Thanks, let's get this in when CI is happy.

PR #490, opened 2024-08-14
Manual benchmarks for PQClean

### LINKED ISSUE #487: ML-DSA: Benchmark PQClean for comparison
We should add a manual benchmark of PQClean's ML-DSA to have a fair comparison.

jschneider-bensch:
In order to compare to PQclean, I've replicated the manual benchmark approach for PQClean.
Note that the PQClean has internal randomness generation, compared to us, so I wonder if we should not include sampling the randomness in our part of the benchmark. We did not do this with the criterion benchmarks, so I didn't do it here for now.

Also, I've split the parameter set benchmarks, so you can run them individually. The bit where I include `bench_utils.rs` via `#[path=...]` is a bit nasty, let me know if I should try a different approach.

Fixes #487


libcrux-ml-dsa/benches/manual44.rs
```diff
@@ -0,0 +1,15 @@
+use libcrux_ml_dsa::{
+    ml_dsa_44::{self, MLDSA44KeyPair, MLDSA44Signature},
+    KEY_GENERATION_RANDOMNESS_SIZE, SIGNING_RANDOMNESS_SIZE,
+};
+
+use pqcrypto_dilithium;
+
+#[path = "./bench_utils.rs"]
+#[macro_use]
```
franziskuskiefer, line 9: There's no need for this anywhere 🙂 

```suggestion
```

libcrux-ml-dsa/benches/bench_utils.rs
```diff
@@ -0,0 +1,170 @@
...
+        "\t\t".to_string()
+    } else {
+        "\t".to_string()
+    };
+
+    println!("{label}:{space}{time}");
+}
+
+pub(crate) const ITERATIONS: usize = 100_000;
+#[allow(unused)]
```
franziskuskiefer, line 37: Is this `allow` needed? This should be used?

franziskuskiefer (approved):
Thanks. lgtm with a few nits.

franziskuskiefer:
> I wonder if we should not include sampling the randomness in our part of the benchmark

Yeah we should measure all the combinations eventually. Can you put that on an issue?

libcrux-ml-dsa/benches/manual44.rs
jschneider-bensch: Huh, alright! Nice to not need this.

libcrux-ml-dsa/benches/bench_utils.rs
jschneider-bensch: If I remove this I get `dead_code` warning. It's weird, right? Since I don't get the same for `ITERATIONS`.

franziskuskiefer: weird, then leave it in

PR #517, opened 2024-08-19
[ML-KEM] Address changes from FIPS 203 IPD to final version

### LINKED ISSUE #492: Domain separation in `K-PKE.KeyGen()`
https://github.com/cryspen/libcrux/blob/bf283a0934a0d386655a63954463497e753f1e37/libcrux-ml-kem/src/kem/kyber/ind_cpa.rs#L159

> Byte 33 of the input to G is the module dimension 𝑘 ∈ {2, 3, 4} ⊂ 𝔹. This is included to establish domain separation
between the three parameter sets. For implementations that use the seed in place of the private key, this ensures
that the expansion will produce an unrelated key if the seed is mistakenly expanded using a parameter set other
than the originally intended one.
-- [FIPS 203](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf#algorithm.13)

### LINKED ISSUE #493: Swapped indices when sampling matrix `A` in `K-PKE.KeyGen()` and `K-PKE.Encrypt()`
https://github.com/cryspen/libcrux/blob/bf283a0934a0d386655a63954463497e753f1e37/libcrux-ml-kem/src/kem/kyber/matrix.rs#L23-L24

> Additionally, FIPS 203 ipd had inadvertently swapped the indices of matrix 𝐀̂ in K-PKE.KeyGen and
K-PKE.Encrypt. This was changed back in the final version of ML-KEM to match CRYSTALS-KYBER.
--- (FIPS 203)[https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf#subsection.C.2]

### LINKED ISSUE #494: Update ML-KEM test vectors

jschneider-bensch:
This PR addresses
- [x] domain separation for CPA key generation seed. Fixes #492 
- [x] Indices are already in the right order. Fixes #493 
- [x] regeneration of test vectors. Fixes #494 

I also noticed that we weren't actually testing Kyber on CI, but the verified ML-KEM code (since we only had `--features kyber` in the workflow, not `--features pre-verification,kyber`). In addition, I've separated out testing of the old verified version (which is the FIPS 203 IPD), while we still have that.

### DIFF
--- a/libcrux-ml-kem/tests/kats/nistkats_mlkem_ipd_1024.json
+++ b/libcrux-ml-kem/tests/kats/nistkats_mlkem_ipd_1024.json
@@ -1,0 +1,802 @@
+[
+    {
+        "key_generation_seed": "7c9935a0b07694aa0c6d10e4db6b1add2fd81a25ccb148032dcd739936737f2d8626ed79d451140800e03b59b956f8210e556067407d13dc90fa9e8b872bfb8f",
+        "sha3_256_hash_of_public_key": "8a39e87d531f3527c207edcc1db7faddcf9628391879b335c707839a0db051a8",
+        "sha3_256_hash_of_secret_key": "ed1f6cb687c37931ea2aa80d9c956f277a9df532649661035c6e2f9872132638",
+        "encapsulation_seed": "147c03f7a5bebba406c8fae1874d7f13c80efe79a3a9a874cc09fe76f6997615",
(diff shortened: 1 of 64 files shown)


libcrux-ml-kem/tests/kats/nistkats_mlkem_ipd_1024.json
franziskuskiefer, line 1: Ah, we need them for the old code. Let's make sure to kill all of this when the new version is fully verified.

franziskuskiefer (approved):
Thanks! lgtm.

PR #518, opened 2024-08-19
ML-KEM: AVX2 implementation of serialize_11 and deserialize_11.

xvzcf:


franziskuskiefer:
We want to get #613 done first.

jschneider-bensch:
This should be re-based to current main to see what the state is relative to that.

PR #521, opened 2024-08-20
PSQ Registration protocol

jschneider-bensch:
This PR implements a Request-Response protocol for establishing and mutually registering a PQ-PSK under a storage handle designated by the responder.

The `Credential` trait does need a little work to be usable e.g. with a real signature scheme, but serves primarily as a proof of concept.


franziskuskiefer (approved):
lgtm, let's discuss beginning of September

jschneider-bensch:
I've done a tiny bit of cleanup and will be merging this, we'll discuss next steps later today.

PR #523, opened 2024-08-22
Replace HPKE with a specialization of hpke-rs

keks:
This PR removes the existing HPKE code (which lives on over at https://github.com/cryspen/hpke-spec) and instead implements a libcrux-backed provider for hpke-rs, and re-exports the types from that crate, specializing them to the provider implemented here.

This should be merged together with https://github.com/cryspen/hpke-rs/pull/68


franziskuskiefer:
@keks let's resurrect this as part of #648

PR #525, opened 2024-08-22
Rework Actions to work well with Merge Queues and Enable Benchmarks in there

keks:
This PR reworks the actions a little bit to work well with merge queues, and running the benchmarks in the merge queues. Since this is difficult to test, this PR will probably see some more commits :)

In order to get the structure, let's go over the way the rules work. There is a list of required checks, which is just a list of strings. Github now requires that at least one job of that name ran, and that all jobs of that name succeeded. It does not care whether these are actually the same jobs.

There now are three classes of workflow files:
- Sporadically run: These do not run automatically, but only when triggered manually. e.g.: flake-update.yml
- PR jobs: These are run on every PR. From now on, they also will run in merge queues. This includes stuff like running tests
- Merge-Queue-Only: This is mostly benchmarks.

So far the problem was that in the merge queue, github only runs the jobs that have a name that is listed in the required checks list. This makes it difficult to run something _only_ in the merge queue, because it would then also run in the PRs, or the PR checks wouldn't go because it's blocked on a check that won't run. But remember: github doesn't care whether these are the same checks! It only cares that there are checks that have the right name, and that the are successful, even if there is just one check of that name.

We only use the `on` setting of the workflow to configure which events trigger the jobs therein. 

We now do the following:
- we move the benchmark jobs from the PR jobs to the merge queue jobs
- the merge queue jobs are `on: [ merge_group, workflow_dispatch ]`
- we trigger PR jobs in merge queues as well and remove all the checks of event types
- we add a PR-only job that has a job with name "benchmark" that does nothing

This way, the PR checks are satisfied, and when in the merge queue, it runs the benchmarks.

Oh, and this PR also adds the benchmark html generation step.

Working towards #54.

Implementation note: We have to add
```toml
[lib]
bench = false
```
to the Cargo.toml of all the crates with criterion benchmarks that we want to track in a graph. The reason is that the graph wants the benchmarks in a special format and we have to pass `--output-format bencher`. However, by default `cargo bench` only accepts command line arguments that libtest also accepts. It seems the only way is to ask libtest to not handle benchmarks. Cf. the [criterion faq](https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options)


Cargo.toml
```diff
@@ -72,6 +73,9 @@ rand = { version = "0.8" }
 log = { version = "0.4", optional = true }
 # WASM API
 wasm-bindgen = { version = "0.2.87", optional = true }
+hpke-rs = { version = "0.2.0", optional = true, git = "https://github.com/cryspen/hpke-rs.git", branch = "keks/tests-crate", package = "hpke-rs"}
```
franziskuskiefer, line 76: These probably shouldn't be here.

.github/workflows/rust-bench.yml
```diff
@@ -0,0 +1,98 @@
...
+        if: ${{ matrix.bits == 64 && matrix.os == 'windows-latest' }}
+        shell: pwsh
+        run: |
+          echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
+          vcpkg install openssl:x64-windows-static-md
+
+      # Benchmarks ...
+
+      - name: 🏃🏻‍♀️ Benchmarks
+        run: cargo bench --verbose $RUST_TARGET_FLAG -p benchmarks -- --output-format bencher | tee bench.txt
```
franziskuskiefer, line 90: Do you want to file an issue to track what we all want to get on the benchmarks? And also other stuff like stack consumption etc.
Does the action allow running multiple times to combine the result?

franziskuskiefer (approved):
I think this is doing what it's supposed to.
Though, I have been thinking that before as well 😉 
Let's give it a try!

.github/workflows/rust-bench.yml
keks: https://github.com/cryspen/libcrux/issues/526

PR #546, opened 2024-08-30
Add randomised APIs to ml-kem

franziskuskiefer:


libcrux-ml-kem/src/kem/kyber/kyber1024.rs
```diff
@@ -44,17 +44,12 @@ pub type MlKem1024PublicKey = MlKemPublicKey<CPA_PKE_PUBLIC_KEY_SIZE_1024>;
 /// Validate a public key.
 ///
 /// Returns `Some(public_key)` if valid, and `None` otherwise.
```
jschneider-bensch, line 46:
```suggestion
/// Returns `true` if valid, and `false` otherwise.
```

libcrux-ml-kem/src/kem/kyber/kyber512.rs
```diff
@@ -42,17 +42,12 @@ pub type MlKem512PublicKey = MlKemPublicKey<CPA_PKE_PUBLIC_KEY_SIZE_512>;
 /// Validate a public key.
 ///
 /// Returns `Some(public_key)` if valid, and `None` otherwise.
```
jschneider-bensch, line 44:
```suggestion
/// Returns `true` if valid, and `false` otherwise.
```

libcrux-ml-kem/src/kem/kyber/kyber768.rs
```diff
@@ -43,17 +43,12 @@ pub type MlKem768PublicKey = MlKemPublicKey<CPA_PKE_PUBLIC_KEY_SIZE_768>;
 /// Validate a public key.
 ///
 /// Returns `Some(public_key)` if valid, and `None` otherwise.
```
jschneider-bensch, line 45:
```suggestion
/// Returns `true` if valid, and `false` otherwise.
```

jschneider-bensch (changes requested):
Looks good to me, with tiny doc nits :+1:

franziskuskiefer (commented):
Ah thanks. Copy pasta

jschneider-bensch:
Huh, somehow the extraction diff fails now because `.gitignore` is missing in the extracted version. In the extract jobs, `actions/upload-artifact@v4` has option `include-hidden-files` set to `false` so it makes sense that the `.gitignore` should be ignored. It must have been included before though, since that job would have had to fail every time previously, otherwise. Indeed, on an earlier run of that job the action did not have this option: https://github.com/cryspen/libcrux/actions/runs/10636686836/job/29488994163

jschneider-bensch (approved):

jschneider-bensch:
Ah, I see, it was [announced as a breaking change](https://github.com/actions/upload-artifact/issues/602). Still a bit weird to have that without a version bump

franziskuskiefer:
> Ah, I see, it was [announced as a breaking change](https://github.com/actions/upload-artifact/issues/602). Still a bit weird to have that without a version bump

🙄 looks like everyone is annoyed by this. @jschneider-bensch would you mind pushing a fix to this branch?

jschneider-bensch:
Ah sorry, I would have, but I've already pushed a fix to it's own PR #555

PR #554, opened 2024-09-02
Basic fuzzing for C extraction of ML-KEM768

djmdjm:
Here are some very basic fuzzers for the main operations of the ML-KEM C extraction: public key validation, encapsulate and decapsulate. They aren't at all sophisticated, but I wrote them to double-check the soundness of the C extraction wrt undefined behaviour and memory safety. No findings after running over the weekend on 128 cores :)


djmdjm:
BTW I won't be at all offended if you reject this as unsuitable for the project; like I said, this was mostly for my own paranoia.

franziskuskiefer:
Nice, thanks! 

> BTW I won't be at all offended if you reject this as unsuitable for the project; like I said, this was mostly for my own paranoia.

Not at all. We'll need to find time to look at it. But this is great to have here!

franziskuskiefer (approved):
Thanks, this is great!
I finally go around to looking at it. I'll merge as is. I only formatted and changed to clang-18.

PR #556, opened 2024-09-03
Remove unsafe code from include to avoid being rejected by hax.

maximebuyse:
Fixes build failure that happened here: https://github.com/cryspen/libcrux/actions/runs/10677857822/job/29593758713?pr=555
This comes after merging https://github.com/hacspec/hax/pull/867 that rejects any AST containing unsafe blocks.
We still extract the interfaces for these functions.


jschneider-bensch (approved):
Thanks :pray:

PR #524, opened 2024-08-22
ML-DSA 44 instantiations and multiplexing

franziskuskiefer:
- add build.rs for feature detection
- add instantiations for 44
- add neon sha3

Towards #275

PR #558, opened 2024-09-03
Allow hax extraction for ML-DSA

jschneider-bensch:
This PR will make it possible to extract ML-DSA to F*.

- [x] Cast `match` scrutinees to known-size types as a workaround for https://github.com/hacspec/hax/issues/464
- [x] Replace `loop/continue/break`-style rejection sampling with a bounded loop. As @xvzcf had already pointed out, the probability of the bounded rejection sampling to not succeed within 576 attempts is smaller than 2^128. Nonetheless, I've introduced a `SigningError` type to indicate this, so we don't panic, even in this most unlikely case.
- [ ] `RefMut` issues in `encoding::signature`
- [ ] `FunctionalizeLoops` issue in `encoding::signature`

### DIFF
--- a/libcrux-ml-dsa/src/ml_dsa_generic.rs
+++ b/libcrux-ml-dsa/src/ml_dsa_generic.rs
@@ -5,7 +5,7 @@ use crate::{
         decompose_vector, make_hint, power2round_vector, use_hint, vector_infinity_norm_exceeds,
     },
     constants::*,
-    encoding,
+    encoding::{self, commitment},
     hash_functions::{shake128, shake256},
     matrix::{
         add_vectors, compute_A_times_mask, compute_As1_plus_s2, compute_w_approx, subtract_vectors,
@@ -100,6 +100,11 @@ pub enum VerificationError {
     CommitmentHashesDontMatchError,
 }
 
+#[derive(Debug)]
+pub enum SigningError {
+    RejectionSamplingError,
+}
+
 #[allow(non_snake_case)]
 pub(crate) fn sign<
     SIMDUnit: Operations,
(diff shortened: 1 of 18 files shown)


libcrux-ml-dsa/src/ml_dsa_generic.rs
```diff
@@ -224,44 +228,63 @@ pub(crate) fn sign<
...
+                        &challenge_times_t0,
+                    );
+                    let (hint_candidate, ones_in_hint) = make_hint::<SIMDUnit, ROWS_IN_A, GAMMA2>(
+                        w0_minus_c_times_s2_plus_c_times_t0,
+                        commitment,
+                    );
+
+                    if ones_in_hint > MAX_ONES_IN_HINT {
+                    } else {
+                        attempt = 576; // exit loop now
```
franziskuskiefer, line 264: Let's make the 576 a constant and use that instead everywhere.

franziskuskiefer (approved):
Thanks. Two nits, then we can get this in.

franziskuskiefer:
Oh and I created a conflict with #524 that need to get resolved.

jschneider-bensch:
After merging and resolving, more of the AVX2 implementation seems to be going through hax :thinking:
To fix the extraction I had to update the intrinsics interfaces for hax and cast match scrutinees in more places.

libcrux-intrinsics/src/avx2_extract.rs
franziskuskiefer, line 1: Ah extraction is always using this one? That shouldn't be necessary. But now we have it 🙂

franziskuskiefer (approved):

PR #559, opened 2024-09-03
Port proofs for constant_time_ops.rs

mamonet:
This PR apply the proofs from `Libcrux.Kem.Kyber.Constant_time_ops.fst` in constant_time_ops.rs for all functions. The invariant loop feature is used for `Hax.Folds.fold_range` loops.


libcrux-ml-kem/src/constant_time_ops.rs
```diff
@@ -28,39 +71,125 @@ fn is_non_zero(value: u8) -> u8 {
 
 /// Return 1 if the bytes of `lhs` and `rhs` do not exactly
 /// match and 0 otherwise.
-#[cfg_attr(hax, hax_lib::requires(
-    lhs.len() == rhs.len()
-))]
+#[hax_lib::requires(lhs.len() == rhs.len())]
+#[hax_lib::ensures(|result| fstar!("Hax_lib.implies ($lhs =. $rhs <: bool)
```
karthikbhargavan, line 75: Here, since you are in F* anyway, why not use "($lhs == $rhs) ==> "

libcrux-ml-kem/src/constant_time_ops.rs
```diff
@@ -28,39 +71,125 @@ fn is_non_zero(value: u8) -> u8 {
...
+    (fun temp_0_ ->
```
karthikbhargavan, line 76: Same here, once you are in F* you can simplify this significantly.

karthikbhargavan (changes requested):

karthikbhargavan (approved):

PR #574, opened 2024-09-12
[ML-DSA] Repair / Extend multiplexing

jschneider-bensch:
I was wondering why seemingly none of the changes in the NTT I tried had any effect on ML-DSA 44, until I actually looked at the functions that are called using samply. Turns out we were using the portable NTT in ML-DSA 44 because the instantiation had a typo.
I've taken this opportunity to extend multiplexing to the other variants as well.


franziskuskiefer (approved):
Nice catch. Thanks!

PR #576, opened 2024-09-13
Make ntt panic free

mamonet:
This PR makes generic ntt.rs panic-free, it also removes the module from Makefile.


mamonet:
I pushed a commit that makes invert_ntt panic free but it has an assume here https://github.com/cryspen/libcrux/pull/576/files#diff-bb32900a2db27f0a645fa56145df42a13ad97c6e29e12ea3cd75c74f77e427d2R127-R128
I wonder if there's an easy way to resolve that assume

karthikbhargavan (approved):
Fixed up some things.

Issue #571, opened 2024-09-11
improve NTT performance of avx2 ml-dsa


franziskuskiefer:

jschneider-bensch:
We found that playing around with different shufflings made essentially no difference, maybe with the potential exception of keeping vectors in the NTT domain in a shuffled state to avoid some shufflings altogether. This would require to touch all places where NTT domain vectors are handled and is probably not worth it in terms of performance, since there are easier changes to be made still, e.g. applying the butterfly optimization to the inverse NTT.

PR #584, opened 2024-09-16
[ML-DSA] AVX2 performance improvements in NTT

jschneider-bensch:
- **Forward NTT**:
  - [x] Process two rounds of butterflies at layers 2,1 & 0
- [ ] **Inverse NTT**: Anything we do forward can be done on the inverse as well

With this, I have on my machine:
```
(libcrux) KeyGen 65 simd256:	52 μs
(libcrux) Sign 65 simd256:	148 μs
(libcrux) Verify 65 simd256:	46 μs

(pqclean) KeyGen 65:	34 μs
(pqclean) Sign 65:	94 μs
(pqclean) Verify 65:	38 μs
```

This PR works towards #571


libcrux-ml-dsa/src/simd/portable/ntt.rs
```diff
@@ -1,67 +1,77 @@
...
-pub fn ntt_at_layer_0(
+pub fn simd_unit_ntt_at_layer_0(
     mut simd_unit: PortableSIMDUnit,
     zeta0: i32,
     zeta1: i32,
     zeta2: i32,
     zeta3: i32,
 ) -> PortableSIMDUnit {
-    let t = montgomery_multiply_fe_by_fer(simd_unit.coefficients[1], zeta0);
+    let t = arithmetic::montgomery_multiply_fe_by_fer(simd_unit.coefficients[1], zeta0);
```
franziskuskiefer, line 18: What's the reason to use the path here instead of importing?

franziskuskiefer (approved):
Thanks!

libcrux-ml-dsa/src/simd/portable/ntt.rs
jschneider-bensch: Ah, I don't see a reason to avoid the import, thanks :+1:

PR #592, opened 2024-09-20
Private keys should persist between invocations

djmdjm:
Missing `static` qualifier on the private keys used for fuzzing.

Spotted by the Qualys Security Advisory team


franziskuskiefer (approved):
Oh, nice catch.

PR #594, opened 2024-09-22
Lax Checking ML-DSA

karthikbhargavan:
This PR extracts F* code for ML-DSA and lax-checks it.
To bypass certain F* limitations, the source Rust code is restructured in places.


jschneider-bensch (approved):
The ML-DSA Rust changes look fine to me :+1:
I've just addressed some warnings about unexpected `cfg(hax)`.

PR #609, opened 2024-10-01
[ML-DSA] Add domain separation for key generation

jschneider-bensch:
In key generation, the variant parameters are prepended to the key generation randomness before extracting the initial seed. This is supposed to prevent issues when the same seed is used for key generation in different parameter sets.


libcrux-ml-dsa/src/ml_dsa_generic.rs
```diff
@@ -52,7 +52,10 @@ pub(crate) fn generate_key_pair<
 ) -> ([u8; SIGNING_KEY_SIZE], [u8; VERIFICATION_KEY_SIZE]) {
     // 128 = SEED_FOR_A_SIZE + SEED_FOR_ERROR_VECTORS_SIZE + SEED_FOR_SIGNING_SIZE
     let mut seed_expanded = [0; 128];
-    Shake256::shake256::<128>(&randomness, &mut seed_expanded);
+    let mut shake = Shake256Absorb::new();
+    shake.absorb(&[ROWS_IN_A as u8, COLUMNS_IN_A as u8]);
```
franziskuskiefer, line 56: This appears to be in the wrong order. First, the randomness goes in, then the parameters.

franziskuskiefer (changes requested):
I threw a vector from https://github.com/usnistgov/ACVP-Server/blob/master/gen-val/json-files/ML-DSA-keyGen-FIPS204/prompt.json on this and it failed. With the fix in the comment it passes.

franziskuskiefer (approved):
Thanks!

PR #618, opened 2024-10-07
ACVP test vectors for ML-KEM

jschneider-bensch:
I've taken up Franziskus' branch that did KeyGen testing with the ACVP vectors and extended it to run the Encaps/Decaps tests as well.


libcrux-ml-kem/tests/acvp.rs
```diff
@@ -0,0 +1,345 @@
...
+                        .find(|t| {
+                            if let EncapDecapResult::EncapResult { tcId, .. } = t {
+                                *tcId == test.tcId
+                            } else {
+                                false
+                            }
+                        })
+                        .unwrap();
+
+                    if let EncapDecapResult::EncapResult { c, k, .. } = expected_result {
```
franziskuskiefer, line 252: You could use `let else` here instead to make it more readable.
```suggestion
                    let EncapDecapResult::EncapResult { c, k, .. } = expected_result else {
                        unreachable!()
                    };
```

libcrux-ml-kem/tests/acvp.rs
```diff
@@ -0,0 +1,345 @@
...
+                        let ek = test.ek;
+                        let randomness = test.m;
+                        match parameter_set.as_str() {
+                            #[cfg(feature = "mlkem512")]
+                            "ML-KEM-512" => {
```
franziskuskiefer, line 257: Nothing for here. But we should really add that dynamic API on top.

libcrux-ml-kem/tests/acvp.rs
```diff
@@ -0,0 +1,345 @@
+#![cfg(all(
+    feature = "pre-verification",
+    any(feature = "mlkem512", feature = "mlkem768", feature = "mlkem1024",)
+))]
+
+use serde::{de::DeserializeOwned, Deserialize};
+use serde_json::Value;
```
franziskuskiefer, line 7: Unused import

franziskuskiefer (approved):
Thanks, lgtm with two small nits.

libcrux-ml-kem/tests/acvp.rs
jschneider-bensch: Ah, that's a bit better, thanks! It would be even better if I found a nicer way to deserialize these test vectors, but I didn't come up with anything better than an untagged enum for now. (Could use the same type for both variants and make `c` optional, but I don't think that makes it easier to understand :unamused:.)

libcrux-ml-kem/tests/acvp.rs
jschneider-bensch: Can you elaborate? Do you mean provided randomness vs internally generated?

PR #619, opened 2024-10-08
ACVP test vectors for ML-DSA

jschneider-bensch:
This PR adds the ACVP test vectors for ML-DSA.
Since the vectors are for the `_internal` versions of signing and verification I had to add a feature `acvp` which exposes these in a way that is compatible with the testing regime, i.e. without any `domain_separation_context` which is assumed to be already folded into the message during ACVP testing.


libcrux-ml-dsa/src/ml_dsa_generic.rs
```diff
@@ -417,9 +420,12 @@ pub(crate) fn sign_internal<
 /// This corresponds to line 6 in algorithm 7 in FIPS 204 (line 7 in algorithm
 /// 8, resp.).
 ///
-/// Applies domain separation and length encoding to the context string,
+/// If `domain_separatino_context` is supplied, applies domain
```
franziskuskiefer, line 423: `domain_separation_context`

franziskuskiefer (approved):
Thanks, lgtm with a few nits.

libcrux-ml-dsa/src/ml_dsa_generic.rs
jschneider-bensch: Thanks!

franziskuskiefer (approved):

PR #621, opened 2024-10-08
Fix for latest eurydice 

protz:
This contains the fixes for libcrux to work with the latest version of the toolchain -- this just needs additional work to update libcrux's CI


protz:
Upgrading to latest Eurydice *should* give more stable hash names.

franziskuskiefer:
I fixed this in #622

PR #622, opened 2024-10-09
Update C extraction to toolchanges

franziskuskiefer:
This PR also disables the SHA3 benchmarks that used internal functions.


libcrux-ml-kem/c/benches/sha3.cc
```diff
@@ -58,29 +58,29 @@ sha3_512_64(benchmark::State &state)
...
-    libcrux_sha3_generic_keccak_squeeze_first_three_blocks_ed(&st, out);
-
-    for (auto _ : state)
-    {
-        auto st = libcrux_sha3_avx2_x4_incremental_init();
-        libcrux_sha3_generic_keccak_absorb_final_7f(&st, last);
-        libcrux_sha3_generic_keccak_squeeze_first_three_blocks_ed(&st, out);
-    }
-}
+// __attribute__((target("avx2"))) static void
```
jschneider-bensch, line 61: Might we want to re-enable this later?

jschneider-bensch (approved):

libcrux-ml-kem/c/benches/sha3.cc
franziskuskiefer: Yeah, we may want to expose this properly. The issue is that the name changes of internal functions. And we misuse the code here.

PR #636, opened 2024-10-17
ML-KEM: AVX2 target feature edition

jschneider-bensch:
This is a first step of addressing #633.
It should get us most of the way back to reasonable performance, e.g. on my machine the difference between a `RUSTFLAGS="-C target-feature=+avx2" cargo bench --features pre-verification` and simply `cargo bench --features pre-verification` seems to be within 10-20% across keygen, encaps & decaps. 

So, there seem to be more things still, and I didn't particularly look at key validation, yet. But it should be a good first step, anyway.

### DIFF
--- a/libcrux-ml-kem/benches/ml-kem.rs
+++ b/libcrux-ml-kem/benches/ml-kem.rs
@@ -107,8 +107,7 @@ pub fn pk_validation(c: &mut Criterion) {
     init!(mlkem1024, "PK Validation", c);
 }
 
-#[target_feature(enable = "avx2")]
-pub unsafe fn encapsulation(c: &mut Criterion) {
+pub fn encapsulation(c: &mut Criterion) {
     macro_rules! fun {
         ($name:expr, $p:path, $group:expr) => {
             $group.bench_function(format!("libcrux {} (external random)", $name), |b| {
@@ -161,8 +160,7 @@ pub unsafe fn encapsulation(c: &mut Criterion) {
     init!(mlkem1024, "Encapsulation", c);
 }
 
-#[target_feature(enable = "avx2")]
-pub unsafe fn decapsulation(c: &mut Criterion) {
+pub fn decapsulation(c: &mut Criterion) {
     macro_rules! fun {
         ($name:expr, $p:path, $group:expr) => {
             $group.bench_function(format!("libcrux {}", $name), |b| {
@@ -221,8 +219,8 @@ pub unsafe fn decapsulation(c: &mut Criterion) {
 pub fn comparisons(c: &mut Criterion) {
     pk_validation(c);
     key_generation(c);
-    unsafe { encapsulation(c) };
-    unsafe { decapsulation(c) };
+    encapsulation(c);
+    decapsulation(c);
 }
 
 criterion_group!(benches, comparisons);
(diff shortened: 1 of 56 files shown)


libcrux-ml-kem/benches/ml-kem.rs
```diff
@@ -107,7 +107,8 @@ pub fn pk_validation(c: &mut Criterion) {
     init!(mlkem1024, "PK Validation", c);
 }
 
-pub fn encapsulation(c: &mut Criterion) {
+#[target_feature(enable = "avx2")]
```
franziskuskiefer, line 110: These changes need to get undone in this file

libcrux-ml-kem/src/hash_functions.rs
```diff
@@ -213,44 +213,48 @@ pub(crate) mod avx2 {
...
-            }
-            3 => {
-                x4::shake256(
-                    &input[0], &input[1], &input[2], &input[0], &mut out0, &mut out1, &mut out2,
-                    &mut out3,
-                );
-                out[0] = out0;
-                out[1] = out1;
-                out[2] = out2;
+        #[target_feature(enable = "avx2")]
```
franziskuskiefer, line 216: I don't think this made a difference?

libcrux-ml-kem/src/ind_cca/instantiations.rs
```diff
@@ -405,9 +415,10 @@ macro_rules! instantiate {
 // Portable generic implementations.
 instantiate! {portable, crate::vector::portable::PortableVector, crate::hash_functions::portable::PortableHash<K>}
 
-// AVX2 generic implementation.
+// // AVX2 generic implementation.
+// instantiate! {avx2, crate::vector::SIMD256Vector, crate::hash_functions::avx2::Simd256Hash}
```
franziskuskiefer, line 419:
```suggestion
```

libcrux-ml-kem/src/lib.rs
```diff
@@ -68,7 +68,7 @@ analogously for encapsulation and decapsulation."##
 
 #![no_std]
 #![deny(missing_docs)]
-#![forbid(unsafe_code)]
+// #![forbid(unsafe_code)]
```
franziskuskiefer, line 71: We should warn here at least.

libcrux-ml-kem/src/mlkem768.rs
```diff
@@ -56,11 +56,13 @@ macro_rules! instantiate {
             ///
             /// Returns `true` if valid, and `false` otherwise.
             pub fn validate_public_key(public_key: &MlKem768PublicKey) -> bool {
+                #[allow(unused_unsafe)]
```
franziskuskiefer, line 59: We have all these unsafes everywhere now. Shall we put safe wrappers into `avx2.rs` instead? Then the `unsafe` is contained to that file.

libcrux-ml-kem/src/hash_functions.rs
jschneider-bensch: You're right. I've reverted that

libcrux-ml-kem/benches/ml-kem.rs
jschneider-bensch: Done

franziskuskiefer (approved):
Lets get this in

PR #642, opened 2024-10-21
ML-DSA: AVX2 target feature

jschneider-bensch:
This PR performs the same kind of restructuring to ML-DSA as #636 does to ML-KEM, i.e. once we know we're running on an AVX2 machine, manually enable AVX2 specific optimizations using `#[target_feature(enable = "avx2")]` and then inline most functions that are further down in the call-tree.
Doing this should get us close to compiling the whole crate using `RUSTFLAGS="-C target-feature=+avx2"` in terms of performance.


jschneider-bensch:
The F* lax checking changes have regressed performance again. Looking into why this happened.

jschneider-bensch:
I have another version of this branch at https://github.com/cryspen/libcrux/tree/jonas/ml-dsa-target-feature-backup which is pre-rebase on top of the laxing changes. There, reverting the above commit containing SHA-3 changes fully restores performance, but leads to a stack overflow. Reverting the commit on this branch does not seem to help performance.

@franziskuskiefer @karthikbhargavan Shall we try and get this PR in with somewhat degraded performance (e.g. for me the difference between the overflowing performant code and the laxing, non-overflowing code is ~10x), and look into fully restoring performance in a follow-up?

PR #649, opened 2024-11-03
Fix big-endian support in cg

### LINKED ISSUE #643: ML-KEM cg/ extraction broken on big-endian systems
The `libcrux-ml-kem/cg` extraction is broken on big-endian systems. The root cause seems to be [these functions](https://github.com/cryspen/libcrux/blob/13a5dead2cffbec8e7653b532a7e12545801e0a5/libcrux-ml-kem/cg/eurydice_glue.h#L132-L145)  in `eurydice_glue.h`:

```C
static inline void core_num__u64_9__to_le_bytes(uint64_t v, uint8_t buf[8]) {
  memcpy(buf, &v, sizeof(v));
}
static inline uint64_t core_num__u64_9__from_le_bytes(uint8_t buf[8]) {
  uint64_t v;
  memcpy(&v, buf, sizeof(v));
  return v;
}

static inline uint32_t core_num__u32_8__from_le_bytes(uint8_t buf[4]) {
  uint32_t v;
  memcpy(&v, buf, sizeof(v));
  return v;
}
```
which don't do any byte-swapping. We just fixed them in OpenSSH using `endian.h` macros: https://github.com/openssh/openssh-portable/commit/cf3e48ee8ba1beeccddd2f203b558fa102be67a2#diff-36ddc54729ced4a7ff0c5351a9fa6db5229d806a9268fc33f10c61d6cef9811e but it would be nice for them to be fixed upstream.

franziskuskiefer:
The C code in the cg directory was minimal and dropped all the endianness support.

This PR also disables benchmarks by default for the C code. They can be enabled with `LIBCRUX_BENCHMARKS=1`.

Fixes #643

### DIFF
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -162,4 +162,4 @@ jobs:
 
       - name: 🏃🏻‍♀️ Test
         run: ./build/ml_kem_test
-        if: ${{ matrix.os == 'windows-latest' }}
+        if: ${{ matrix.os != 'windows-latest' }}
--- a/.github/workflows/c.yml
+++ b/.github/workflows/c.yml
@@ -118,7 +118,7 @@ jobs:
           
       - name: 🔨 Build
         run: |
-          cmake -B build
+          LIBCRUX_BENCHMARKS=1 cmake -B build
           cmake --build build
 
       - name: 🏃🏻‍♀️ Test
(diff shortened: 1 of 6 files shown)


jschneider-bensch (changes requested):
Looks good, overall, just needs an update of the `cg` extraction.

.github/workflows/c.yml
```diff
@@ -159,14 +159,7 @@ jobs:
...
-      # FIXME: Benchmark build for cg on Windows CI are not working right now.
-      # - name: 🏃🏻‍♀️ Test (cg)
-      #   working-directory: libcrux-ml-kem/cg
-      #   run: ./build/Debug/ml_kem_test
-      #   if: ${{ matrix.os == 'windows-latest' }}
 
       - name: 🏃🏻‍♀️ Test
         run: ./build/ml_kem_test
-        if: ${{ matrix.os != 'windows-latest' }}
+        if: ${{ matrix.os == 'windows-latest' }}
```
jschneider-bensch, line 165: This doesn't work, you probably wanted to do it the same way as for the regular C extraction?

jschneider-bensch (approved):

PR #658, opened 2024-11-06
[CI] Hax extract on different architectures

jschneider-bensch:
This is to catch a potential feature misconfiguration during extraction.

### DIFF
--- a/.github/workflows/hax.yml
+++ b/.github/workflows/hax.yml
@@ -21,7 +21,62 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  hax:
+  hax-extract:
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


franziskuskiefer:
@jschneider-bensch what's the state here? Looks like CI didn't go through.

jschneider-bensch:
Right, it looks like the Nix-based setup for hax on CI here does not work on all the platforms where I wanted to confirm we can extract, so I'll have to set up hax manually for these platforms.

I still think it would be good to know that extraction works on common platforms, but it would also be more churn on the CI, the way it is configured here. So I would actually like to run it on the merge queue instead. What do you think?

franziskuskiefer:
👍🏻  Yes, running it on the merge queue and without nix sounds good.

Issue #648, opened 2024-10-31
[Meta] Initial pure Rust release


franziskuskiefer:
A pure Rust libcrux release in early 2025 with a limited set of primitives for use in different project.

Tracking of primitives backed by hacl-rs has been moved to #537.

keks:
Is there anything in ML-DSA and ML-KEM itself that is not pure rust? My understanding is that we SHA3 and SHAKE are C right now, but other than that we are fine?

franziskuskiefer:
ML-KEM and ML-DSA are pure Rust. We use the Rust SHA3.

franziskuskiefer:
hmac could do with a `verify` function that tries to do constant time comparison of the two mac values.

franziskuskiefer:
Closing this since we have all crates out already. Filing follow ups for the meta libcrux crate

PR #659, opened 2024-11-06
Back implementations of SHA2, HMAC-SHA1, HMAC-SHA2 and HKDF-SHA2 by hacl-rs

keks:
This PR replaces some use of hacl-c by hacl-rs.

The Rust files in the directories `libcrux-hacl-rs` and `libcrux-hacl-rs-xxx` are autogenerated, with two caveats:

- some have been rustfmt'd
- I added some `use` items to modules inside the same crate in some cases
- Some modules paths changed when copying them into this crate, so I had to adapt that sometimes.

The reason for the last two items is that upstream they define several crates, and here I wanted to keep the number crates low. So when e.g. they use `crate::bignum_base` in their `bignum` crate, and I move that crate into a module `bignum` inside our `libcrux-hacl-rs` crate, this becomes `crate::bignum::bignum_base`.

Maybe we should ask if they can put everything in a single crate (except the proc-macro of course), or maybe we can split these into separate crates. Not sure what is better.

In some places it also changed the API a little:

### Pass in &mut [u8; N] instead of returning [u8; N]

Seems like it’s the more general thing, even though the API is a bit more clunky. We can implement the returning API on top of the borrowing API, as well.

### Trait and Structs

Use structs and traits for the implementations rather than modules. Somehow doing it this way seemed cleaner to me, but I can understand if that is not how it should be. More of a conversation starter, the old structure is still around. Happy to hear feedback around this.

### Renamed “tag length” (in comments, $tag_len, …) to “hash length”

That’s what the RFC calls it.

### removed one of the two errors

Until now, we had (a) libcrux_hkdf::Error and (b) libcrux_hkdf::hacl_hkdf::Error. However, (a) did not contain all error conditions, but was the only that was exposed, so we returned “too long okm” when in fact the input buffer was too large. Now we just have a single error, with both error conditions, and that’s just used everywhere.

## HMAC

Nothing besides the change from “return array” to “take &mut”.

### Removed the additional layering between the public facing types and the state types

Previously, it looked like the inner type would only hide the unsafe-ness of the hacl-c implementation, and then we would just forward most of the API to the used in a wrapper around that. Now that this is safe, the utility became even more questionable, so I got rid of it. The digest module now uses hacl-rs directly.

## ed25519

Nothing besides the change from “return array” to “take &mut”.


keks:
Alright, I added all the versions and description fields and renamed the crates.

I also made it so the hacl-rs crate only contains code that is accessed by at least two actual crates. That means that most of the algorithm-specific crate code is now in the actual crates, and it is only exposed if the `hacl` feature is set (which it is by default, but that might change in the future).

Note that one outlier here is curve25519, which has the implementation in the hacl-rs crate, because ed25519 also depends on it. I suppose ed25519 could also depend on curve25519 - Let me know if you prefer that and I'll make the change.

I also introduced another feature in the crates: `portable_hacl`. The idea here is that it indicates that "the hacl code is used to provide a portable implementation". The idea is that we can then add e.g. an "avx_jasmin" or so to provide non-portable implementations, and then have code that does feature detection at runtime. Not sure that is valuable, but that is the idea behind the feature.

I also rebased on main.

hacl-rs/src/bignum.rs
```diff
@@ -0,0 +1,12 @@
+pub mod bignum;
```
franziskuskiefer, line 1: This generates a clippy warning
```
warning: module has the same name as its containing module
```

curve25519/src/impl_hacl.rs
```diff
@@ -0,0 +1,18 @@
...
+
+/// Implementation of Curve25519 backed by Hacl.
+pub struct HaclCurve25519;
+
+impl Curve25519 for HaclCurve25519 {
+    fn secret_to_public(pk: &mut [u8; PK_LEN], sk: &[u8; SK_LEN]) {
+        crate::hacl::secret_to_public(pk, sk)
+    }
+
+    fn ecdh(out: &mut [u8; SHK_LEN], pk: &[u8; PK_LEN], sk: &[u8; SK_LEN]) -> Result<(), Error> {
```
franziskuskiefer, line 11: The glue code that's calling into the hacl code should have enough comments to understand how it ensures that all the preconditions of the underlying, verified code are ensured. Since we're in all Rust, that's relatively easy. But let's make sure that we have comments on the glue explaining how it works.

ed25519/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#[cfg(feature = "hacl")]
+pub mod hacl {
```
franziskuskiefer, line 2: We shouldn't expose any underlying names. The caller doesn't care where the code may come from. That was my last bullet point in the previous review. This is again for all crates.
The crate should expose one libcrux API, independent of the implementation(s) underneath.
You could solve that by renaming things, I don't think there's the need to change anything else.

hacl-rs/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// Utility modules. In the generated hacl-rs, these are individual crates.
```
franziskuskiefer, line 1: We should add a warning into the docs or somewhere else, that this crate is only for internal use.

franziskuskiefer, line 1: Also, let's put somewhere the hacl-star revision this code is coming from.

franziskuskiefer (changes requested):
Thanks, only a few things we need to fix to get this first version in.

Follow-ups, please transfer to sub issues in #648 where there isn't one already
- Add labels for verification status. All the crates here should get verified, potentially with a hacl-rs note.
- The code is missing all the inlines. That's also why you didn't run into build performance issues. We should annotate all functions with a must inline in the hacl-rs code. Otherwise this code will be slow.
- Related: Let's make sure to have basic benchmarks on everything. There are benchmarks for most things on the main libcrux crate that we could re-use.
- Clearly document what's hand written and what's generated code.

ed25519/src/lib.rs
keks: We need to in some cases, because e.g. the hmac crate needs access to sha2. But yes, I can drop it where we don't specifically need it.

curve25519/src/impl_hacl.rs
keks: Can you expand on what you mean? I am just passing on arguments. I could say something like "the documentation of HACL says that this should be a 64-byte buffer and I pass in a [u8; 64], so the type system enforces that the inputs are correct", but I am not sure that is what you mean.

franziskuskiefer: Yes, the hacl code should have pretty good comments already on that. I just want to make sure we have a little in here so that we know what's going on. Thanks for adding some.
When you'd use doc comments, they'd show up in the IDE.

ed25519/src/impl_hacl.rs
```diff
@@ -0,0 +1,54 @@
...
+    SigningError,
+    InvalidSignature,
+}
+
+/// The hacl implementation requires that
+/// - the private key is a 32 byte buffer
+/// - the signature is a 64 byte buffer,
+/// - the payload buffer is not shorter than payload_len.
+///
+/// We enfore the first two using types, and the latter by using `payload.len()` and `payload_len`.
```
franziskuskiefer, line 12:
```suggestion
/// We enforce the first two using types, and the latter by using `payload.len()` and `payload_len`.
```

ed25519/src/impl_hacl.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    Ok(signature)
+}
+
+/// The hacl implementation requires that
+/// - the public key is a 32 byte buffer
+/// - the signature is a 64 byte buffer,
+/// - the payload buffer is not shorter than payload_len.
+///
+/// We enfore the first two using types, and the latter by using `payload.len()` and `payload_len`.
```
franziskuskiefer, line 32:
```suggestion
/// We enforce the first two using types, and the latter by using `payload.len()` and `payload_len`.
```

ed25519/src/lib.rs
franziskuskiefer: Right. I think we should have a good way of marking internal APIs that we only expose because another libcrux crate needs them. Maybe putting them behind an additional feature is the easiest.
But let's file a follow up for that and merge this as is for now.

hacl-rs/src/bignum.rs
```diff
@@ -0,0 +1,12 @@
+pub mod base;
+pub mod bignum256;
+pub mod bignum256_32;
+pub mod bignum32;
+pub mod bignum4096;
+pub mod bignum4096_32;
+pub mod bignum64;
+pub mod bignum_base;
+
+pub mod test {
```
franziskuskiefer, line 10: This kind of code we may want to delete when we know that we don't update a lot anymore.

PR #660, opened 2024-11-06
Basic ml kem fuzzing

### LINKED ISSUE #129: Add Kyber fuzzing to CI
We should clean up the fuzzing code and make it more comprehensive. We could fuzz the Kyber code for a couple of minutes on each PR, and run a longer job where we fuzz the code for an hour or so every night.

franziskuskiefer:
This adds some basic fuzzing for the Rust ml-kem code, also on ci.

closes #129


.github/workflows/mlkem.yml
```diff
@@ -168,3 +168,35 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: 🛠️ Setup Rust Nightly
+        run: rustup toolchain install nightly
+
+      - name: Update dependencies
+        run: cargo update
+
+      - run: 🏃🏻‍♀️ Decaps
```
jschneider-bensch, line 195:
```suggestion
      - name: 🏃🏻‍♀️ Decaps
```
Ditto the places where `run` is duplicated below.

.github/workflows/mlkem.yml
jschneider-bensch, line 1: Isn't it necessary to install `cargo install cargo-fuzz`?

jschneider-bensch (changes requested):
With the CI errors fixed, I think it's good!
Do we also want to fuzz the other parameter sets?

.github/workflows/mlkem.yml
franziskuskiefer: Ah, of course. That somehow got lost ...

jschneider-bensch (approved):

PR #671, opened 2024-11-13
More efficient butterfly in inverse NTT layers 0-2

jschneider-bensch:
This PR applies the optimization of batching two butterflies in AVX2 for layers 0 through 2 of the inverse NTT, as is already done on the forward NTT.
I've also pushed all of the inverse NTT into the respective instantiation modules with the hope of improving optimizations done by the compiler.


franziskuskiefer (approved):
Thanks, let's get this in.

The main slow down is from #646 not propagating inlines for the hashing. Simply adding them back gets us into trouble with the stack again, but back to what I've seen after my changes the last time.

PR #678, opened 2024-11-14
Improve the flake

Nadrieril:
A few flake improvements that will make it possible to remove the `Cargo.lock` and to use it more effectively in CI.


W95Psp (approved):
Thanks @Nadrieril, that looks great to me :)

PR #680, opened 2024-11-14
[ML-DSA] Measure stack frame size on CI

jschneider-bensch:
This PR contains a script for obtaining stack frame measurements in ML-DSA and checking they don't exceed a given limit.
It also introduces a new job on the ML-DSA CI for running the script and

From the script docs:
 This script runs `cargo stack-sizes` to obtain stack frame sizes for
 functions in the crate. It
  - temporarily installs the cargo stack-sizes tool,
  - temporarily reconfigure the crate to the requirements of cargo
    stack-sizes (installs a minimal Rust nightly toolchain and sets a
    bogus `version` in Cargo.toml),
  - runs the stack-sizes tool on a given example.

 The script will then print all stack frame sizes it collects in
 increasing order and will exit normally if the `FRAME_SIZE_LIMIT` is
 not exceeded by any of them. If the limit is exceeded the script
 will error and print the name of the offending function.

 Before exiting, the script will restore the previous state of the
 crate and uninstall the stack-sizes tool.

 Usage:
```
    ./stack-sizes.sh <example>
```
 where `<example>` is an example program as you would run it with `cargo
 run --example <example>`.

### DIFF
--- a/.github/workflows/mldsa.yml
+++ b/.github/workflows/mldsa.yml
@@ -104,3 +104,19 @@ jobs:
         run: |
           cargo clean
           cargo bench --no-run
+
+  stack-size:
+        runs-on: ubuntu-latest
+        defaults:
+          run:
+            shell: bash
+            working-directory: libcrux-ml-dsa
+        steps:
+            - uses: actions/checkout@v4
+        
+            - name: 📚 Check stack sizes
+              run: |
+                cargo clean
+                chmod +x ./stack-sizes.sh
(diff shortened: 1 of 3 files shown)


jschneider-bensch:
After some testing and discussion, the tool might not be as precise as we hoped. ([Some discussion](https://github.com/rust-lang/rust/issues/54192) on this.)
Closing this in favor of a different approach for now.

PR #689, opened 2024-11-18
Delete Cargo.lock

franziskuskiefer:
This creates a ton of PRs and local git conflicts. Lets drop this unless someone has strong opinions on this.


keks (approved):
As per our discussion earlier, I think we can merge this :+1:

PR #694, opened 2024-11-21
Swap order of hybrid KEM components for `X25519MlKem768`

jschneider-bensch:
The parts of the public key, the shared secret and the ciphertext are in reverse order: https://datatracker.ietf.org/doc/html/draft-kwiatkowski-tls-ecdhe-mlkem-02#name-construction


franziskuskiefer (approved):
Thanks

franziskuskiefer (approved):

PR #699, opened 2024-11-28
Temporary fixes for performance/build times/stack usage for ML-DSA AVX2

jschneider-bensch:
This PR presents a compromise for achieving passable levels of performance and build times in Release builds, while somewhat cutting back on stack usage in Debug build.


franziskuskiefer (approved):
Thanks! Let's get this in and move to more productive things.

PR #701, opened 2024-11-30
[ML-DSA] Inline hash_functions.rs

franziskuskiefer:
Adding the missig inlines to `hash_functions.rs` again to get back some speed.


jschneider-bensch (approved):
That's great, thanks!
On my machine, I'm seeing the AVX2 versions of keygen and verification at about the same speed (or slower in the case of keygen) as the portable version, but for signing it's roughly `portable = 2 * avx2` and `avx2 = 2 * pqclean`.

PR #705, opened 2024-12-03
fixing code to address review comments

karthikbhargavan:


karthikbhargavan:
I still want to add a file describing verification status and improve the state of polynomials.rs, but other comments should be addressed.

libcrux-ml-kem/src/mlkem512.rs
```diff
@@ -3,31 +3,29 @@ use super::{constants::*, ind_cca::*, types::*, *};
...
-// [hax]: hacspec/hacspec-v2#27 stealing error
-// block_len::<VECTOR_U_COMPRESSION_FACTOR_512>()
-const C1_BLOCK_SIZE_512: usize = 320;
-// [hax]: hacspec/hacspec-v2#27 stealing error
-// serialized_len::<RANK_512, C1_BLOCK_SIZE_512>()
-const C1_SIZE_512: usize = 640;
+const C1_BLOCK_SIZE_512: usize =
+    (COEFFICIENTS_IN_RING_ELEMENT * VECTOR_U_COMPRESSION_FACTOR_512) / 8;
+const C1_SIZE_512: usize = C1_BLOCK_SIZE_512 * RANK_512;
 const VECTOR_V_COMPRESSION_FACTOR_512: usize = 4;
```
franziskuskiefer, line 13: Because I just commented on the stealing issue. I think it is fixed and we could use these const functions now. I don't think we want to do this in here. But we could make use of it in ml-dsa (if we don't switch to a style where we don't need it in the first place).

franziskuskiefer (approved):
lgtm

libcrux-ml-kem/src/mlkem512.rs
karthikbhargavan: Good point, added an issue for cleanup: https://github.com/cryspen/libcrux/issues/708

Issue #706, opened 2024-12-03
[ML-DSA] C Extraction


franziskuskiefer:
Tracking issue for C extraction of the ML-DSA code.

* #707 
* #710

---

Eurydice issues we work around for now, or block us
- [ ] https://github.com/AeneasVerif/eurydice/issues/102
- [ ] https://github.com/AeneasVerif/eurydice/issues/100
- [ ] https://github.com/AeneasVerif/eurydice/issues/99
- [ ] https://github.com/AeneasVerif/eurydice/issues/104
- [ ] https://github.com/AeneasVerif/eurydice/issues/105
- [ ] https://github.com/AeneasVerif/eurydice/issues/106
- [ ] https://github.com/AeneasVerif/eurydice/issues/107
- [ ] https://github.com/AeneasVerif/eurydice/issues/115

hax issues we work around for now
- [ ] https://github.com/hacspec/hax/issues/1171

franziskuskiefer:
[Eurydice doesn't translate enums correctly](https://github.com/AeneasVerif/eurydice/issues/102) right now. We may have to change the error handling for that

franziskuskiefer:
A first version was merged in #710

PR #707, opened 2024-12-03
First changes for ML-DSA C extraction

franziskuskiefer:
First steps towards extracting C code for ML-DSA.

- ML-DSA key size feature gates
- C extraction infrastructure
- Moved Signature type
- Moved to named struct for types

cc #706


libcrux-ml-dsa/src/helper.rs
```diff
@@ -0,0 +1,66 @@
...
+    (for $i:ident in ($start:literal..$end:expr).step_by($step:literal) $body:block) => {
+        for $i in $start..$end / $step {
+            let $i = $i * $step;
+            $body
+        }
+    };
+}
+
+#[cfg(not(eurydice))]
+macro_rules! cloop {
```
keks, line 45: couldn't these be generalized to

```rust
macro_rules! cloop {
  (for $pattern:pat in $iterator:expr $body:block) => {for $pattern in $interator $block}
```

or something similar? I don't think we need to restate the macro for every case.

keks (changes requested):
I focussed on the Rust code because I don't know F-Star well enough and the shell scripts looked like they were copied from elsewhere.

I left a few comments. I am happy to discuss anything I wrote, in case you disagree.

libcrux-ml-dsa/src/helper.rs
keks: Got it. Also, I think it makes sense to have the same patterns as in the macro_rules for eurydice, so we know the macros expand in both worlds. Let's leave this as-is.

keks (approved):
Thanks, looks good!

Issue #717, opened 2024-12-16
Use most recent nightly for macOS ASAN on CI


franziskuskiefer:
We switched to a fixed nightly version (`2024-12-14`) for macOS ASAN for now because the most recent one appears to be broken.

jschneider-bensch:
Attempting to unpin this in #1258.

PR #718, opened 2024-12-16
disable macos asan

franziskuskiefer:
See #717


jschneider-bensch:
Why don't we use a specific nightly where we know it still worked until it works again on the most recent nightly?

jschneider-bensch (approved):
I've changed it to use `nightly-2024-12-14`, I believe that still worked.

PR #719, opened 2024-12-16
Introduce AVX2 `target_feature` boundary in x4 sampling

jschneider-bensch:
The PR takes @franziskuskiefer's work on introducing another target feature boundary in x4 sampling for ring elements.

I've introduced a new enum that selects which sampler to use and allows switching on the target feature for AVX2.

Also, I tried to write `SampleArgs` a bit more generically, in case we want to do the same for error sampling.


libcrux-ml-dsa/src/ml_dsa_generic.rs
```diff
@@ -26,6 +26,7 @@ pub(crate) mod multiplexing;
 
 /// Generate a key pair.
 #[inline(always)]
+#[allow(unsafe_code)]
```
franziskuskiefer, line 29: Are these unsafes here in generic still needed? It looks like you pushed that down into the sampling.

libcrux-ml-dsa/src/sample.rs
```diff
@@ -34,20 +34,66 @@ fn rejection_sample_less_than_field_modulus<SIMDUnit: Operations>(
     done
 }
 
-pub(crate) fn sample_four_ring_elements<SIMDUnit: Operations>(
+pub(super) struct SampleArgs<
```
franziskuskiefer, line 37: Let's add some comments so we remember what this is and how it works.

libcrux-ml-dsa/src/sample.rs
```diff
@@ -34,20 +34,66 @@ fn rejection_sample_less_than_field_modulus<SIMDUnit: Operations>(
...
+    const ROWS_IN_A: usize,
+    const COLUMNS_IN_A: usize,
+> {
+    pub(super) rand_stack: &'a mut (
+        [u8; STACK_SIZE],
+        [u8; STACK_SIZE],
+        [u8; STACK_SIZE],
+        [u8; STACK_SIZE],
+    ),
+    pub(super) tmp_stack: &'a mut [[i32; 263]],
```
franziskuskiefer, line 50: I'm experimenting with this one to avoid conversions. (Not for this PR, just FYI.)

franziskuskiefer (commented):
thanks, generally lgtm. I'll do another full pass later.
Can you update the C and F* extraction to make sure we don't do anything we shouldn't?

libcrux-ml-dsa/src/ml_dsa_generic.rs
jschneider-bensch: :+1: You're right, forgot to drop these exceptions.

libcrux-ml-dsa/src/samplex4.rs
```diff
@@ -1,35 +1,46 @@
...
-fn update_matrix<SIMDUnit: Operations, const ROWS_IN_A: usize, const COLUMNS_IN_A: usize>(
-    m: &mut Matrix<SIMDUnit, ROWS_IN_A, COLUMNS_IN_A>,
-    i: usize,
-    j: usize,
-    v: PolynomialRingElement<SIMDUnit>,
-) {
-    m[i][j] = v;
+/// A call to sample four ring elements from $seed into $memory at indices $a, $b
+/// $c, $d.
+macro_rules! sample_four_ring_elements_into {
```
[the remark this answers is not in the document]
jschneider-bensch: You're right, the macro is now a bit pointless. It made things more concise before, when we didn't pass in everything individually, but I didn't revisit it after making those changes.

PR #732, opened 2025-01-07
ML-DSA Parameter Set Monomorphization with Macros

jschneider-bensch:
The motivation for this PR is to mostly get rid of allocations below the level of `ml_dsa_generic`. To this end, `const` generics are replaced by a macro-based manual monomorphization to the different parameter sets. In doing this, I've attempted to simplify parameter set specific constants, as well as platform instantiation and multiplexing.

Essentially `ml_dsa_generic`, `constants`, the instantiations in `ml_dsa_generic::instantiation::{portable, neon, avx2}` and `ml_dsa_generic::multiplexing` now have (macro-generated) submodules for each parameter set which should be guarded by their respective feature flag. They contain:
- `constants::ml_dsa_44/65/87`: Parameter set specific constants go here. The `constants` module also defines `const fn`s for deriving everything but the basic constants in the parameter set specific submodules. 
- `ml_dsa_generic::ml_dsa_44/65/87`: A version of the `ml_dsa_generic::generic` module where the parameter specific constants from `constants` are pulled in and derived constants are derived using the `const fn`s from `constants`. The module also defines type aliases `MLDSA{44/65/87}SigningKey, MLDSA{44/65/87}VerificationKey, etc`. This is achieved by the `ml_dsa_parameter_sets` proc macro.
- `instantiations::{portable, neon, avx2}::ml_dsa_44/65/87: The platform specific instantiations of the versions from `ml_dsa_generic`. In the case of `avx2` that includes the application of the `target_platform`  cfg.
- `multiplexing`::ml_dsa_44/65/87`: Runtime multiplexing, reverting to a portable version if platform-specific version was not compiled in.

I've verified that F* extraction still works and laxes. C extraction pending further updates.


libcrux-ml-dsa/src/encoding/signing_key.rs
```diff
@@ -10,32 +10,23 @@ use crate::{
...
+    error_ring_element_size: usize,
+    seed_matrix: &[u8],
+    seed_signing: &[u8],
     verification_key: &[u8],
-    s1: [PolynomialRingElement<SIMDUnit>; COLUMNS_IN_A],
-    s2: [PolynomialRingElement<SIMDUnit>; ROWS_IN_A],
-    t0: [PolynomialRingElement<SIMDUnit>; ROWS_IN_A],
-) -> [u8; SIGNING_KEY_SIZE] {
-    let mut signing_key_serialized = [0u8; SIGNING_KEY_SIZE];
+    s1_2: &[PolynomialRingElement<SIMDUnit>],
```
karthikbhargavan, line 20: does this argument name come from the ML-DSA standard? if so, its ok, otherwise, a more informative name would be useful.

libcrux-ml-dsa/src/encoding/t1.rs
```diff
@@ -7,16 +7,15 @@ use crate::{
 
 #[inline(always)]
 pub(crate) fn serialize<SIMDUnit: Operations>(
-    re: PolynomialRingElement<SIMDUnit>,
+    re: &PolynomialRingElement<SIMDUnit>,
 ) -> [u8; RING_ELEMENT_OF_T1S_SIZE] {
```
karthikbhargavan, line 11: Is there a reason this is not an `&mut` argument?

libcrux-ml-dsa/src/matrix.rs
```diff
@@ -1,163 +1,121 @@
...
-
-            result[i] = invert_ntt_montgomery(result[i]);
+pub(crate) fn compute_matrix_x_mask<SIMDUnit: Operations>(
+    rows_in_a: usize,
+    columns_in_a: usize,
+    matrix: &[PolynomialRingElement<SIMDUnit>],
+    mask: &[PolynomialRingElement<SIMDUnit>],
+    result: &mut [PolynomialRingElement<SIMDUnit>],
+) {
+    for i in 0..rows_in_a {
```
karthikbhargavan, line 43: So much nicer without the enumerate!

libcrux-ml-dsa/src/ml_dsa_generic.rs
karthikbhargavan, line 1: I did not review the correctness of computations in this module and whether it matches the algorithm. Will give it a deeper look when linking it to the specification.

libcrux-ml-dsa/src/samplex4.rs
```diff
@@ -1,629 +1,147 @@
...
-        #[cfg(feature = "mldsa65")]
-        (6, 5) => matrix_A_6_by_5::<SIMDUnit, Shake128, ROWS_IN_A, COLUMNS_IN_A>(seed),
-        #[cfg(feature = "mldsa87")]
-        (8, 7) => matrix_A_8_by_7::<SIMDUnit, Shake128, ROWS_IN_A, COLUMNS_IN_A>(seed),
-        _ => unreachable!(),
-    }
 }
 
-#[cfg(feature = "mldsa44")]
+// Not inling this causes a 10x slow-down
```
karthikbhargavan, line 123: Would be great to open an issue so we understand why inlining is so crucial here, or if it is well understood, then some documentation on what is going on, so that we don't forget and make a mistake with inlining later.

libcrux-ml-dsa/src/simd/avx2.rs
```diff
@@ -7,163 +10,151 @@ mod ntt;
...
-    fn to_coefficient_array(&self) -> [i32; 8] {
-        vector_type::to_coefficient_array(&self)
+    fn to_coefficient_array(value: &Self::Coefficient, out: &mut [i32]) {
+        vector_type::to_coefficient_array(value, out)
     }
+
     #[inline(always)]
-    fn add(lhs: &Self, rhs: &Self) -> Self {
-        arithmetic::add(lhs.coefficients, rhs.coefficients).into()
+    fn add(lhs: &mut Self::Coefficient, rhs: &Self::Coefficient) {
```
karthikbhargavan, line 35: I guess the switch to using a borrow for the RHS and a mutable borrow for the LHS is primarily to have consistency with the portable version so we can share the Operations trait between them? Might be worth a comment.
Otherwise it looks a bit weird for AVX2 since Vec256 is a scalar.

libcrux-ml-dsa/src/simd/portable/vector_type.rs
```diff
@@ -4,23 +4,22 @@ use crate::simd::traits::COEFFICIENTS_IN_SIMD_UNIT;
 pub(crate) type FieldElement = i32;
 
 #[derive(Clone, Copy)]
-pub struct PortableSIMDUnit {
-    pub(crate) coefficients: [FieldElement; COEFFICIENTS_IN_SIMD_UNIT],
-}
+pub(crate) struct PortableSIMDUnit {}
```
karthikbhargavan, line 7: What is this used for?

karthikbhargavan (approved):
This is a big one! Left a few comments, but happy for most of these being follow-ups.

However, let's document the general style guidelines implemented here, so that new code can also follow this pattern. I am adding a comment to the PR noting this.

karthikbhargavan:
This PR does a bit more cleanup than just the const generic monomorphization. 
Let's try to document some of these style decisions so that we can be uniform in the codebase (and backport to ML-KEM and SHA-3 when we touch that code again.). Here are some things I noticed:
* All struct and array arguments to functions are now borrows 
* We uniformly use `&mut` arguments instead of returning a new array or struct
* We uniformly use slices (with implicit lengths) instead of fixed-size arrays
* We eliminate `const` generic arguments and replace them by real arguments (when truly needed)
* In several places, we replace calls to complex iterators like `enumerate` with simpler for loops over ranges.

jschneider-bensch:
You're right, there's more here than the monomorphization. Thanks for noting down the style changes!

libcrux-intrinsics/src/avx2.rs
```diff
@@ -352,7 +357,16 @@ pub fn mm256_testz_si256(lhs: Vec256, rhs: Vec256) -> i32 {
 
 #[inline(always)]
 pub fn mm256_xor_si256(lhs: Vec256, rhs: Vec256) -> Vec256 {
-    unsafe { _mm256_xor_si256(lhs, rhs) }
+    // This floating point xor may or may not be faster than regular xor.
+    // Local testing seems to indicate that it's a little more stable in
+    // benchmarks though.
+    // See https://stackoverflow.com/questions/27804476/difference-between-mm256-xor-si256-and-mm256-xor-ps
```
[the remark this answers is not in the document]
jschneider-bensch: Resolving, since it makes trouble for the Windows CI and we took it out, for now: https://github.com/cryspen/libcrux/pull/732/commits/66afce2b7d2b86febb97fb1fc5de2fbba7419d74

PR #734, opened 2025-01-08
ML-DSA: Work around Eurydice limitations

franziskuskiefer:
This PR works around limitations and drops the associated type in the operations trait.
See https://github.com/AeneasVerif/eurydice/issues/125.

Eurydice can't handle associated types right now.
So this changes it such that the Operations trait is implemented
on top of a wrapper struct for the coefficients again.
This requires the implementations of the trait to point into the struct
again. But this is more self contained than propagating the generic type
for the trait.

I couldn't measure an impact on performance or stack size with this change

Note that C extraction is using slightly newer versions of the tools here. I'm not sure if that makes a difference.


libcrux-ml-dsa/src/polynomial.rs
```diff
@@ -15,8 +15,7 @@ impl<SIMDUnit: Operations> PolynomialRingElement<SIMDUnit> {
         }
     }
 
-    // This is useful for debugging.
-    // XXX: Used in `make_int`
+    // This is used in `make_int` and for tests
```
jschneider-bensch, line 18:
```suggestion
    // This is used in `make_hint` and for tests
```

libcrux-ml-dsa/src/simd/avx2/arithmetic.rs
```diff
@@ -3,9 +3,9 @@ use crate::{
     simd::traits::{FIELD_MODULUS, INVERSE_OF_MODULUS_MOD_MONTGOMERY_R},
 };
 
-use libcrux_intrinsics::avx2::*;
+use libcrux_intrinsics::avx2::{mm256_setzero_si256, *};
```
jschneider-bensch, line 6:
```suggestion
use libcrux_intrinsics::avx2::*;
```
This is redundant, no?

libcrux-ml-dsa/src/simd/portable/encoding/error.rs
```diff
@@ -40,6 +40,8 @@ fn serialize_when_eta_is_4(simd_unit: &Coefficients, serialized: &mut [u8]) {
 
 #[inline(always)]
 pub(crate) fn serialize(eta: Eta, simd_unit: &Coefficients, serialized: &mut [u8]) {
+    // [eurydice] injects an unused variable here in the C code for some reason.
+    //            That's why we don't match here.
```
jschneider-bensch, line 44: Is the comment correct? We do a match at this point.

jschneider-bensch (approved):
Looks good, thanks :+1: 
Just some really minor questions/nits below.

libcrux-ml-dsa/src/simd/portable/encoding/error.rs
franziskuskiefer: Ah, no. I first wanted to change all of these. But it's not too bad and I think something that could get fixed.

franziskuskiefer:
@jschneider-bensch if you're happy with this, please merge it into #732.

PR #736, opened 2025-01-09
ML-DSA: update benchmarks pqclean

franziskuskiefer:
This updates the pqclean benchmarks to use ml-dsa rather than dilithium.

### DIFF
--- a/.github/workflows/mldsa.yml
+++ b/.github/workflows/mldsa.yml
@@ -101,8 +101,6 @@ jobs:
 
       # Benchmarks
       - name: 🔨 Build Benchmarks
-        # Comparison benchmarks don't work on intel mac. Re-enable if we think it's worthwhile.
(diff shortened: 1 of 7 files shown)


jschneider-bensch (changes requested):
Thanks, that makes sense!
The `criterion` ones are not updated, though. Maybe this is a point where we can decide if we want to keep them? For me, they are still very inconsistent from run to run (large run-to-run diffs in signing, mainly) and for this reason I don't look to them at all, I find. So I would be fine retiring them. But you might get some use out of them still?

franziskuskiefer:
Good point. I totally forgot about those. I updated them just in case we want to use them again...

jschneider-bensch (approved):

franziskuskiefer:
pq-clean mldsa apparently doesn't build on intel macos 🤷🏻  I'll disable it there.

PR #737, opened 2025-01-13
`no_std` support (via `alloc`)

jschneider-bensch:
This is a quick step towards `no_std` support for libcrux, which is necessary e.g. for https://github.com/cryspen/bertie/issues/126

We do still assume presence of a global allocator for the `no_std` target in question, and I've made a note of that in the Readme.

### DIFF
--- a/libcrux-kem/src/kem.rs
+++ b/libcrux-kem/src/kem.rs
@@ -850,26 +850,6 @@ impl Ct {
                     ct_x.try_into().map_err(|_| Error::InvalidCiphertext)?,
                 ))
             }
-            #[cfg(feature = "kyber")]
-            Algorithm::X25519Kyber768Draft00 => {
-                let key: [u8; MlKem768Ciphertext::len() + 32] =
-                    bytes.try_into().map_err(|_| Error::InvalidCiphertext)?;
-                let (xct, kct) = key.split_at(32);
-                Ok(Self::X25519Kyber768Draft00(
-                    kct.try_into().map_err(|_| Error::InvalidCiphertext)?,
-                    xct.try_into().map_err(|_| Error::InvalidCiphertext)?,
-                ))
-            }
-            #[cfg(feature = "kyber")]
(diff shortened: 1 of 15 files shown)


libcrux-kem/src/kem.rs
```diff
@@ -850,26 +850,6 @@ impl Ct {
                     ct_x.try_into().map_err(|_| Error::InvalidCiphertext)?,
                 ))
             }
-            #[cfg(feature = "kyber")]
```
franziskuskiefer, line 853: I'm not sure I understand why this is gone.

src/aead.rs
```diff
@@ -17,6 +17,12 @@ use crate::hacl::chacha20_poly1305;
 
 use libcrux_platform::{aes_ni_support, simd128_support, simd256_support};
 
+#[cfg(feature = "std")]
```
franziskuskiefer, line 20: The global way of handling this is usually to add something like
```rust
#[cfg(feature = "std")]
extern crate std;
```
and mapping std and alloc to the same name with a `use`.
Also, when possible, we should add an `alloc` feature at some point, to allow for full `no_std`.

But this here is fine.

franziskuskiefer (changes requested):
lgtm except for that dropped code that I don't quite understand.

libcrux-kem/src/kem.rs
jschneider-bensch: These two match arms are fully redundant with the two directly above. (We get a warning as well with the `kyber` feature.)
Maybe this was a different slip up than just duplication, and the lower ones were supposed to do something different for `cfg(not(feature = "kyber"))`. In any case, I can take this change out of the PR.

franziskuskiefer (approved):
Thanks

Do you want to put the removal of those redundant branch in a separate PR? Then we can figure out what's going on there.

jschneider-bensch:
Yes, I put it on https://github.com/cryspen/libcrux/pull/739 :+1:

PR #738, opened 2025-01-13
Update ML-DSA Signature deserialize

franziskuskiefer:
Since we can use returns now, we can change these while loops into nicer for loops. This also removes a particularly ugly piece of C code that eurydice generates.
This also addresses a comment from @karthikbhargavan in #732 and should make this function easier to prove.

Related charon issue (C control flow): https://github.com/AeneasVerif/charon/issues/507


jschneider-bensch (approved):
Nice!
Can we break from loops, in general, then? That would be really nice for the rejection sampling loops.

franziskuskiefer:
> Can we break from loops, in general, then?

hax generally supports it now, yes. I think a bunch of F* libraries are still missing for full support (https://github.com/hacspec/hax/issues/1204). But the case here seems to be ok.

PR #739, opened 2025-01-14
Remove redundant match arms

jschneider-bensch:


franziskuskiefer (commented):
Ok, looking at the entire function again I'm not sure why this is here and we should remove it. I see mlkem variants further up, so a `not(kyber)` config doesn't seem what we want either.

franziskuskiefer (approved):

PR #750, opened 2025-01-15
Remove double return workarounds.

maximebuyse:
https://github.com/cryspen/hax/issues/720 is now fixed in hax main (by https://github.com/cryspen/hax/pull/1223) so we can remove the corresponding workarounds.


maximebuyse:
> @maximebuyse I suspect that the F* output changed with this? Did you re-generate F* (`./hax.py extract`)? Otherwise I'm happy with this when the CI is happy with it. Very happy to get rid of those workaround 🙂

Right, I re-generated for ml-dsa and ml-kem now.

franziskuskiefer (approved):
Awesome, thanks!

PR #751, opened 2025-01-15
kem: fix decoding of hybrid private keys

franziskuskiefer:


jschneider-bensch (approved):
Thank you!
I will file a follow up issue to increase test coverage.

PR #755, opened 2025-01-16
Add pure-Rust Blake2 

keks:
This PR adds the pure-rust implementation from hacl-rs. So far this only features a low-level byte-oriented API.

One thing I am not sure about is the `last_node` flag. From reading the spec I think it maybe should be set to `true`, but that makes it produce wrong/inconsistent results.

It also reduces the visibility of some hacl-related submodules of ed25519, that is an unrelated cleanup change.


franziskuskiefer (commented):
Just a quick drive-by. I'll look at it tomorrow.

blake2/src/impl_hacl.rs
```diff
@@ -0,0 +1,376 @@
...
+        Self { salt, ..self }
+    }
+
+    /// Constructs the [`Blake2s`] hasher.
+    pub fn build(self) -> Blake2s<KEY_LEN, OUT_LEN> {
+        // these are safe because they bot are at most 32
+        let key_length = KEY_LEN as u8;
+        let digest_length = OUT_LEN as u8;
+
+        // NOTE: I am not entirely sure that this is the correct value. From reading the spec I
```
franziskuskiefer, line 216: `false` is correct here. Aymeric wrote some notes when adding tests for `true` to hacl. https://github.com/hacl-star/hacl-star/pull/934. I don't think we need to expose the `true` version for now.

franziskuskiefer (changes requested):

blake2/src/impl_hacl/blake2s.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// Creates the builder for an unkeyed hasher.
+    pub fn new_unkeyed() -> Result<Self, Error> {
+        Ok(Self {
+            key: &(),
+            personal: &[0; PARAM_LEN],
+            salt: &[0; PARAM_LEN],
+        })
+    }
+
+    /// Constructs the [`Blake2sK`] hasher.
```
franziskuskiefer, line 29: Hm, I don't see a type `Blake2sK`.

blake2/src/impl_hacl/blake2b.rs
```diff
@@ -0,0 +1,145 @@
...
+            key: &[],
+            personal: &[0; PARAM_LEN],
+            salt: &[0; PARAM_LEN],
+        })
+    }
+}
+impl<'a, const KEY_LEN: usize, const OUT_LEN: usize> Blake2bBuilder<'a, KEY_LEN, OUT_LEN> {
+    /// Creates the builder and returns an error if the lengths `KEY_LEN` or `OUT_LEN` are not in
+    /// the allowed range.
+    pub fn new() -> Result<Self, Error> {
```
franziskuskiefer, line 39: I'm not sure if I understand how this can be used without a length that's known at compile time.
How would I implement the following function with this implementation?
```rust
fn blake2b(out: &mut [u8]) {
    // Call libcrux blake2b to write out.len() bytes
}
```

blake2/src/impl_hacl/blake2s.rs
keks: Ah yes, that was a remnant from earlier in the refactor. Fixed that and improved comments in general.

blake2/src/impl_hacl/blake2b.rs
keks: Oh, I just realized I didn't communicate very clearly. So far I only did Blake2s and I wanted to check with you before applying the same pattern to Blake2b.

blake2/src/impl_hacl/blake2s.rs
```diff
@@ -0,0 +1,464 @@
...
+    pub fn new_unkeyed() -> Result<Self, Error> {
+        Ok(Self {
+            key: &(),
+            personal: &[0; PARAM_LEN],
+            salt: &[0; PARAM_LEN],
+        })
+    }
+
+    /// Constructs the [`Blake2s`] hasher for unkeyed hashes and dynamic digest length.
+    pub fn build(self, digest_length: u8) -> Result<Blake2s<Blake2sStaticKey<0>>, Error> {
```
franziskuskiefer, line 30: Do I understand correctly that the `Blake2sStaticKey` is only there to hold the key length? Why can't we use a const generic directly for that? Also, we don't really use it.

franziskuskiefer (commented):
I think I agree that the code feels a little overcomplicated but the usage is fine. I'd be fine with doing this.

The alternative could be to use two separate structs. Then you can drop all the generic types. Looking at the code you don't seem to be sharing a lot of code between the different versions such that you wouldn't get much more code with two separte structs.

I don't have a strong opinion on which version is better.

blake2/src/impl_hacl/mod.rs
```diff
@@ -0,0 +1,249 @@
...
+
+/// Type that holds the constant in case just the digest length is known at compile time.
+pub struct ConstsStaticOut<const OUT_LEN: usize>;
+
+/// Type that holds the constant in case neither the key length nor the digest length is known at
+/// compile time.
+pub struct ConstsDynamic;
+
+#[cfg(test)]
+mod test {
```
franziskuskiefer, line 26: These tests should be for the crate, not only the hacl implementation. Let's move them up to the lib? Er even out as integration tests, we don't seem to use internals here.

blake2/src/impl_hacl/blake2s.rs
```diff
@@ -0,0 +1,471 @@
...
+            state: malloc_raw(kk, key),
+            _phantom: PhantomData,
+        })
+    }
+}
+
+impl<'a, const KEY_LEN: usize> Blake2sBuilder<'a, &'a [u8; KEY_LEN]> {
+    /// Creates the builder for an keyed hasher for keys where the length is known at compile
+    /// time.
+    pub fn new_keyed_static(key: &'a [u8; KEY_LEN]) -> Result<Self, Error> {
```
franziskuskiefer, line 119: Do we have `_static`, `_const`, and `_var` now? Either I don't understand what they are used for, or we could unify this a little.

franziskuskiefer (approved):
Generally lgtm, let's address the remaining comments and get this in.

PR #757, opened 2025-01-19
[ML-KEM] incremental API

franziskuskiefer:
An incremental API for ML-KEM that allows encapsulating to a partial public key.

Only the multiplexing API is exposed for incremental.


libcrux-ml-kem/src/ind_cca/incremental.rs
```diff
@@ -0,0 +1,1003 @@
...
+    impl<const K: usize, Vector: Operations + 'static> Keys for MlKemKeyPairUnpacked<K, Vector> {
+        fn as_any(&self) -> &dyn Any {
+            self
+        }
+    }
+
+    /// The partial ciphertext c1 - first part.
+    #[repr(transparent)]
+    pub struct Ciphertext1<const LEN: usize> {
+        pub value: [u8; LEN],
```
gram-signal, line 120: it would be nice if PublicKey2 were like this, rather than just being a `&[u8]`, so it had a known, type-checked length.

gram-signal (commented):
Seems easily understandable and good to go.  A few top-level comments:

1) why `Box<dyn ...>` rather than just returning a struct of the right size and with the right stuff inside it as arrays (for `State` and `Keys`)
2) our use case requires storing/retrieving things we use, particularly `State` and `Keys`, so we'll need some way to serialize and deserialize them.  If they're just byte arrays or structs containing sets of byte arrays (that are public), we can do the {de,}seriallization ourselves.
3) As mentioned elsewhere, it would be nice to get a (pubkey, privkey) rather than an opaque `Keys`, so we could throw away the public key and just keep the privkey around for decapsulation, to potentially save some memory space.
4) There's a bunch of structs in the public API that aren't themselves public.  I believe I've added comments for all of them.

Thanks again for this, it's super nice to see!

franziskuskiefer (commented):
> 1. why `Box<dyn ...>` rather than just returning a struct of the right size and with the right stuff inside it as arrays (for `State` and `Keys`)

Serialising the key and state costs extra cycles, or is very unsafe. That's why we opted for a safe, efficient version as the default API. But when we add serialisation for everything anyway, we can also add an additional API that serialises the values before it returns them.

> 2. our use case requires storing/retrieving things we use, particularly `State` and `Keys`, so we'll need some way to serialize and deserialize them.  If they're just byte arrays or structs containing sets of byte arrays (that are public), we can do the {de,}seriallization ourselves.

Everything, except PK1 and the ciphertexts are platform dependent unfortunately. That's why we need the extra hoops. But we can add serialisation for all of them. The cheapest versions are just pretty unsafe. I'll think about a good solution and propose something.

> 3. As mentioned elsewhere, it would be nice to get a (pubkey, privkey) rather than an opaque `Keys`, so we could throw away the public key and just keep the privkey around for decapsulation, to potentially save some memory space.

As commented inline, you need parts of the public key. So you can't just throw it away. We can add a separate struct that drops unneeded parts of the key. I'm not sure if it's worth it though. Let's look at this when we're happy with the rest.

> 4. There's a bunch of structs in the public API that aren't themselves public.  I believe I've added comments for all of them.

I commented on them. They are not needed, but we can of course expose them.

libcrux-ml-kem/src/mlkem768.rs
```diff
@@ -653,6 +653,98 @@ pub(crate) mod kyber {
...
+///
+/// **NOTE:** This is a non-standard API. Use with caution!
+pub mod incremental {
+    pub use self::incremental::types::PublicKey1;
+    use self::incremental::types::{Ciphertext1, Ciphertext2, Error, State};
+
+    use super::*;
+    extern crate alloc;
+    use alloc::boxed::Box;
+    use ind_cca::incremental::{self, types::Keys};
```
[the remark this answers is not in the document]
gram-signal: I agree it's not needed when you do:

```
  let foo = x();
  foo.y();
```

However, I need to put this in a struct:

```
struct Foo {
  keys: incremental::Keys,
}

fn a() -> Self {
  Self{ keys: x() }
}
fn b(&mut self) {
  self.keys.y();
}
```

For this purpose, it appears that public stuff is actually necessary?

libcrux-ml-kem/src/mlkem768.rs
```diff
@@ -653,6 +653,98 @@ pub(crate) mod kyber {
         >(private_key, ciphertext)
     }
 }
+
+/// Incremental API.
+///
+/// **NOTE:** This is a non-standard API. Use with caution!
+pub mod incremental {
+    pub use self::incremental::types::PublicKey1;
+    use self::incremental::types::{Ciphertext1, Ciphertext2, Error, State};
```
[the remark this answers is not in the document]
gram-signal: One other thing:  because Ciphertext1/2 is parameterized, C1_SIZE_768 also appears to be necessary.  And it's somewhat annoying to use... maybe a type alias?  Would something like this work?:

```
pub type Ciphertext1 = types::Ciphertext1<C1_SIZE_768>;
```

libcrux-ml-kem/src/ind_cca/incremental.rs
gram-signal: I'm having trouble understanding this... how can PK2 be platform dependent?  I'm going to pass its bytes between different platforms.  Or is it that the serialization is constant-sized, but the struct isn't?  In that case, can we make a publicly facing PublicKey2 struct that's the right size and use it for returning bytes to the user?

libcrux-ml-kem/src/mlkem768.rs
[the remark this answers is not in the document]
gram-signal: Actually, I'm taking this (partially) back.  We shouldn't need to have the structs themselves be public if we can serialize/deserialize them.  So, if we just get {de,}serialization, we can deserialize when needed and store the serialized stuff within our structs at other times.  This will probably be both more space and more computation efficient for our use case.

libcrux-ml-kem/src/ind_cca/incremental.rs
```diff
@@ -157,6 +157,58 @@ pub mod types {
...
+        pub const fn num_bytes() -> usize {
+            SHARED_SECRET_SIZE
+                + vec_len_bytes::<K, Vector>()
+                + PolynomialRingElement::<Vector>::num_bytes()
+                + 32
+        }
+
+        /// Get the state as bytes
+        pub fn to_bytes(self, out: &mut [u8]) {
+            debug_assert!(out.len() >= Self::num_bytes());
```
gram-signal, line 171: if there's any chance that this serialization will change in the future, it might be prudent to add a "version byte" to the head of this serialization.  These serializations may be stored persistently for long periods of time on user devices, and may be processed by different versions of the libcrux library.  Should any changes occur here (including length changes, or changes in the layout of bytes or additional information), new library versions should be able to process old library serialization, and old library versions should at least be able to fail gracefully to process new serializations (return a `Result`?).  Some initial byte saying "this is serialization version 1" could aid with that.

libcrux-ml-kem/src/ind_cca/incremental.rs
```diff
@@ -157,6 +157,58 @@ pub mod types {
...
+
+            self.error2.to_bytes(&mut out[offset..]);
+            offset += PolynomialRingElement::<Vector>::num_bytes();
+
+            out[offset..offset + 32].copy_from_slice(&self.randomness);
+        }
+
+        /// Build a state from bytes
+        pub fn from_bytes(bytes: &[u8]) -> Self {
+            debug_assert!(bytes.len() >= Self::num_bytes());
```
gram-signal, line 187: This particular assert is probably best as a Result, if there's any possibility of future serialization change.

libcrux-ml-kem/src/mlkem512.rs
```diff
@@ -677,6 +677,34 @@ pub mod incremental {
         RANK_512 * 16 * 32
     }
 
+    /// The size of the key pair in bytes.
+    pub const fn key_pair_len() -> usize {
```
gram-signal, line 681: I'm 99% of the way there on using this, but there's still one thing I think should be made public here: the expected lengths of CT1 and CT2.  I need to set up a receiver with the expected number of bytes I expect to receive:

```
  let receiver = ExpectToReceive(CT1_SIZE);
  while !receiver.done() {
    receiver.read_byte(b);
  }
```

Right now, I'm not seeing a good way to do this with the constnats and functions that are currently publicly exposed.  Of course, I could recompute the constants myself, but that seems a bit brittle.

libcrux-ml-kem/tests/self.rs
```diff
@@ -150,17 +152,30 @@ macro_rules! impl_consistency_incremental {
...
+                assert_eq!(ct1.value, ct12.value);
+
                 // ... and then to pk2.
                 // pk2 is passed in as bytes because the deserializaiton is runtime
                 // platform dependent.
                 let ct2 = encapsulate2(state.as_ref(), &pk2_bytes).unwrap();
 
+                // encaps2 with serialized state
+                let ct22 = encapsulate2_serialized(&serialized_state, &pk2_bytes).unwrap();
+                assert_eq!(ct2.value, ct22.value);
```
gram-signal, line 177: how does one extract the shared secret from serialized state?

libcrux-ml-kem/src/ind_cca/incremental.rs
franziskuskiefer: That's why it needs to be serialized. Rust doesn't allow us to just cast this, that's what all the serialize functions are doing. But the size is always fixed, but depending on the platform we have bytes, or SIMD registers in the key.

libcrux-ml-kem/src/ind_cca/incremental.rs
franziskuskiefer: That's a very good question. This is really just the flat byte representation of the state and don't see a reason this should ever change.
We could add a version byte. But I don't want to make things more complicated than necessary.
I think if something here changes, we should just create a new state struct.

libcrux-ml-kem/src/ind_cca/incremental.rs
franziskuskiefer: Yeah, the functions here still panic all over the place. I'll start cleaning things up now.

libcrux-ml-kem/tests/self.rs
franziskuskiefer: That's further down, the `state.shared_secret()`.

karthikbhargavan:
Proofs are failing currently (changes to constants.rs) so need some fixing.
I will do those and then add my review.

karthikbhargavan (approved):
I went through the code and fixed up the proofs, so this should hopefully pass CI.

PR #766, opened 2025-01-23
Make RSA API more usable

keks:
This PR makes changes to the API of the `libcrux-rsa` crate:

- Get rid of the signature type and use byte arrays instead
- Add an API that uses slices instead of array references and does the relevant length checks. The idea here is that this makes it easier to use if the lengths of the keys and signatures are not known at compile time.
- Add a test using that API


jschneider-bensch (commented):
I think it looks good!
Could you add some tests that trigger the new errors, as well?

rsa/tests/rsa_pss.rs
```diff
@@ -55,6 +58,47 @@ fn self_test_rsa_pss() {
...
+    // test the variable length signing fails if the length is wrong
+    let mut signature = [0u8; 257];
+    let err = sign(
+        DigestAlgorithm::Sha2_256,
+        &sk.as_var_len(),
+        &msg,
+        &salt,
+        &mut signature[..],
+    )
+    .expect_err("expected singing with wrong length to fail");
```
jschneider-bensch, line 90:
```suggestion
    .expect_err("expected signing with wrong length to fail");
```

jschneider-bensch (approved):
Nice! Thanks for adding more docs, as well. :)

franziskuskiefer:
@keks can the conversations be resolved?

Issue #733, opened 2025-01-07
[ML-KEM] Run proofs on CI


franziskuskiefer:
The CI currently only lax checks on CI. It should also run the full verification.

franziskuskiefer:
@maximebuyse what's the state here?

Issue #585, opened 2024-09-16
Add ML-DSA verification to CI


franziskuskiefer:
Attraction is on CI. But a mechanism is needed to easily add modules for full type checking and do that on CI.

- [x] lax checking
- [ ] type checking

franziskuskiefer:
Has been fixed in #957

PR #771, opened 2025-01-23
Update hax CI

franziskuskiefer:
Towards #585 and #733 

This PR uses the hax action and splits the action into multiple steps
- extraction: `hax.py extract`
- diff
- lax: `hax.py prove --admit`

Based on this we can add verification with `hax.py prove` for ML-KEM and ML-DSA.


karthikbhargavan (approved):
Looks good.

PR #774, opened 2025-01-29
Adapt to new hax naming.

maximebuyse:
This removes some hardcoded names in raw fstar strings. We use the $ feature of hax-lib to make this more robust. With this changes, the extracted code lax-checks with both hax main and https://github.com/cryspen/hax/pull/1199.

In some places, we need to use a workaround documented in https://github.com/cryspen/hax/issues/1276#issuecomment-2621982382


libcrux-ml-kem/Cargo.toml
```diff
@@ -26,7 +26,7 @@ rand = { version = "0.8", optional = true }
 libcrux-platform = { version = "0.0.2-beta.2", path = "../sys/platform" }
 libcrux-sha3 = { version = "0.0.2-beta.2", path = "../libcrux-sha3" }
 libcrux-intrinsics = { version = "0.0.2-beta.2", path = "../libcrux-intrinsics" }
-hax-lib = { version = "0.1.0", git = "https://github.com/hacspec/hax/" }
+hax-lib = { version = "0.1.0", path = "../../hax/hax-lib" }
```
franziskuskiefer, line 29: Leftover? This should be reverted.

libcrux-ml-kem/proofs/fstar/extraction/ML.KEM.fst.config.json
franziskuskiefer, line 1: hm, this should probably stay?

franziskuskiefer (commented):
lgtm with two nits

franziskuskiefer (approved):
Please never force push on a branch in a PR. That makes reviewing impossible.

PR #775, opened 2025-01-29
Add option to skip hax diffs.

maximebuyse:
This makes it easier to test if hax PRs don't break lax-checking.


franziskuskiefer (approved):
Thanks

PR #776, opened 2025-01-29
Functional correctness for portable encoding commitment

### LINKED ISSUE #745: Verify ML-DSA: portable/encoding/commitment
Verify https://github.com/cryspen/libcrux/blob/main/libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs

- [ ] Write a spec for the functions
- [ ] Design tactic to prove correctness
- [ ] Prove panic freedom
- [ ] Prove correctness

W95Psp:
This PR adds a specification and a proof for the serialize function of `simd/portable/encoding/commitment` in ML-DSA.

To do so, I had to take split the function in many smaller functions.

Fixes #745.


libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+#[hax_lib::ensures(|out| {
+    let serialized_future = future(serialized);
+    fstar!(r"
+Seq.length ${serialized_future} == Seq.length ${serialized} /\
+(let inp = bit_vec_of_int_t_array #I32 #(mk_usize 8) ${simd_unit.values} 4 in
+ let out = bit_vec_of_int_t_array #U8  #(mk_usize 4) ${serialized_future} 8 in
+ forall (i: nat {i < 8 * 4}). inp i == out i)
+")
+})]
+pub fn serialize_4(simd_unit: &Coefficients, serialized: &mut [u8]) {
```
franziskuskiefer, line 52:
```suggestion
fn serialize_4(simd_unit: &Coefficients, serialized: &mut [u8]) {
```

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+    let serialized_future = future(serialized);
+    fstar!(r"
+let d = Seq.length ${serialized} in
+( Seq.length ${serialized_future} == d /\
+  (let inp = bit_vec_of_int_t_array #I32 #(mk_usize 8) ${simd_unit.values} d in
+   let out = bit_vec_of_int_t_array #U8  #(mk_usize d) ${serialized_future} 8 in
+   forall (i: nat {i < 8 * d}). inp i == out i))
+")
+})]
+pub fn serialize(simd_unit: &Coefficients, serialized: &mut [u8]) {
```
franziskuskiefer, line 152:
```suggestion
pub(in crate::simd::portable) fn serialize(simd_unit: &Coefficients, serialized: &mut [u8]) {
```

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+    // 4 bits.
+    cloop! {
+        for (i, coefficients) in simd_unit.values.chunks_exact(2).enumerate() {
+            hax_lib::loop_invariant!(|i: usize| {fstar!(r"
+            Seq.length $serialized == 4 /\ (
+              let inp = bit_vec_of_int_t_array #I32 #(mk_usize 8) ${simd_unit.values} 4 in
+              let out = bit_vec_of_int_t_array #U8 #(mk_usize 4) $serialized 8 in
+              forall (n: nat {n < v i * 8}). out n == inp n
+            )")});
+            let r0 = encode_4(coefficients);
```
franziskuskiefer, line 63: Do we need `r0` as a temporary here? If not, assigning it directly would look nicer.

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+    // The commitment has coefficients in [0,15] => each coefficient occupies
+    // 4 bits.
+    cloop! {
+        for (i, coefficients) in simd_unit.values.chunks_exact(2).enumerate() {
+            hax_lib::loop_invariant!(|i: usize| {fstar!(r"
+            Seq.length $serialized == 4 /\ (
+              let inp = bit_vec_of_int_t_array #I32 #(mk_usize 8) ${simd_unit.values} 4 in
+              let out = bit_vec_of_int_t_array #U8 #(mk_usize 4) $serialized 8 in
+              forall (n: nat {n < v i * 8}). out n == inp n
+            )")});
```
franziskuskiefer, line 62:
```suggestion
            )")});

```

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+            let r0 = encode_4(coefficients);
+            serialized[i] = r0;
```
franziskuskiefer, line 64:
```suggestion
            serialized[i] = r0;

```

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+    for (i, coefficients) in simd_unit.values.chunks_exact(4).enumerate() {
+        hax_lib::loop_invariant!(|i: usize| {
+            fstar!(
+                r"
+            Seq.length $serialized == 6 /\ (
+              let inp = bit_vec_of_int_t_array #I32 #(mk_usize 8) ${simd_unit.values} 6 in
+              let out = bit_vec_of_int_t_array #U8 #(mk_usize 6) $serialized 8 in
+              (forall (n: nat {n < v i * 24}). out n == inp n))"
+            )
+        });
```
franziskuskiefer, line 108:
```suggestion
        });

```

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+        let (r0, r1, r2) = encode_6(coefficients);
+        serialized[3 * i] = r0;
+        serialized[3 * i + 1] = r1;
+        serialized[3 * i + 2] = r2;
```
franziskuskiefer, line 112:
```suggestion
        serialized[3 * i + 2] = r2;

```

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,158 @@
...
+        hax_lib::loop_invariant!(|i: usize| {
+            fstar!(
+                r"
+            Seq.length $serialized == 6 /\ (
+              let inp = bit_vec_of_int_t_array #I32 #(mk_usize 8) ${simd_unit.values} 6 in
+              let out = bit_vec_of_int_t_array #U8 #(mk_usize 6) $serialized 8 in
+              (forall (n: nat {n < v i * 24}). out n == inp n))"
+            )
+        });
+        let (r0, r1, r2) = encode_6(coefficients);
```
franziskuskiefer, line 109: Why doesn't `encode_6` (and `encode_4`) take a mutable reference to `serialized` to write right into it?

franziskuskiefer (commented):
Only a few nits and questions

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
W95Psp: Yes, indeed, that was useful at some point for writing the proof, but it's no longer useful, will drop it!

franziskuskiefer:
I filed https://github.com/AeneasVerif/eurydice/issues/138 for the C failure. Unfortunately, I don't see a workaround for this right now.

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
W95Psp: Stateful updates make both proofs and proof statements more complicated: we need to state that nothing changes but the updated value(s), which implies a universally quantifier
So here I prefer to cut the proof in two: the serialization problem (ie prove correctness for shifts & friends) and then the loopy-imperative-update things.

franziskuskiefer: Ok, if there's a good reason to do this, it's fine. But please add an explanation into comments why it is done this way, as it's not what we'd usually write in Rust.

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,161 @@
...
-                    let coefficient1 = coefficients[1] as u8;
-
-                    serialized[i] = (coefficient1 << 4) | coefficient0;
-                }
-            }
-            ()
-        }
+#[hax_lib::fstar::before(
+    r#"
+#set-options "--fuel 0 --ifuel 1 --z3rlimit 300 --z3version 4.13.3 --ext context_pruning"
```
karthikbhargavan, line 6: Do we really depend on a specific version of z3? I would prefer to leave such version setting to the Makefile

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
karthikbhargavan: I think we should be able to do a version where the `encode` functions mutate the output array instead of returning a tuple. I am putting this improvement for the followup issue: https://github.com/cryspen/libcrux/issues/777

karthikbhargavan (approved):

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
W95Psp: Would be happy to use a recent Z3 by default (I've set that locally not to mess with anything else)

karthikbhargavan: Let's just remove the version here then. System default is fine, and any version pinning should happend outside the Rust files.

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -1,40 +1,161 @@
...
+Seq.length ${serialized_future} == Seq.length ${serialized} /\
+(let inp = bit_vec_of_int_t_array #I32 #(mk_usize 8) ${simd_unit.values} 6 in
+ let out = bit_vec_of_int_t_array #U8  #(mk_usize 6) ${serialized_future} 8 in
+ forall (i: nat {i < 8 * 6}). inp i == out i)
+")
+})]
+pub fn serialize_6(simd_unit: &Coefficients, serialized: &mut [u8]) {
+    // The commitment has coefficients in [0,43] => each coefficient occupies
+    // 6 bits.
+    for (i, coefficients) in simd_unit.values.chunks_exact(4).enumerate() {
```
franziskuskiefer, line 100: This needs a `cloop` for eurydice extraction.

karthikbhargavan: done

karthikbhargavan:
@W95Psp I adapted the code and changed the proof accordingly.
Please review and then: address the z3 version issue on CI (likely pinning to 4.13.3 would work)

I have no idea what the build failures with address sanitizer are doing on macos-latest.

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -127,9 +135,17 @@ pub fn serialize_6(simd_unit: &Coefficients, serialized: &mut [u8]) {
...
                 == {} out (v i * 24 + n);
-                })
+                });
+           (* Find a better style for these sequence updates so that such silly assertions become unnecessary *)
+           assert (forall (n:nat{n >= 24 * v i /\ n < 24 * v i + 24}). inp (24 * v i + (n - 24 * v i)) == out (24 * v i + (n - 24 * v i)));
+           assert (forall (n:nat{n >= 24 * v i /\ n < 24 * v i + 24}). inp n == out n);
+           assert (forall (n:nat{n < v i * 24}). n / 8 < 3 * v i);
+           assert (forall (j:nat{j < 3 * v i}). Seq.index ${serialized} j == Seq.index (Seq.slice ${serialized} 0 (3 * v i)) j);
+           assert (forall (j:nat{j < 3 * v i}). Seq.index ${_old_serialized} j == Seq.index (Seq.slice ${_old_serialized} 0 (3 * v i)) j);
+           assert (forall (n:nat{n < 24 * (v i + 1)}). inp n == out n))
```
W95Psp, line 145: We need both the calc proof and the assertions? I guess that's because of the `serialized[*..*]`?

W95Psp: I wonder if we should just not have a lemma that generalize "updating contiguous blocks of an array in a loop".
I will take a look at the other portable serialize. That pattern isprobably everywhere, so it might be worth writing such a lemma

W95Psp:
Seems OK to me, though we should do something about the proofs: already my calc thing is too big for such a small function, now with the asserts that's a bit overwhelming.
But that works.

For Z3, I updated the job to use latest F*, which ships with Z3 4.13.3.

W95Psp:
I experimented a bit with unrolling loops @karthikbhargavan, in branch lf-portable-encoding-commitment-experiment-unroll.

Sadly, this doesn't work out of the box, because of subslices.
I tried to define locally `[].` and update slice to have quantifiers everywhere, but that doesn't help.
Shall I investigate more?

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
karthikbhargavan: Let's add any improvements as notes to https://github.com/cryspen/libcrux/issues/777

karthikbhargavan:
This PR still has an ASAN failure on Rust nightly on macos-latest. 
However, I am going to merge it and do any necessary fixes in a future PR.

Issue #777, opened 2025-01-30
Meta: Improve Specs and Reduce Annotations for ML-DSA Proofs


karthikbhargavan:
While the first goal for ML-DSA verification is to get robust proofs in place, we would like to investigate and implement improvements in the style and follow them as we go forward. This issue collects such improvements, which can then be broken into separate sub-issues.

- [ ] Use `&mut` inputs rather than returning tuples in inner encoding functions
    - evaluate verification performance and document if doing this change would hurt perf
- [ ] Write serialization annotations (pre-/post-conditions, invariants) in Rust (not in F*)
    - Reflect bit-vector functions in Rust to make writing annotations in Rust possible
- [ ] Unroll loops in hax or in F*
    - Remove calc proofs to get more automation
- [ ] Investigate the use of SMT for these proofs
    - Are we already getting enough automation from F*->SMT?
- [ ] #783
    * Initial experiment here: https://hax-playground.cryspen.com/#fstar/3d2cbc0a9a/gist=1644de52204954e0a260187a5909403c
- [ ] Mutable arrays cannot change length, but we still need preconditions saying that the length did not change. Fix.

franziskuskiefer:
@karthikbhargavan is this still something you plan on doing?

karthikbhargavan:
yes. i will update with status and plans

PR #780, opened 2025-01-30
Add proofs for the basic functions of MLDSA arithmetic

mamonet:
This PR adds proofs for `add`, `subtract`, and `get_n_least_significant_bits` function for arithmetic module of ML-DSA


karthikbhargavan (approved):
Looks good. Let's merge these functions now but let us also switch to Rust annotations in the next PR on this file, whenever possible:
See: https://github.com/cryspen/libcrux/issues/777

libcrux-ml-dsa/src/simd/portable/arithmetic.rs
```diff
@@ -9,22 +9,79 @@ use crate::{
...
 
 #[inline(always)]
+#[hax_lib::fstar::options("--z3rlimit 150")]
+#[hax_lib::requires(fstar!(r#"forall i. i < 8 ==> 
+    Spec.Utils.is_intb (pow2 31 - 1) (v (Seq.index ${lhs}.f_values i) + v (Seq.index ${rhs}.f_values i))"#))]
+#[hax_lib::ensures(|result| fstar!(r#"forall i. i < 8 ==>
+    (v (Seq.index ${lhs}_future.f_values i) == 
+     v (Seq.index ${lhs}.f_values i) + v (Seq.index ${rhs}.f_values i))"#))]
 pub fn add(lhs: &mut Coefficients, rhs: &Coefficients) {
+    let _lhs0 = lhs.clone();
```
franziskuskiefer, line 19: This may or may not be ok. But it will certainly increase the debug stack usage. For release builds these will get eliminated. But not in debug.

mamonet:
I pushed an update that provides proofs for `montgomery_reduce_element()` and `montgomery_multiply_fe_by_fer()`

libcrux-ml-dsa/src/simd/portable/arithmetic.rs
mamonet: The cloning is because the argument is passed by reference and not by value, but I assume it has the same cost. The loop invariant needs the original value to keep track on the not-yet-modified elements of the array.

mamonet: The solution would be to return a new array that has the modified values but AFAIK Rust allocates these arrays on stack too.

franziskuskiefer: Yep, we can't return values here. These have to be mutable arguments. We can probably tolerate a few of these clones. But probably not enough to do it in every function. So it's not an issue for this PR, but this style of proof. @karthikbhargavan lets look at this.

mamonet:
I added proofs for `montgomery_multiply_by_constant()` and `montgomery_multiply()`, I think this PR is in a good shape now. I will push a proofs for the rest arithmetic functions in a different PR.

libcrux-ml-dsa/src/simd/portable/arithmetic.rs
karthikbhargavan: This is now noted in https://github.com/cryspen/libcrux/issues/777
And I added a small style example for ghost state there, which we could implement within hax (it has to be handled in the engine in a specific way.)

Issue #417, opened 2024-07-18
Build C extraction docker image on CI


jschneider-bensch:
Can you file an issue to have the CI build the docker image at some point? Then we don't need to do that manually every time we update something here.

_Originally posted by @franziskuskiefer in https://github.com/cryspen/libcrux/pull/416#discussion_r1682455336_

- [x] publish docker image to GitHub packages
- [x] rebuild image when a file in the directory changes
- [x] don't run `c.yml` when something in `docker/c` changes
- [x] use GH hosted image in all workflows
- [x] publish new image when PR that updates it gets merged
- [x] update C extraction bash scripts to write revisions when extracting in the docker image
- [ ] clean docker image to reduce size, e.g. clean Rust builds (install and clean), drop unused Rust toolchains, etc. #818
- [ ] add status job to docker build job, like in `c.yml` #819

---

- Workflow that
  - builds the docker image
  - publishes the docker image to ghcr.io with an `unstable` tag (maybe with the revision as tag?)
  - uses the new docker image to extract C code (from c.yml)
  - run tests on the newly extracted C code (from c.yml)
  - fail if any of these step fails
- Second workflow to tag an existing docker image on ghrc.io as latest

### Steps from c.yml
There are more places where things are extracted and tested. But here are two parts for ml-kem as example.
https://github.com/cryspen/libcrux/blob/20dc15a510c56ba44b3dcbc3054cc3f446f05c2b/.github/workflows/c.yml#L16
https://github.com/cryspen/libcrux/blob/20dc15a510c56ba44b3dcbc3054cc3f446f05c2b/.github/workflows/c.yml#L144

franziskuskiefer:
We finally got [a new F* release](https://github.com/FStarLang/FStar/releases/tag/v2023.09.03) and should update to use new versions of everything. As part of that we should also fix this issue.
I started [an action](https://github.com/cryspen/libcrux/tree/franziskus/docker-build-tests) that could be a starting point, but it [fails with permission issues](https://github.com/cryspen/libcrux/actions/runs/11823573758/job/32943170174) (ERROR: unauthorized: access token has insufficient scopes) for now.

franziskuskiefer:
> we may want to allow using other tags for testing other revisions.

_posted by @franziskuskiefer in https://github.com/cryspen/libcrux/pull/791#discussion_r1944269047_

franziskuskiefer:
The main parts are done here. Closing this. The two open items will be done in the corresponding follow up issues.

PR #782, opened 2025-01-31
Refresh C code + fix c.yaml following latest eurydice/krml changes

protz:
The latest changes in krml/eurydice generate more precise header inclusion, meaning the mlkem avx2 public header no longer includes the sha3 avx2 public header (this is not necessary: only the mlkem avx2 C file needs that header), meaning that the intrinsics need to be directly included from the mlkem avx2 public header.


msprotz:
This PR also showcases improved control-flow reconstruction (CC @Nadrieril and @sonmarcho).

franziskuskiefer:
This will require updating the toolchain. We probably won't get to that his week. Until #417 is fixed, that's a manual process.

franziskuskiefer:
This has been addressed through a series of other PRs with #809 being the last one.

Issue #746, opened 2025-01-14
Verify ML-DSA: set up SIMD scaffolding


karthikbhargavan:
- [x] Verify constants.rs
- [x] Verify portable/vector_type.rs
- [x] Verify avx2/vector_type.rs
- [x] Verify simd/traits.rs

karthikbhargavan:
This is currently blocked by https://github.com/cryspen/hax/issues/1266
Working around to make progress.

karthikbhargavan:
This PR is currently blocked by some issues with the translations of attributes.
In particular, in the treatment of bool vs Prop.
https://github.com/cryspen/hax-evit/issues/11

franziskuskiefer:
@karthikbhargavan what's the state here? https://github.com/cryspen/hax-evit/issues/12 got merged.

karthikbhargavan:
Addressed in https://github.com/cryspen/libcrux/pull/788

PR #788, opened 2025-02-03
SIMD Traits Spec and Pre-Condition Style

karthikbhargavan:
This PR addresses https://github.com/cryspen/libcrux/issues/746
It needs to be merged after #776  and #780 
We should also discuss and agree on the style here

As a side-effect, this PR also improves the specs and proofs for ML-DSA in the following ways:
* the specs for simd/portable/arithmetic.rs now use the `add_pre, add_post, sub_pre, sub_post` functions from simd/traits.rs. This reduces clutter and will help verification performance
* the proofs for simd/portable/encoding/commitment.rs are now fully automated, relying on a hand-unrolling of the loops


franziskuskiefer:
I think here too, a transparent attribute could make things nicer. Let's look at this together.

karthikbhargavan:
This PR is currently blocked by some issues with the translations of attributes.
In particular, in the treatment of bool vs Prop.
https://github.com/cryspen/hax-evit/issues/11

karthikbhargavan:
With the work in https://github.com/cryspen/hax/pull/1301, I've now updated this PR so it passes F*.
Let's work towards merging that hax PR next and then get back to this one (should be quick)

karthikbhargavan:
@W95Psp @mamonet you may want to look at the styles here:

* We should aim for full automation using SMT in the style of `simd/portable/encoding/commitment.rs` whenever possible
* We should aim to re-use predicates like `add_pre` (see `simd/portable/arithmetic.rs`) as much as possible. (We can them make these predicates opaque if necessary, without the need to make the functions `add`, `sub` opaque.)

@W95Psp: we still need a little more work on https://github.com/cryspen/hax/pull/1301 for this.
In particular, the `forall` predicates within the `add_pre` etc get compiled into valid F* but do not easily lend themselves to use in proofs. It would be better to have a custom translation of everything in Hax_lib.Prop so that these specs can be super-compact in F*.

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
```diff
@@ -54,37 +12,24 @@ Seq.length ${serialized_future} == Seq.length ${serialized} /\
...
-                    == {} out (v i * 8 + n);
-                  })
-            ");
-        }
-    }
-    ()
+    serialized[0] = byte0;
+    serialized[1] = byte1;
+    serialized[2] = byte2;
+    serialized[3] = byte3;
```
mamonet, line 32: Interesting. I also think there will be run-time performance benefits when unrolling complicated loops, but the question is how convenient is unrolling such loops.

libcrux-ml-dsa/src/simd/portable/arithmetic.rs
```diff
@@ -6,60 +6,53 @@ use crate::{
...
 
 #[inline(always)]
 #[hax_lib::fstar::options("--z3rlimit 150")]
-#[hax_lib::requires(fstar!(r#"forall i. i < 8 ==> 
-    Spec.Utils.is_intb (pow2 31 - 1) (v (Seq.index ${lhs}.f_values i) + v (Seq.index ${rhs}.f_values i))"#))]
-#[hax_lib::ensures(|result| fstar!(r#"forall i. i < 8 ==>
-    (v (Seq.index ${lhs}_future.f_values i) == 
-     v (Seq.index ${lhs}.f_values i) + v (Seq.index ${rhs}.f_values i))"#))]
+#[hax_lib::requires(add_pre(&lhs.values, &rhs.values))]
+#[hax_lib::ensures(|result| add_post(&lhs.values, &rhs.values, &(future(lhs).values)))]
```
mamonet, line 17: Sounds a good idea. I will try to apply that where I find a condition match with functions of `Operations` trait.

libcrux-ml-dsa/src/simd/portable/encoding/commitment.rs
karthikbhargavan: I would like such unrollings to be done automatically in hax or elsewhere. Essentially, the effect of this unrolling is to make the proof trivial for SMT, and whichever way we can get this effect is fine, but would need some experimentation in coming PRs. Whatever works, we can backport to ML-KEM.

libcrux-ml-dsa/src/simd/portable/arithmetic.rs
karthikbhargavan: Yes, more generally, whenever we see a formula that is being repeated in multiple places, we can now define it once and use it multiple times, using the new Prop features provided by https://github.com/cryspen/hax/pull/1301

karthikbhargavan:
Seeing this failure (outside CI): https://github.com/cryspen/libcrux/actions/runs/13447960542

libcrux-ml-dsa/src/simd/traits.rs
```diff
@@ -16,6 +19,140 @@ pub const INVERSE_OF_MODULUS_MOD_MONTGOMERY_R: u64 = 58_728_449;
...
+
+#[cfg(not(eurydice))]
+#[hax_lib::attributes]
+pub(crate) trait Repr: Copy + Clone {
+    #[requires(true)]
+    fn repr(&self) -> SIMDContent;
+}
+
+#[cfg(hax)]
+pub(crate) fn int_is_i32(i: hax_lib::Int) -> bool {
```
franziskuskiefer, line 32: Why not put all the proof pieces into a common module here instead of polluting the main code? Then you also need only one `cfg` guard.

karthikbhargavan: yes, good idea.

karthikbhargavan: I did the change, but had to tinker a little bit to avoid creating a recursive bundle. In a future PR, we may consider moving Repr also into the specs module.

franziskuskiefer (approved):
Let's get this in.

PR #791, opened 2025-02-05
Build C extraction Docker image and run tests on CI

wysiwys:
This pull request creates the first GitHub Actions workflow for the C extraction Docker container, outlined in #417:

- Build the Docker image, and publish it to GitHub packages with `unstable` tag
- Use the new Docker image to extract C code, and run tests on newly extracted C code

This PR does not contain the second workflow from #417, which will publish the image under the `latest` tag, and will be run after all the tests in this first workflow pass.

### DIFF
--- a/.github/workflows/docker-c.yml
+++ b/.github/workflows/docker-c.yml
@@ -1,0 +1,212 @@
+name: Rebuild C Docker image
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: ['main']
+    paths: ['.docker/c/**']
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}-c
+  VERSION_TAG: unstable
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+
+    # We need write access to packages
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Docker image
+        id: push
+        uses: docker/build-push-action@v6
+        with:
+          context: .docker/c/
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.VERSION_TAG }}
+    outputs:
+      image: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.VERSION_TAG }}
+  extract:
+    needs: [build-and-push-image]
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ needs.build-and-push-image.outputs.image }}
+    defaults:
+      run:
+        working-directory: libcrux-ml-kem
+        shell: bash
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Extract C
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-c.yml
```diff
@@ -0,0 +1,211 @@
...
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@9780b0c442fbb1117ed29e0efdff1e18412f7567
```
jschneider-bensch, line 29: I think these revisions on the actions are still from Franziskus' initial version of the action. I think this here is the latest release of `docker/login-action` at this time and below is an older release of `docker/build-push-action`. @franziskuskiefer: Any reason not to use `@v3` / `@v6`?

.github/workflows/docker-c.yml
```diff
@@ -0,0 +1,211 @@
...
+        uses: docker/build-push-action@4f58ea79222b3b9dc2c8bbdd6debcef730109a75
+        with:
+          context: .docker/c/
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.VERSION_TAG }}
+  extract:
+    needs: [build-and-push-image]
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/cryspen/libcrux-c:unstable
```
jschneider-bensch, line 46: You could use the environment variables in these jobs, too, in case the exact image name changes.

.github/workflows/docker-c.yml
```diff
@@ -0,0 +1,211 @@
...
+
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+
+    # We need write access to packages
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
```
jschneider-bensch, line 22: I think we don't need this, since the job doesn't do artifact attestation at the moment. I'm guessing the initial version was based on the one from [here](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-ghcrio) which creates an attestation in the last step. We could just also include that step, but I don't think it's necessary.

jschneider-bensch (approved):
Thanks, this looks good! I have some comments, but nothing that needs to be changed, I think.

.github/workflows/docker-c.yml
wysiwys: I updated the workflow to use the latest version of `login-action` and `build-push-action`.

.github/workflows/docker-c.yml
```diff
@@ -0,0 +1,211 @@
+name: Rebuild C Docker image
+
+on:
+  workflow_dispatch:
+  push:
```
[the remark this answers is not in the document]
wysiwys: I ended up implementing this as below:
```
on:
  workflow_dispatch:
  pull_request:
    branches: ['main']
    paths: ['.docker/c/**']
```

This should run on `pull_request`s that are opened, synchronized, or reopened.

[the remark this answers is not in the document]
franziskuskiefer: 👍🏻  thanks

franziskuskiefer:
Thanks, let's merge this.

.github/workflows/docker-c.yml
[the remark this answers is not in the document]
wysiwys: In `Actions` I can see that the workflow run was indeed triggered when this PR, which changes `.docker/c/Dockerfile`, was synchronized. So it is working :+1:

PR #792, opened 2025-02-06
Fix randomness consumption in XWing draft 2 implementation

kkohbrok:
Both in version 2 and in version 6 (the current version at the time of writing), encap of XWing should take 64 bytes of randomness, where it uses the first 32 bytes for ML-KEM and the second 32 bytes for X25519. Right now, it uses the first 32 bytes for both ML-KEM _and_ X25519. This PR fixes that.


jschneider-bensch (approved):
You're absolutely right, thanks for fixing this!

PR #794, opened 2025-02-06
Update `rand` and `rand_core`

wysiwys:
This pull request updates the dependencies `rand` from 0.8 -> 0.9 and `rand_core` from 0.6 -> 0.9. It is based on dependabot's PRs (see #787 and #786).

- Use new trait `TryRngCore` where necessary
- Use `OsRng::try_fill_bytes()` instead of `OsRng::fill_bytes()`, which was removed. Also, to handle the case where `try_fill_bytes()` returns an error, add an `Error::InsufficientOSRandomness` variant

### DIFF
--- a/libcrux-psq/Cargo.toml
+++ b/libcrux-psq/Cargo.toml
@@ -21,7 +21,7 @@ classic-mceliece-rust = { version = "2.0.0", features = [
     "mceliece460896f",
     "zeroize",
 ] }
-rand = { version = "0.8" }
+rand = { version = "0.9" }
 libcrux-ecdh = { version = "0.0.2-beta.2", path = "../libcrux-ecdh" }
 libcrux = { version = "0.0.2-beta.2", path = ".." }
 
--- a/libcrux-sha3/Cargo.toml
+++ b/libcrux-sha3/Cargo.toml
@@ -33,7 +33,7 @@ harness = false
 [dev-dependencies]
 criterion = "0.5.1"
 hex = "0.4.3"
-rand = "0.8.5"
+rand = "0.9.0"
 cavp = { version = "0.0.2-beta.2", path = "../cavp" }
 pretty_env_logger = "0.5.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ libcrux-ed25519 = { version = "=0.0.2-beta.2", path = "ed25519" }
 libcrux-ecdh = { version = "=0.0.2-beta.2", path = "libcrux-ecdh" }
 libcrux-ml-kem = { version = "=0.0.2-beta.2", path = "libcrux-ml-kem" }
 libcrux-kem = { version = "=0.0.2-beta.2", path = "libcrux-kem" }
-rand = { version = "0.8" }
+rand = { version = "0.9" }
 log = { version = "0.4", optional = true }
 # WASM API
 wasm-bindgen = { version = "0.2.87", optional = true }
@@ -93,7 +93,7 @@ hax-lib = { version = "0.1.0", git = "https://github.com/hacspec/hax/" }
 [dev-dependencies]
 libcrux = { path = ".", features = ["rand", "tests"] }
 pretty_env_logger = "0.5"
-rand = { version = "0.8" }
+rand = { version = "0.9" }
 rand_core = { version = "0.6" }
 quickcheck = "1"
 quickcheck_macros = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ hax-lib = { version = "0.1.0", git = "https://github.com/hacspec/hax/" }
 libcrux = { path = ".", features = ["rand", "tests"] }
 pretty_env_logger = "0.5"
 rand = { version = "0.9" }
-rand_core = { version = "0.6" }
+rand_core = { version = "0.9" }
 quickcheck = "1"
 quickcheck_macros = "1"
 serde_json = { version = "1.0" }
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -19,7 +19,7 @@ rand = { version = "0.9" }
 libcrux = { path = "../", features = ["rand", "tests"] }
 libcrux-kem = { path = "../libcrux-kem", features = ["tests"] }
 libcrux-ml-kem = { path = "../libcrux-ml-kem" }
-rand_core = { version = "0.6" }
+rand_core = { version = "0.9" }
 # Benchmarking "RustCrypto"
 chacha20poly1305 = "0.10"
 sha2 = "0.10"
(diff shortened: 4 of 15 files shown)


franziskuskiefer:
@wysiwys what's the state here? Do you need input or want someone else to take over?

wysiwys:
@franziskuskiefer One thing I ran into that is holding back development on this branch is connected to a third-party crate, which is built on top of `rand` version `0.8`. Specifically, there is a trait definition issue that arises when calling  `classic_mceliece_rust::keypair_boxed()` in `psq.rs`. 

To go into more detail, the inputs to that function must implement the `rand::CryptoRng` and `rand_core::RngCore` traits, but the versions of those traits from older versions of those libraries. However, in `rand` version `0.9` and the new, corresponding `rand_core` version, the definitions of these traits have been changed, which leads to a build error in this crate, since the input previously passed into it is no longer valid.

Once `classic-mceliece-rust` has released a new version, and depends on the latest version of `rand`, the development on the dependency update in this branch should go smoothly. In the meantime, is there a different solution or workaround that could be applied for this issue?

franziskuskiefer:
We can't keep everything back because there's a bad dependency in a small part.
Either you update everything but psq to 0.9, or everything including psq but add a second API entrypoint for mceliece that gets the old rng.

wysiwys:
> We can't keep everything back because there's a bad dependency in a small part. Either you update everything but psq to 0.9, or everything including psq but add a second API entrypoint for mceliece that gets the old rng.

Thanks @franziskuskiefer . I went with the option of adding a second entrypoint for mceliece, since psq depends on `libcrux_kem` (and then the same `rand` dependency issue would apply to it as well).

wysiwys:
The only other changes to the crates' public APIs are replacing inputs that are `&mut (impl RngCore + CryptoRng)` with `&mut impl CryptoRng`, since `CryptoRng` is now a supertrait of `RngCore`.

wysiwys:
Also thanks @jschneider-bensch for providing this workaround for the  `rand=0.8` dependency requirement for `classic-mceliece-rust`. I incorporated the commit from your branch into this pull request.

keks (commented):
I think this looks good in general! I am a bit unsure around the generated code: 

1. The toolchain hashes are gone in the headers and codegen.txt, I don't think this is supposed to happen.
2. This adds a header.txt, which seems like it maybe shouldn't be there?
3. The rlimit in the F* code is different, is that correct?

libcrux-intrinsics/proofs/fstar/extraction/Libcrux_intrinsics.Arm64_extract.fst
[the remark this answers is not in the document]
wysiwys: I agree - actually, these were updated when I ran `./hax.py extract` in `libcrux-ml-kem`. But they don't seem relevant

libcrux-ml-kem/c/code_gen.txt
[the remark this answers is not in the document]
wysiwys: Indeed - I regenerated this with the correct toolchain, and it turns out that there were no changes beyond the revisions.

libcrux-intrinsics/proofs/fstar/extraction/Libcrux_intrinsics.Arm64_extract.fst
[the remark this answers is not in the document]
wysiwys: Thanks for the review! Yes, it appears that the rlimit for the F* code was different. I am not familiar with the reason for this change, however.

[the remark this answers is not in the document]
wysiwys: Actually, these files are for dependencies of mlkem which were re-extracted when running `./hax.py extract`, so I have added them back in.

libcrux-psq/src/classic_mceliece.rs
```diff
@@ -35,27 +35,65 @@ impl Encode for Ciphertext {
...
+    /// The KEM's shared secret.
     type SharedSecret = SharedSecret<'static>;
+    /// The KEM's encapsulation key.
     type EncapsulationKey = PublicKey<'static>;
+    /// The KEM's decapsulation key.
     type DecapsulationKey = SecretKey<'static>;
 
+    /// Generate a pair of encapsulation and decapsulation keys.
     fn generate_key_pair(
         rng: &mut (impl rand::CryptoRng + rand::Rng),
```
franziskuskiefer, line 79: The trait should be only using `CryptoRng` here and in `encapsulate` now.

libcrux-psq/Cargo.toml
```diff
@@ -22,7 +22,8 @@ classic-mceliece-rust = { version = "2.0.0", features = [
     "mceliece460896f",
     "zeroize",
 ], optional = true }
-rand = { version = "0.8" }
+rand = { version = "0.9" }
+rand_old = { version = "0.8", package = "rand" }
```
franziskuskiefer, line 26: This is only needed when the `classic-mceliece` feature is enabled.

libcrux-sha3/Cargo.toml
```diff
@@ -33,7 +33,7 @@ harness = false
 [dev-dependencies]
 criterion = "0.5.1"
 hex = "0.4.3"
-rand = "0.8.5"
+rand = "0.9.0"
```
franziskuskiefer, line 36: 0.9.1 is the latest. Let's use that everywhere. Or just 0.9.

benchmarks/benches/x25519.rs
```diff
@@ -976,14 +995,15 @@ fn nym_sphinx_process(c: &mut Criterion) {
...
 
     group.bench_function("Dalek", |b| {
-        use rand_core::OsRng;
+        use rand_core_old::{CryptoRng, OsRng};
         use x25519_dalek::{EphemeralSecret, PublicKey};
 
         b.iter_batched(
             || {
-                let sk1 = EphemeralSecret::random_from_rng(OsRng);
+                let mut rng = OsRng;
```
franziskuskiefer, line 1003: Unused `mut`

benchmarks/benches/x25519.rs
```diff
@@ -976,14 +995,15 @@ fn nym_sphinx_process(c: &mut Criterion) {
     });
 
     group.bench_function("Dalek", |b| {
-        use rand_core::OsRng;
+        use rand_core_old::{CryptoRng, OsRng};
```
franziskuskiefer, line 998: Unused import `CryptoRng`

franziskuskiefer (changes requested):
Generally lgtm with a couple nits and there's a merge conflict.

libcrux-sha3/Cargo.toml
wysiwys: I went with "0.9" to match the rest of the codebase

libcrux-psq/src/classic_mceliece.rs
wysiwys: Should the `KEM` trait itself (in `libcrux-traits`) also be updated to only require `CryptoRng`?

franziskuskiefer: Yes, since that's enough now, let's update it.

PR #795, opened 2025-02-07
Update C extraction tools

franziskuskiefer:
This tests the new action to build the docker image for C extraction on CI by updating all tools to main.

### DIFF
--- a/.docker/c/ext-tools.sh
+++ b/.docker/c/ext-tools.sh
@@ -4,23 +4,23 @@ set -v -e -x
 
 source $HOME/.profile
 
-curl -L https://github.com/AeneasVerif/charon/archive/db4e045d4597d06d854ce7a2c10e8dcfda6ecd25.zip \
+curl -L https://github.com/AeneasVerif/charon/archive/30cab88265206f4fa849736e704983e39a404d96.zip \
     --output charon.zip
 unzip charon.zip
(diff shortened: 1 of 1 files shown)


.docker/c/Dockerfile
```diff
@@ -31,6 +31,9 @@ ENV HACL_HOME=$HOME/hacl-star
 ENV KRML_HOME=$HOME/karamel
 ENV EURYDICE_HOME=$HOME/eurydice
 ENV CHARON_HOME=$HOME/charon
+ENV KRML_REV=97a06e07e7e423df192c40d5a88bf6c85fd4d278
```
franziskuskiefer, line 34: All the envs here should be done more intelligently, i.e. re-use what's in `ext-tools.sh`, or put the revisions elsewhere.

wysiwys: Another possibility is to move these definitions higher up in the Dockerfile, before `ext-tools.sh` is run. That way, they only need to be defined once (in the Dockerfile).

wysiwys: Alternatively, these variables could be passed in to the `build-and-test` Action as `build-arg`s, but that would involve updating the CI workflows and might be unneeded complexity for now.

jschneider-bensch (approved):
Looks good to me!

PR #803, opened 2025-02-10
don't run `c.yml` when something in `.docker/c/` is changed

wysiwys:
Only run the `c.yml` workflow if the `.docker/c/` directory is unchanged. See #417


franziskuskiefer:
The hax job fails because of a hax change. https://github.com/cryspen/libcrux/pull/804 updates the F* code.

franziskuskiefer (approved):
lgtm, the false case we should see on #795.

PR #809, opened 2025-02-11
Update C extraction tools to current main

franziskuskiefer:


keks (approved):
I am not very confident in my Docker-fu, but I think this looks good.

Maybe the landing page of the `libcrux-c` container on ghcr could use some comment on what it is for, though; namely that it provides the tools needed for the C extraction (Karamel, Eurydice, Charon).

franziskuskiefer:
> Maybe the landing page of the `libcrux-c` container on ghcr could use some comment on what it is for

yeah, we should add something there. It's just taking the readme by default. But that's an unrelated to the update here. I put it on an issue.

PR #810, opened 2025-02-11
Update AEAD Benchmarks to use libcrux-chacha20poly1305 crate

### LINKED ISSUE #762: Chacha20Poly1305 benchmarks
Add benchmarks for the chacha20poly1305 crate to the benchmarks crate.

keks:
Fixes #762


chacha20poly1305/src/lib.rs
```diff
@@ -26,7 +29,35 @@ pub enum AeadError {
     InvalidCiphertext,
 }
 
+impl alloc::fmt::Display for AeadError {
```
jschneider-bensch, line 32:
```suggestion
impl core::fmt::Display for AeadError {
```
I think that also works? Then you don't need to pull in `alloc` at all.

keks: huh, TIL!

wysiwys (approved):
Looks good to me!

benchmarks/benches/aead.rs
jschneider-bensch, line 1: You could rename this to `chacha20poly1305.rs`, since that's the only AEAD under test here.

jschneider-bensch (approved):
Yep, looks good to me, as well! With a tiny, totally optional nit. :)

PR #815, opened 2025-02-12
SIMD: add a model

W95Psp:
This PR inits a new crate, `minicore`.
For now, this covers only a small subset of `core::arch::x86_64` and `core::arch::x86`.

This PR adds:
 - abstractions for bit and bitvecs
 - some specifications written in Rust (mainly a backport of what can be found in F* in fstar-helpers)
 - tests


minicore/src/arch/x86.rs
```diff
@@ -0,0 +1,285 @@
+use crate::abstractions::{bit::*, bitvec::*};
+
+pub(crate) mod upstream {
+    #[cfg(target_arch = "x86")]
+    pub use core::arch::x86::*;
+    #[cfg(target_arch = "x86_64")]
+    pub use core::arch::x86_64::*;
+}
+
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
```
franziskuskiefer, line 10: Shouldn't this be guarding the entire file?

minicore/src/arch/x86.rs
```diff
@@ -0,0 +1,285 @@
...
+    #[test]
+    fn mm256_sllv_epi32() {
+        for _ in 0..100 {
+            let vector: BitVec<256> = BitVec::rand();
+            let counts: BitVec<256> = BitVec::rand();
+            assert_eq!(super::_mm256_sllv_epi32(vector, counts), unsafe {
+                upstream::_mm256_sllv_epi32(vector.into(), counts.into()).into()
+            });
+        }
+    }
```
franziskuskiefer, line 252:
```suggestion
    }

```

franziskuskiefer (changes requested):
Thanks. This looks like it's quite ready though.

Let's move this into a separate directory for proofs (where we can move the other proof related parts too). Having this on the top level clutters things even more. We'll move crates as well soon.

Also, the high level description needs to explain what this crate is doing and how.

minicore/src/abstractions/bit.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn from(b: bool) -> Bit {
+        match b {
+            false => Bit::Zero,
+            true => Bit::One,
+        }
+    }
+}
+
+/// A trait for types that represent machine integers.
+pub trait MachineInteger {
```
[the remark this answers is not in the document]
W95Psp: I agree, but I don't know where to put this.
According to me we should just use what exists already (e.g. num-traits).

minicore/src/abstractions/bitvec.rs
```diff
@@ -0,0 +1,99 @@
+//! This module provides a specification-friendly bit vector type.
+use super::bit::{Bit, MachineInteger};
+
+// TODO: this module uses `u128/i128` as mathematic integers. We should use `hax_lib::int` or bigint.
+
+use std::fmt::Formatter;
+
+/// A straight-forward bitvector type.
```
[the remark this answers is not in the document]
W95Psp: I guess you mean that this description is not very clear and that I should write something more clear here?
Will do.

minicore/src/abstractions/bitvec.rs
```diff
@@ -0,0 +1,99 @@
...
+        if is_negative {
+            -s
+        } else {
+            s
+        }
+    } else {
+        u64_int_from_bit_slice(bits) as i128
+    };
+    let Ok(n) = result.try_into() else {
+        panic!("cannot try_into {result}")
```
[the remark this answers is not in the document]
W95Psp: I should put `unreachable!` here indeed: result should be guaranteed to be in bounds for `T` since `bits.len() >= T::BITS`.

minicore/src/abstractions/bitvec.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Constructor for BitVec. `BitVec::<N>::from_fn` constructs a bitvector out of a function that takes usizes smaller than `N` and produces bits.
+    pub fn from_fn<F: FnMut(usize) -> Bit>(f: F) -> Self {
+        Self(core::array::from_fn(f))
+    }
+    /// Convert a slice of machine integers where only the `d` least significant bits are relevant.
+    pub fn from_slice<T: Into<i128> + MachineInteger + Copy>(x: &[T], d: usize) -> Self {
+        Self::from_fn(|i| Bit::of_int(x[i / d], (i % d) as u32))
+    }
+    /// Construct a BitVec out of a machine integer.
+    pub fn from_int<T: Into<i128> + MachineInteger + Copy>(n: T) -> Self {
```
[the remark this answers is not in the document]
W95Psp: The intent here was to have a model quickly to conduct proofs, hence the TODO.

minicore/src/arch/x86.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Conversions impls between `BitVec<N>` and `__mNi` types.
+mod conversions {
+    use super::upstream::{
+        __m128i, __m256i, _mm256_loadu_si256, _mm256_storeu_si256, _mm_loadu_si128,
+        _mm_storeu_si128,
+    };
+    use super::BitVec;
+    impl From<BitVec<256>> for __m256i {
+        fn from(bv: BitVec<256>) -> __m256i {
+            let bv: &[u8] = &bv.to_vec()[..];
```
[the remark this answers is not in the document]
W95Psp: I don't see how we could implement a `AsRef<[u8]>` or a `as_slice: &'a BitVec -> &'a [u8]` on `BitVec`, bit vec doesn't contain a slice of u8 internally.

W95Psp:
Thanks for the review.
Shall I move that crate to `fstar-helpers` then?

minicore/src/arch/x86.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+}
+
+#[allow(non_camel_case_types)]
+type __m256i = BitVec<256>;
+#[allow(non_camel_case_types)]
+type __m128i = BitVec<128>;
+
+pub fn _mm_storeu_si128(output: *mut __m128i, a: __m128i) {
```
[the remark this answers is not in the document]
franziskuskiefer: The rust docs are pretty limited and sometimes wrong I think. But the the gist from the intel docs with a link would be good.

franziskuskiefer:
> Shall I move that crate to `fstar-helpers` then?

Yeah, I guess that should be a more general `verification` folder or something like that at some point. But we can start using it for that and rename later. There is the `formal_verification` directory that has all the stuff on other projects that we could also repurpose. But `fstar-helpers` sounds the easiest for now.

minicore/src/abstractions/bit.rs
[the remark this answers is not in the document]
W95Psp: Resolving this issue since we don't know where to put it otherwise.

minicore/src/arch/x86.rs
W95Psp: Ah, no, actually, the specs should not be guarded: the specs are for modeling the x86 architecture, but there is no reason to disable them for other archs.

W95Psp:
I did the changes, but I'm not sure about the docs of the intrinsics actually.
Those are specifications: so adding the imperative style spec from intel seems a bit odd.

I wonder if we should not instead expose the body of the function as docstring, somehow.
Anyway, that's not clear, I think we should think about a good style before commiting to copy/pasting 10 snippets of code from the intel doc.

fstar-helpers/minicore/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "minicore"
+edition = "2021"
+publish = false
+
+[dependencies]
+rand = "0.9.0"
```
franziskuskiefer, line 7: 0.9.0 is not the latest.

```suggestion
rand = "0.9"
```

Cargo.toml
```diff
@@ -28,8 +28,9 @@ members = [
     "rsa",
     "ecdsa",
     "p256",
-    "blake2",
+    "blake2", 
 ]
+exclude = ["fstar-helpers/minicore"]
```
franziskuskiefer, line 33: Why do you want to exclude it from the workspace?

fstar-helpers/minicore/src/lib.rs
```diff
@@ -0,0 +1,2 @@
+pub mod abstractions;
```
franziskuskiefer, line 1: No comment on the library

fstar-helpers/minicore/src/arch/mod.rs
franziskuskiefer, line 1: Please don't use `mod.rs`. Use `arch.rs` on top instead.

fstar-helpers/minicore/src/arch/x86.rs
```diff
@@ -0,0 +1,289 @@
+/// This is a (partial) mirror of [`core::arch::x86`] and [`core::arch::x86_64`].
```
franziskuskiefer, line 1: This is not a mirror. The functions have completely different function signatures. Please let us know what it is supposed to be.

franziskuskiefer (commented):
looks good with a couple nits.
Also, please go through the comments so that we know what's still open.

fstar-helpers/minicore/src/arch/x86.rs
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+}
+
+#[allow(non_camel_case_types)]
+type __m256i = BitVec<256>;
+#[allow(non_camel_case_types)]
+type __m128i = BitVec<128>;
+
+pub fn _mm_storeu_si128(output: *mut __m128i, a: __m128i) {
```
franziskuskiefer, line 60: So you have a version of the intrinsics here using the `BitVec` type instead of the registers?

fstar-helpers/minicore/src/arch/x86.rs
W95Psp: This crate is supposed to be a partial model of libcore, meaning that every item that exists both in core and in minicore should share the same type signature.

I think that mostly holds. I would like a CI to check that, though.

That's what I mean by mirror, I will update the comment to explain that.

W95Psp: I pushed e33ea12b

fstar-helpers/minicore/src/arch/x86.rs
```diff
@@ -0,0 +1,289 @@
...
+            let shift = counts[i / 32];
+            if nth_bit as i128 >= shift as i128 {
+                vector[i - shift as usize]
+            } else {
+                Bit::Zero
+            }
+        })
+    }
+
+    pub fn mm256_sllv_epi32_u32(
```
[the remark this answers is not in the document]
W95Psp: 👍 fixed, those should be exposed

fstar-helpers/minicore/src/arch/x86.rs
W95Psp: Yes, `minicore` is all about defining the semantics of each functions via concrete, simple (and often inefficient) data structures.

E.g. SIMD works over bit arrays instead of registries, arithmetic works with math integers instead of machine integers.

Cargo.toml
W95Psp: I can include it, it felt like it was nicer to exclude

minicore/src/abstractions/bitvec.rs
```diff
@@ -0,0 +1,99 @@
+//! This module provides a specification-friendly bit vector type.
+use super::bit::{Bit, MachineInteger};
+
+// TODO: this module uses `u128/i128` as mathematic integers. We should use `hax_lib::int` or bigint.
```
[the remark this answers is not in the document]
W95Psp: Yes, that's supposed to be a follow up

minicore/src/arch/x86.rs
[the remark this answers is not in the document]
W95Psp: Resolving this conversation, see my comment https://github.com/cryspen/libcrux/pull/815#issuecomment-2668647082

W95Psp:
I think I fixed everything.
Only one discussion is open, see above, not sure if I should do anything. Feel free to resolve if that's OK to you @franziskuskiefer.

fstar-helpers/minicore/src/arch/x86.rs
franziskuskiefer: Yeah, the question was more about that we should probably add more docs. But it's fine for now. Let's add more when we do more with this.

franziskuskiefer (approved):
Thanks, let's get this in and start using and improving as we go.

Issue #822, opened 2025-02-12
Add test that fails if the wrong ciphertext length is used in Chacha20Poly1305


keks:
So far, we are using the wrong ciphertext length when decrypting in chacha20poly1305. This was caught because the benchmark fails. This doesn't fail for the wycheproof tests, which is why we hadn't caught this earlier.

We need to add a test that fails when we use the wrong length to prevent regressions.

PR #823, opened 2025-02-12
fix: chachapoly: use the correct ciphertext length

keks:
@wysiwys pointed out that the decrypt benchmarks are failing, so we jumped on a call to fix this. The issue was that we were using passing the length including the MAC tag to the hacl function, but the buffer only contained the ciphertext without the tag.

This PR changes the code to use the correct length (the length of `ctxt_cpa` instead of `ctxt`).

As a followup I filed an issue to add a test that fails if we use the wrong value here: https://github.com/cryspen/libcrux/issues/822.


jschneider-bensch (approved):
Thanks!

PR #824, opened 2025-02-12
RSA: Add Wycheproof Test

keks:
This PR adds the remaining test that `libcrux::digest` does that so far was not part of the test suite for `libcrux-rsa`: wycheproof. In order to do this, I also exposed the functions for verifying and signing with variable-length keys.

Originally I also wanted to adapt the RSA benchmarks to use `libcrux-rsa`, but then I realized that we don't have any yet, so I skipped that for now.

### DIFF
--- a/rsa/Cargo.toml
+++ b/rsa/Cargo.toml
@@ -20,3 +20,8 @@ libcrux-macros = { version = "=0.0.2-beta.2", path = "../macros" }
 libcrux-sha2 = { version = "=0.0.2-beta.2", path = "../sha2", features = [
     "expose-hacl",
 ] }
+
+[dev-dependencies]
+hex = { version = "0.4.3", features = ["serde"] }
+serde = { version = "1.0.216", features = ["derive"] }
+serde_json = "1.0.133"
--- a/rsa/tests/rsa_pss.rs
+++ b/rsa/tests/rsa_pss.rs
@@ -1,6 +1,9 @@
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
 use libcrux_rsa::{
     sign, sign_2048, verify, verify_2048, DigestAlgorithm, Error, PrivateKey, PublicKey,
-    VarLenPrivateKey,
+    VarLenPrivateKey, VarLenPublicKey,
 };
 
 const MODULUS: [u8; 256] = [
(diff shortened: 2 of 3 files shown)


rsa/tests/rsa_pss.rs
```diff
@@ -171,3 +174,146 @@ fn wycheproof_single_test() {
...
+                _ => panic!("Unknown hash algorithm {}", test_group.sha),
+            };
+            for test in test_group.tests.iter() {
+                println!("Test {:?}: {:?}", test.tcId, test.comment);
+
+                let valid = test.result.eq("valid") || test.result.eq("acceptable");
+                let msg = &test.msg;
+                let signature = &test.sig;
+
+                let result = libcrux_rsa::verify_varlen(
```
jschneider-bensch, line 300: Hm, the regular `verify` works as well, since it also expects a variable length public key, so you wouldn't need to change visibility on `verify_varlen` (`sign_varlen` is not used here).

rsa/Cargo.toml
```diff
@@ -20,3 +20,8 @@ libcrux-macros = { version = "=0.0.2-beta.2", path = "../macros" }
 libcrux-sha2 = { version = "=0.0.2-beta.2", path = "../sha2", features = [
     "expose-hacl",
 ] }
+
+[dev-dependencies]
```
jschneider-bensch, line 24: In some of the other tests we use the `wycheproof` crate to get test vectors. I'm not sure it's better from a reliability perspective, but might make the test code more uniform across the workspace crates.

jschneider-bensch (commented):
Thanks! I think it's good, but still have some questions.

rsa/tests/rsa_pss.rs
keks: That's what I originally had and it was very clumsy and in the end didn't work well. Specifically, I couldn't reuse the test code between versions, because I couldn't easily make it generic, since libcrux only defines it for a few values, not "for all N: usize", which is basically what doing it generically means.

I thought exposing the variable lengths API would be nice for other users of the API as well. I'll double check if I need to do some more checks so all the preconditions of the hacl code are satisfied.

rsa/Cargo.toml
keks: Huh, interesting. I copied the approach from another crate I had been working on, but I am happy to use the `wycheproof` crate.

rsa/tests/rsa_pss.rs
jschneider-bensch: Thanks!

PR #825, opened 2025-02-13
Adapt ed25519 tests to top-level crate API

### LINKED ISSUE #798: Adapt tests in `ed25519.rs`

wysiwys:
This PR adds tests for the `libcrux-ed25519` crate, adapted from the tests in `tests/ed25519.rs`. 

- Skips `wycheproof` tests whose provided signature has a length not equal to 64 bytes
- Skips `wycheproof` test groups whose provided private key has a length not equal to 32 bytes

Resolves #798

### DIFF
--- a/ed25519/tests/ed25519.rs
+++ b/ed25519/tests/ed25519.rs
@@ -16,6 +16,9 @@ fn run_wycheproof() {
                     "Skipping test group {k}: public key length {} != 32",
                     pk.len()
                 );
+                for test in test_group.tests.into_iter() {
+                    assert_eq!(test.result, wycheproof::TestResult::Invalid);
+                }
                 continue;
             }
             let pk_buf: [u8; 32] = pk.as_slice().try_into().unwrap();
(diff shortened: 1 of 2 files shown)


ed25519/tests/ed25519.rs
```diff
@@ -0,0 +1,45 @@
...
+                continue;
+            }
+            let pk_buf: [u8; 32] = pk.as_slice().try_into().unwrap();
+
+            for (i, test) in test_group.tests.into_iter().enumerate() {
+                let comment = &test.comment;
+                println!("Test {i}: {comment}");
+
+                let sig_len = test.sig.len();
+                if sig_len != 64 {
```
jschneider-bensch, line 26:
```suggestion
                if sig_len != 64  && test.result == wycheproof::TestResult::Invalid {
```
Just to make sure that in the skipped tests the Wycheproof suite expects us to error, which seems to be the case.

jschneider-bensch (changes requested):
This looks good, thanks! I just have one more idea to make sure we get the intent of the skipped tests right.

ed25519/tests/ed25519.rs
wysiwys: Thanks! I updated the code to reflect this.

jschneider-bensch (approved):
Thanks! Let's get it merged.

PR #831, opened 2025-02-13
Migrate remaining Chacha20Poly1305 tests to new crate structure

### LINKED ISSUE #797: Adapt tests in `chachapoly.rs`

### LINKED ISSUE #822: Add test that fails if the wrong ciphertext length is used in Chacha20Poly1305

keks:
Fixes #797
Fixes #822

### DIFF
--- a/chacha20poly1305/tests/chachapoly.rs
+++ b/chacha20poly1305/tests/chachapoly.rs
@@ -3,6 +3,12 @@ use std::{fs::File, io::BufReader};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use serde_json::Value;
 
+fn randbuf<const N: usize>(rng: &mut impl rand_core::RngCore) -> [u8; N] {
+    let mut buf = [0; N];
+    rng.fill_bytes(&mut buf);
+    buf
+}
+
 pub(crate) trait ReadFromFile {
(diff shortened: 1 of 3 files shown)


chacha20poly1305/tests/chachapoly.rs
```diff
@@ -138,3 +144,48 @@ fn wycheproof() {
     );
     assert_eq!(num_tests - skipped_tests, tests_run);
 }
+
+#[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
+#[test]
+fn chachapoly_self_test() {
+    let _ = pretty_env_logger::try_init();
```
jschneider-bensch, line 151: I don't think we use this logger anywhere here, or am I missing something?

jschneider-bensch (approved):
I think it's good! Just have a question about the `pretty_env_logger` dependency.

chacha20poly1305/tests/chachapoly.rs
keks: Good point, got rid of it.

PR #833, opened 2025-02-17
Split workflows and give consistent names

jschneider-bensch:
I thought a little about how we can reduce the number of jobs that are run. For instance, we run almost all jobs on PRs, regardless of where the changes happened. I think we could be smarter and introduce dependencies such that only relevant jobs will be run, e.g. running only those jobs relating to the crate where the change happened and using `workflow_call` to call jobs that depend on that crate.
To make this work, it should be clear for every workflow, what it relates to. To that end I've split some workflows up here that did things on multiple crates and I've introduced a naming scheme that indicates the crate (or `github` for things relating to github and `workspace` for thing that should run on the whole workspace).

The nix build workflow was manually disabled by name before, so since I've changed its name it was no longer disabled. I've renamed it to `ml-kem-nix.yml.disabled` to prevent it from running. We could also delete it if we feel that it might never come in handy again.


keks:
I think this looks good in general, though I gotta say I can't quite tell what the changes of the ML-DSA part do, but it looks like there are more tests now? Not sure though

.github/workflows/mldsa-c.yml
keks, line 1: Can you talk a little bit about these changes? It looks to me that this isn't only moving around or renaming, but maybe I'm missing something

.github/workflows/mlkem-hax.yml
wysiwys, line 1: This is a very minor nitpick, but the names of these jobs don't need the `mlkem-` prefix anymore, since the jobs have been moved to a separate file. (The same applies to the jobs in `mldsa-hax.yml`.)

wysiwys (approved):
Looks good to me! I added one small suggestion about job naming. However, it's just a minor nitpick, so ignore at will :)

.github/workflows/mlkem-hax.yml
jschneider-bensch: Yes, that makes it more consistent, for sure :+1:. In general there is lots of duplication going on in all our jobs (e.g. all the build & test ones basically have the same structure). Maybe we can make our own re-usable actions for these sometime.

.github/workflows/mldsa-c.yml
jschneider-bensch: @keks: Does that clear it up?

keks: I thought I saw new code that I couldn't find the origin of, but I can't find it anymore. Looks good!

keks (approved):

keks:
Ah, is it possible that the new workflows are incompatible with the names of the required checks configured in the repository?

franziskuskiefer:
> Ah, is it possible that the new workflows are incompatible with the names of the required checks configured in the repository?

Oh you changed names. Yes that will need changes in the required jobs. But the names look better, so let's change.

PR #834, opened 2025-02-17
EDSA-P256 benchmarks

jschneider-bensch:
This adds benchmarks comparing ECDSA-P256 against `ring` and `openssl`. Only the SHA-256 variant for now.


franziskuskiefer (approved):
lgtm, thanks!

PR #836, opened 2025-02-17
Use the `libcrux_ecdh::x25519...` API for key generation

wysiwys:
This pull request updates the `x25519` benchmarks to use the `libcrux_ecdh::x25519...` API where possible.
- `libcrux_ecdh::x25519_key_gen()` for generating (private, public) key pairs
-  `libcrux_ecdh::x25519_generate_secret()` for generating individual private keys, when needed


jschneider-bensch (approved):
Thanks!
I'm not sure hacl would actually fail early if we didn't provide a good value for the secret key, but this way we're safe in any case.

PR #839, opened 2025-02-18
Readmes with verification status

### LINKED ISSUE #672: Add labels for Verification Status
All the crates here should get verified, potentially with a hacl-rs note.

Prompted by https://github.com/cryspen/libcrux/pull/659#pullrequestreview-2428753961.

jschneider-bensch:
This PR introduces a new verification badge for HACL-r and adds an explanatory note to the Readme of all the crates that come from HACL-rs. For most crates there's not much else in these Readmes for now, but this is a start.

Fixes #672


p256/Readme.md
keks, line 1: I think this file should contain a note that the crate is just an internal dependency and shouldn't really be used directly unless you really know what you are doing.

keks (changes requested):
Just one note, otherwise looks good!

franziskuskiefer (approved):

PR #840, opened 2025-02-19
RSA: Add benchmarks and another test.

keks:
Adds benchmarks for several RSA implementations for key sizes 2024, 3072, 4096, 6144, 8192. Not all implementations support all key sizes.

There is one strange thing happening in key generation, where OpenSSL would generate keys where `d` is a bit too short and libcrux rejected it. To work around this I do rejection sampling and sample a few times until a key that looks good is found. If after ten consecutive attempts still no good key is found, the test panics. Maybe we need a better solution for this in the crate itself, that may be something we have to discuss.


benchmarks/benches/rsa.rs
```diff
@@ -0,0 +1,325 @@
...
+use pkcs1::DecodeRsaPrivateKey;
+use rand_core::OsRng;
+use ring::signature::KeyPair;
+
+const MSG_STR: &str = "the test message is 32 byte long";
+const MSG: &[u8; 32] = b"the test message is 32 byte long";
+const SALT: &[u8] = b"this is the test salt";
+
+// With the openssl_keygen function above we'd sometimes get keys that are one byte too short.
+// This function sidesteps that issue by rejecting such keys and resampling on rejection.
```
franziskuskiefer, line 16: That's indeed weird. Would you need to pad it to use it with hacl maybe? But since we don't care too much about RSA right now, let's put it on the pile of things to investigate when we need it.

franziskuskiefer (approved):
Thanks! lgtm.

This also shows that it really is as slow as I expected when trying to use it. I'll report to hacl folks. But let's get this in.

benchmarks/benches/rsa.rs
keks: This is due to the way our wrapper code is written - it always assumes that the keys are BITS/8 long. We could loosen that restriction, but (a) at the time I didn't know we would need to loosen it and (b) I am not sure what we'd want to loosen it to. How short is okay? N-1? N-2? Is there a good reason at all for `d` to be shorter? I have opened https://github.com/cryspen/libcrux/issues/841.

franziskuskiefer: Yep, RSA is weird. But a byte less would be legal. Anyway, thanks for filing.

Issue #544, opened 2024-08-29
Enable benchmarks on merge queue


franziskuskiefer:
Fix and enable benchmarks again on the merge queue.

franziskuskiefer:
The merge queue is disabled for now until this is fixed.

franziskuskiefer:
#842 got merged 🥳 
A few things I noticed here
 
- The mlkem fuzz is sill run on the merge queue. It should probably only be run on the merge queue, but then must also influence the status
- mlkem c build and tests don't appear to be run on the merge queue. They should be only run on the merge queue, not PRs

wysiwys:
Thanks @franziskuskiefer! I opened a pull request here to fix both of those issues: #853

PR #842, opened 2025-02-19
Configure workflows in preparation for reenabling the merge queue

wysiwys:
This pull request ensures that only the `mldsa-c.yml` and  `mlkem-c.yml` workflow jobs are run by the merge queue (see #544 )


franziskuskiefer (dismissed):
Thanks. Let's try this

franziskuskiefer (commented):
Actually, the merge queue is waiting for all the required jobs now, but never gets them. So I guess it's not as easy.

wysiwys:
These changes update the workflows to instead be triggered by both `pull_request` and `merge_group` triggers. 

Then, in each workflow with required jobs, an `if` key is used to selectively run the affected jobs based on the `github.event_type` (e.g. `pull_request`, `merge_group`).

franziskuskiefer:
Let's update this once #794 got merged.

franziskuskiefer (approved):
Let's see if this works :)

franziskuskiefer (approved):

wysiwys:
Status check jobs have now been added where necessary. Also, they are named in such a way that collisions between status check job names are avoided.

franziskuskiefer (approved):

PR #843, opened 2025-02-20
Fix Annotation to enable extraction with Hax #1301

karthikbhargavan:
This is a minimal change to libcrux to enable extraction with Hax #1301


W95Psp (approved):
Looks fine to temporarily escape from the hax/libcrux CI dep cycle.
Then https://github.com/cryspen/libcrux/pull/788 will revert that change in favor of an overall better style.

edit: Ah, my approval is not doing anything on this repo, seems like.

karthikbhargavan:
I guess the two failing tasks here are unavoidable.

franziskuskiefer (approved):
lgtm when CI is happy

PR #845, opened 2025-02-21
update getrandom

franziskuskiefer:


jschneider-bensch (approved):
Looks good, I think, thanks! I assume we no longer need feature `js` for emscripten builds, then?

PR #851, opened 2025-02-24
[PSQ] Implement `Credential` for Ed25519 signatures

### LINKED ISSUE #720: [PSQ] Implement signature-based client authentication
At the moment, the `Credential` trait for client authentication is only implemented by the `NoAuth` dummy implementation that does not provide client authentication. It would be good to have at least one other implementation that does provide authentication, e.g. using ECDSA signatures.

jschneider-bensch:
This PR adds an implementation of the `Credential` trait for Ed25519 signatures.

I'm not sure `Credential` is generic enough to go into the `traits` crate as an interface for signatures. I think it would require some more work to make it so, and that should probably go into a follow-up.

Fixes #720

### DIFF
--- a/libcrux-psq/Cargo.toml
+++ b/libcrux-psq/Cargo.toml
@@ -25,10 +25,12 @@ classic-mceliece-rust = { version = "2.0.0", features = [
 rand = { version = "0.9" }
 rand_old = { version = "0.8", package = "rand", optional = true }
 libcrux-ecdh = { version = "0.0.2-beta.3", path = "../libcrux-ecdh" }
+libcrux-ed25519 = { version = "0.0.2-beta.3", path = "../ed25519" }
 libcrux = { version = "0.0.2-beta.3", path = ".." }
 
 [dev-dependencies]
 criterion = "0.5"
+openssl = "0.10" # for generating Ed25519 private keys
 
 [features]
 classic-mceliece = ["dep:classic-mceliece-rust", "rand_old"]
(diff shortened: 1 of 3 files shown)


libcrux-psq/Cargo.toml
```diff
@@ -25,10 +25,12 @@ classic-mceliece-rust = { version = "2.0.0", features = [
 rand = { version = "0.9" }
 rand_old = { version = "0.8", package = "rand", optional = true }
 libcrux-ecdh = { version = "0.0.2-beta.3", path = "../libcrux-ecdh" }
+libcrux-ed25519 = { version = "0.0.2-beta.3", path = "../ed25519" }
 libcrux = { version = "0.0.2-beta.3", path = ".." }
 
 [dev-dependencies]
 criterion = "0.5"
+openssl = "0.10" # for generating Ed25519 private keys
```
franziskuskiefer, line 33: We can ecdh keys, maybe add a keygen function ed25519? That should be quick enough and then we wouldn't need this.

jschneider-bensch: You mean, convert a curve25519 key to Edwards form? I don't think we have that anywhere at the moment.

franziskuskiefer: There's key generation for ed25519 here https://github.com/cryspen/libcrux/blob/main/src/signature.rs#L475

jschneider-bensch: Huh! My bad, thanks for pointing it out.

PR #854, opened 2025-02-26
Run ml-kem proofs on CI.

### LINKED ISSUE #733: [ML-KEM] Run proofs on CI

maximebuyse:
Closes #733


.github/workflows/mlkem-hax.yml
```diff
@@ -87,6 +91,25 @@ jobs:
       - name: 🏃 Lax ML-KEM crate
         working-directory: libcrux-ml-kem
         run: ./hax.py prove --admit
+
+  prove:
+    runs-on: ubuntu-latest
```
franziskuskiefer, line 96: I don't think the Github runner will be able to handle this. This may need to go on the self hosted runner

maximebuyse: It failed indeed, let's try the self-hosted

maximebuyse:
I have been struggling with this for a bit. The main problem is that, as we have no lock file, cargo pins `base64ct` to the latest version. This version requires rust to be at least version 1.85. I changed the jobs to use this version of rust, and made the necessary modifications to the code so that everything works with this version. It fixes the problems with the current CI. 
However to run ML-kem proofs on CI wit the self hosted runners, we need to install hax with nix, and this pins rust version to what is used by hax (which is not necessarily a bad thing but is some 1.84.0-nightly). So the problem comes back. I am not sure of how to proceed, there are several possible ways:
- Try to override the version of rust to 1.85.0 (I don't know enough nix so I didn't succeed in this)
- Update the version of rustc used by hax to latest nightly (this requires some more work on the hax side, and we could have more problems)
- Pin the necessary dependencies (`base64ct` and maybe more) so that older rust versions can be used
Not sure which solution is the best, I'll try the last one.

.github/workflows/mlkem-hax.yml
maximebuyse: I had a few complications to test on the self-hosted runner, still not solved, but I was able to try running the proof manually in the same environment and it fails with:
```
Error 19 at Libcrux_ml_kem.Vector.Portable.Arithmetic.fst(399,15-399,29):
  - Assertion failed
  - SMT solver says:
      unknown because (incomplete quantifiers) (rlimit=500; fuel=0; ifuel=1)
    Note:
      - 'canceled' or 'resource limits reached' means the SMT query timed out,
      so you might want to increase the rlimit
      - 'incomplete quantifiers' means Z3 could not prove the query, so try to
      spell out your proof out in greater detail, increase fuel or ifuel
      - 'unknown' means Z3 provided no further reason for the proof failing
  - See also /home/actions-runner/cryspen-pub-1/_work/libcrux/libcrux/libcrux-ml-kem/proofs/fstar/spec/Spec.Utils.fsti(136,57-136,74)
```
So it seems like we also lack resources on the self-hosted runner

karthikbhargavan: I can debug some of this, but yes, we likely need a lot of memory to verify some files. This can be optimized and made lower over time, but at the moment, it does.

karthikbhargavan:
I think pinning base64ct seems like the best path for now. Then let us revisit the question of how and when to pin etc.

franziskuskiefer:
Yes, just pin an older version `base64ct = "=1.6.0"` should do it

.github/workflows/mlkem-hax.yml
maximebuyse: Ok, so seems like we might want to wait before having this on CI?

maximebuyse:
> Yes, just pin an older version `base64ct = "=1.6.0"` should do it

Seems like something changed on main so this is no longer a problem.

franziskuskiefer:
@maximebuyse what's the state here?

maximebuyse:
> @maximebuyse what's the state here?

When I tried manually running inside the self-hosted runner environment, the proof failed, probably because it was lacking resources, so it seems like this will require some changes. We can try to make the proofs lighter, or to have more resources on the server, but with F* things are always a bit unpredictable.

franziskuskiefer:
@karthikbhargavan can you look at the proofs here? The server should have enough resources to run this. Maybe something broke?

maximebuyse:
Switching to latest F* solved the issue. The job takes about half an hour. Should we merge? I have changes in here to fix some warnings, they are not necessary anymore but probably good anyway.

franziskuskiefer:
Great 🥳
What about the hints? Shall we keep them?
This may clash with #871 on the hax-lib update.

Let's only run this job on the merge queue. It's too long for PRs.

maximebuyse:
> Great 🥳 What about the hints? Shall we keep them? This may clash with #871 on the hax-lib update.
> 
> Let's only run this job on the merge queue. It's too long for PRs.

I removed the hints and the proof still succeeds, the runtime is not significantly worse so I guess we don't need them. I also took care of conflicts and made the proofs merge queue only.

.github/workflows/mlkem-hax.yml
```diff
@@ -82,11 +82,36 @@ jobs:
...
         working-directory: libcrux-ml-kem
         run: ./hax.py prove --admit
+
+  prove:
+    runs-on: self-hosted
+    if: ${{ github.event_name == 'merge_group' || github.event_name == 'workflow_dispatch' }}
+    needs:
+      - extract
+      - diff
+      - lax
```
franziskuskiefer, line 97: We don't actually run these other jobs on the merge queue so we could skip them here.

franziskuskiefer (approved):
Thanks! lgtm with that one nit on the `needs`.

PR #855, opened 2025-02-26
PSQ: Authenticated Registration Example

jschneider-bensch:
This PR contains an example for authenticated PSQ registration using Ed25519.

In the process I've updated the `Credential` trait (now I'm calling it `Authenticator` since the credential is an object of the authentication process itself). To authenticate during the protocol, the initiator sends with the message a signature and a credential. The credential can be validated using a certificate, which produces a corresponding verification key. It is thus assumed that the responder has obtained a certificate on the initiator's credential out of band and can in this way verify the signature in the message.
For the simple Ed25519 authenticator the credential is the same thing as the certificate and the responder just checks that the key received as credential matches the key obtained previously as certificate. Other possiblities for implementing an `Authenticator` could include a PKI, where e.g. the certificate is a PKI-certificate and the credential is an initiator identifier in the PKI. Then the credential can be validated by retrieving the initiator verification key from the PKI.


libcrux-psq/examples/Readme.md
```diff
@@ -0,0 +1,42 @@
...
+The example uses ML-KEM 768 as the PSQ KEM and Ed25519 for client
+authentication. Another option for the PSQ KEM is Classic McEliece
+(under feature `classic-mceliece`) using the
+[`classic-mceliece-rust`](https://crates.io/crates/classic-mceliece-rust/2.0.2)
+crate.
+
+The example implements a setup phase that is to be considered outside
+of the protocol. In this phase the initiator generates a signing key
+and sends over a certificate to the responder (In this case, the
+certificate is just the verification itself).  Similarly, the
```
franziskuskiefer, line 25:
```suggestion
certificate is just the verification key itself).  Similarly, the
```

franziskuskiefer (approved):
lgtm

PR #859, opened 2025-02-28
C code tools (detect)

karthikbhargavan:
Our C code build relies on:
* clang-format
* cmake
* ninja

We should tell the user to install these tools, and not pin to a specific version unless needed.


karthikbhargavan:
Bad idea. We need to pin these.

PR #861, opened 2025-02-28
psq: pure rust libcrux

franziskuskiefer:
Use Rust chacha20 in psq and drop legacy libcrux dependency.

### DIFF
--- a/chacha20poly1305/src/impl_hacl.rs
+++ b/chacha20poly1305/src/impl_hacl.rs
@@ -1,6 +1,47 @@
 use crate::{AeadError, KEY_LEN, NONCE_LEN, TAG_LEN};
 
-/// The ChaCha20-Poly1305 AEAD encryption function. Writes the concatenation of the ciphertexoft
+const DETACHED: bool = true;
+const NOT_DETACHED: bool = false;
+
+/// The ChaCha20-Poly1305 AEAD encryption function with detached tag.
+pub fn encrypt_detached<'a>(
+    key: &[u8; KEY_LEN],
+    ptxt: &[u8],
+    ctxt: &'a mut [u8],
+    tag: &'a mut [u8; TAG_LEN],
+    aad: &[u8],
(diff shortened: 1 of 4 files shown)


chacha20poly1305/src/impl_hacl.rs
```diff
@@ -1,29 +1,60 @@
...
+
+/// The ChaCha20-Poly1305 AEAD encryption function with detached tag.
+pub fn encrypt_detached<'a>(
     key: &[u8; KEY_LEN],
     ptxt: &[u8],
     ctxt: &'a mut [u8],
+    tag: &'a mut [u8; TAG_LEN],
     aad: &[u8],
     nonce: &[u8; NONCE_LEN],
 ) -> Result<(&'a [u8], &'a [u8; TAG_LEN]), AeadError> {
```
jschneider-bensch, line 14: Why return the slices in the `Result`, still, instead of e.g. `Result<(), AeadError>`?

chacha20poly1305/src/impl_hacl.rs
```diff
@@ -1,29 +1,60 @@
...
+    key: &[u8; KEY_LEN],
+    ptxt: &[u8],
+    ctxt: &'a mut [u8],
+    aad: &[u8],
+    nonce: &[u8; NONCE_LEN],
+) -> Result<(&'a [u8], &'a [u8; TAG_LEN]), AeadError> {
+    let (ptxt_len, aad_len) = encrypt_checks(ptxt, ctxt, aad, NOT_DETACHED)?;
+
     // ensure destination slice has just the right length
     let (ctxt_cpa, tag) = ctxt.split_at_mut(ptxt.len());
```
jschneider-bensch, line 59: Could use `ptxt_len` here now.

jschneider-bensch (approved):
Thanks! (For cleaning up around the AEAD, too :pray: )

PR #864, opened 2025-03-03
Reactivate mlkem benchmarks and move to merge queue

wysiwys:
This PR reactivates the mlkem benchmarks, and modifies the workflow so that the benchmarks job will only be run on the merge queue. It also adds the status check `mlkem-bench-status`.

See #544


jschneider-bensch (approved):
This looks reasonable to me!

PR #865, opened 2025-03-03
Add spec and proofs for MLDSA portable sample functions

mamonet:
This PR adds specs for portable sample module in `Spec.MLDSA.Math.fst` as well as adding pre/post-conditions for sample functions. It also adds proofs for that module.


karthikbhargavan:
I will be looking at this PR tomorrow. It will be resolved (one way or another) this week.

PR #871, opened 2025-03-05
`hax-lib`: update to `0.2.0`

W95Psp:
This PR:
 - updates `hax-lib`
 - defines a workspace dependency `hax-lib`, so that there is only one version of `hax-lib` in the repo

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ readme = "Readme.md"
 allow-branch = ["main"]
 
 [workspace.dependencies]
-hax-lib = { version = "0.2", git = "https://github.com/hacspec/hax/" }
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -44,6 +44,9 @@ readme = "Readme.md"
 [workspace.metadata.release]
 allow-branch = ["main"]
 
+[workspace.dependencies]
+hax-lib = { version = "0.2", git = "https://github.com/hacspec/hax/" }
```
franziskuskiefer, line 48: Let's update this while we're here.

```suggestion
hax-lib = { version = "0.2", git = "https://github.com/cryspen/hax/" }
```

franziskuskiefer (approved):
Thanks

PR #875, opened 2025-03-06
[Merge queue] make status checks explicitly depend on all preceding jobs in graph

wysiwys:
This PR fixes an issue in the GitHub Actions workflows `mlkem-c.yml` and `mldsa-c.yml` where, if one of the predecessors of a status check's needed jobs fails, the status check itself may succeed, due to the earlier predecessor not being included in its `needed` list. The fix is to explicitly include all predecessors of the status check in its `needed` list.

cc #544

### DIFF
--- a/.github/workflows/mldsa-c.yml
+++ b/.github/workflows/mldsa-c.yml
@@ -39,6 +39,8 @@ jobs:
   mldsa-c-tests-status:
     if: ${{ always() }}
     needs:
+      - setup
+      - extract-header-only
       - diff-header-only
       - build-header-only
     runs-on: ubuntu-latest
--- a/.github/workflows/mlkem-c.yml
+++ b/.github/workflows/mlkem-c.yml
@@ -39,6 +39,9 @@ jobs:
   mlkem-c-tests-status:
     if: ${{ always() }}
     needs:
+      - setup
+      - extract
+      - extract-header-only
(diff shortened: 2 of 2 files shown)


franziskuskiefer (approved):
lgtm

jschneider-bensch (approved):
That's an annoying issue, thanks for fixing!

Just want to make sure I understand what happens: If, say `extract` fails for `mlkem-c.yml`, then `diff` would be skipped because it has `extract` in it's `needs` field. In the [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) it says that 
> If a job fails or is skipped, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue.

So, since the status check has the conditional to run always, it isn't skipped, but it will instead succeed because (disregarding the other dependencies) `diff` was skipped instead of failing?

wysiwys:
Thanks for the review @jschneider-bensch!  Yes, that is correct. If `extract` were to fail, then all of the jobs that rely on it (except for the ones that have an `always()` in their `if` field) would be skipped. Without `extract` in the status check's `needs` field, and not the `extract` job itself, the status check job would then succeed (since all of its `needs` jobs were skipped), even though we want it to fail in this case.

PR #879, opened 2025-03-11
pushed platform-specific modules behind not-hax

karthikbhargavan:
This PR avoids exposing "static mut" variables to hax from sys/platform. 

Some platform-specific modules were being fed to hax even though they were needed and they were using `static mut` which is now rejected with an error by hax. (Earlier these items were being ignored by hax).

This PR also refreshes the F* code using latest hax main.


karthikbhargavan:
This PR also regenerates the F* with latest hax and so should unblock CI (if it succeeds)

sys/platform/src/lib.rs
```diff
@@ -9,12 +9,12 @@
 #[macro_use]
 extern crate std;
 
-#[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), not(hax)))]
+#[cfg(all(not(hax), any(target_arch = "x86", target_arch = "x86_64"), not(hax)))]
```
franziskuskiefer, line 12: We have `not(hax)` now twice here.

karthikbhargavan: fixed

franziskuskiefer (approved):

PR #880, opened 2025-03-11
Add tracing types and a proc macro for using it to measure functions

keks:
This PR adds the tracing tools we agreed on adding at our meeting yesterday. It adds the new crate `libcrux-test-utils`, which contains  the tracing library, and the new `trace_span` proc macro to `libcrux-macros`.

### DIFF
--- a/test-utils/README.md
+++ b/test-utils/README.md
@@ -11,10 +11,17 @@ is written to the specified trace.
 Usually, that trace will be a static variable with interior mutability. If in
 doubt, define it something like this:
 
+In this example we use `cfg` and `cfg_attr` to ensure that the tracing only runs
+during testing. In many cases one might want to also restrict this to only trace
+if a certain feature is enabled. This can be done using the same mechanism.
+
 ```rust
-lazy_static::lazy_static! {
-    static ref TRACE: MutexTrace<&'static str, Instant> = Default::default();
-}
+#[cfg(test)]
+use std::sync::LazyLock;
+
+#[cfg(test)]
+static TRACE: LazyLock<MutexTrace<&'static str, Instant>> =
+    LazyLock::new(|| MutexTrace::default());
 ```
 
 The `MutexTrace` can be defined and used as a static variable without unsafe, but,
@@ -24,14 +31,15 @@ depending on the setting, my introduce too much overhead. Any type that implemen
 Then, annotate a function like this:
 
 ```rust
-#[libcrux_macros::trace_span("trace_me", TRACE)]
+#[cfg_attr(test, libcrux_macros::trace_span("some_traced_function", TRACE))]
 fn this_function_is_traced() {
   // ...
 }
 ```
 
-The macro is called `trace_span` because it traces a start and end. There also
-are on-the-fly tracing facilities.
+The macro is called `trace_span` because it traces a start and end, identified
+by a label. The type of the label can be chosen generically, here it is
+`&'static str`. There also are on-the-fly tracing facilities.
 
 After the code in question ran, the trace can be inspected. Due to the use
 of  interior mutability, there is no generic way to get a reference to the
--- a/test-utils/README.md
+++ b/test-utils/README.md
@@ -1,0 +1,42 @@
+# `libcrux-test-utils`
+
+This crate contains tools for testing, tracing, profiling and benchmarking.
+
+## Tracing (`tracing`)
+
+This module provides tools for annotating functions such that calls are traced.
+That means that whenever the function is called and when it returns, an entry
+is written to the specified trace.
+
+Usually, that trace will be a static variable with interior mutability. If in
+doubt, define it something like this:
+
+```rust
+lazy_static::lazy_static! {
+    static ref TRACE: MutexTrace<&'static str, Instant> = Default::default();
+}
+```
+
+The `MutexTrace` can be defined and used as a static variable without unsafe, but,
+depending on the setting, my introduce too much overhead. Any type that implements
+`Trace` works.
+
+Then, annotate a function like this:
+
+```rust
+#[libcrux_macros::trace_span("trace_me", TRACE)]
+fn this_function_is_traced() {
+  // ...
+}
+```
+
+The macro is called `trace_span` because it traces a start and end. There also
+are on-the-fly tracing facilities.
+
+After the code in question ran, the trace can be inspected. Due to the use
+of  interior mutability, there is no generic way to get a reference to the
+inner slice; therefore, it is easiest to just clone it.
+
+```rust
+println!((*TRACE).clone().report());
+```
--- a/test-utils/src/tracing.rs
+++ b/test-utils/src/tracing.rs
@@ -1,0 +1,180 @@
+//! This module contains types for producing timed traces of program runs. Make sure to also look
+//! at the `trace_span` macro from `libcrux_macros`.
+
+use std::{
+    borrow::Borrow,
+    cell::{Ref, RefCell},
+    fmt::Display,
+    sync::Mutex,
+};
+
+/// This trait descries a trace that is behind some sort of interior mutability mechanism. It can
+/// log trace events and later make these available. This is usually an argument to the
+/// `trace_span` function attribute macro in `libcrux-macros``, but it can also be called manually.
+///
+/// When used with the `trace_span` macros, this needs to be a global static. For defining and
+/// instantiating this, look into the `lazy_static` crate.
+pub trait Trace: Sized {
+    /// The label type used in events. Typically either `&'static str` or an enum.
+    type Label: Clone;
+
+    /// The type used for timing the events. Typically `std::time::Instant` or a cycle counter.
+    type TimeStamp: TimeStamp;
+
+    /// Writes an event to the trace.
+    fn emit(&self, ev: TraceEvent<Self::Label, Self::TimeStamp>);
+
+    /// Returns a vector of all entries in the trace.
+    fn report(self) -> Vec<TraceEvent<Self::Label, Self::TimeStamp>>;
+
+    /// Emits an event of type [EventType::SpanOpen] and returns a [`SpanHandle`] that emits another
+    /// [`EventType::SpanClose`] when dropped.
+    fn emit_span(&self, label: Self::Label) -> SpanHandle<Self> {
+        let event = TraceEvent {
+            ty: EventType::SpanOpen,
+            at: Self::TimeStamp::now(),
+            label: label.clone(),
+        };
+
+        self.emit(event);
+        SpanHandle(self, Some(label))
+    }
+
+    /// Emits an [`EventType::OnTheFly`] event.
+    fn emit_on_the_fly(&self, label: Self::Label) {
+        let event = TraceEvent {
+            ty: EventType::OnTheFly,
+            at: Self::TimeStamp::now(),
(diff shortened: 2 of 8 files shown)


test-utils/tests/tracing_macros.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        assert_eq!(open.ty, EventType::SpanOpen);
+        assert_eq!(open.label, "maybe_slow");
+
+        assert_eq!(close.ty, EventType::SpanClose);
+        assert_eq!(close.label, "maybe_slow");
+
+        let run_time = close.at - open.at;
+        assert!(run_time.as_micros() > 99500);
+        assert!(run_time.as_micros() < 100500);
```
wysiwys, line 36:
```suggestion
        assert!(run_time.as_micros() > 99000);
        assert!(run_time.as_micros() < 101000);
```
Widening the acceptable interval a bit could help avoid sporadic failures

test-utils/tests/tracing_macros.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        assert_eq!(open.ty, EventType::SpanOpen);
+        assert_eq!(open.label, Label::MaybeSlow);
+
+        assert_eq!(close.ty, EventType::SpanClose);
+        assert_eq!(close.label, Label::MaybeSlow);
+
+        let run_time = close.at - open.at;
+        assert!(run_time.as_micros() > 99500);
+        assert!(run_time.as_micros() < 100500);
```
wysiwys, line 72:
```suggestion
        assert!(run_time.as_micros() > 99000);
        assert!(run_time.as_micros() < 101000);
```
And here too

test-utils/src/tracing.rs
```diff
@@ -0,0 +1,180 @@
...
+//! at the `trace_span` macro from `libcrux_macros`.
+
+use std::{
+    borrow::Borrow,
+    cell::{Ref, RefCell},
+    fmt::Display,
+    sync::Mutex,
+};
+
+/// This trait descries a trace that is behind some sort of interior mutability mechanism. It can
```
wysiwys, line 11:
```suggestion
/// This trait describes a trace that is behind some sort of interior mutability mechanism. It can
```

test-utils/src/tracing.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// This type emits a [`EventType::SpanClose`] event on drop.
+pub struct SpanHandle<'a, T: Trace>(&'a T, Option<T::Label>);
+
+impl<'a, T: Trace> Drop for SpanHandle<'a, T> {
+    fn drop(&mut self) {
+        let event = TraceEvent {
+            ty: EventType::SpanClose,
+            at: T::TimeStamp::now(),
+            label: self.1.take().unwrap(),
```
wysiwys, line 93: For `SpanHandle.1`, could this also be just a `T::Label`?

keks: No, because we need to be able to move the value out despite only having a &mut reference. We can do that using `Option::take`. I added a comment that explains this.

test-utils/tests/tracing_macros.rs
keks: I changed this to +- 0.7ms everywhere, I hope it's not an issue anymore.

wysiwys: All the tests are passing now

franziskuskiefer (commented):
I'm not sure I understand how this is going to be used in the end. Can you add a test that shows usage?

test-utils/src/tracing.rs
```diff
@@ -0,0 +1,190 @@
...
+    type Label: Clone;
+
+    /// The type used for timing the events. Typically `std::time::Instant` or a cycle counter.
+    type TimeStamp: TimeStamp;
+
+    /// Writes an event to the trace.
+    fn emit(&self, ev: TraceEvent<Self::Label, Self::TimeStamp>);
+
+    /// Returns a vector of all entries in the trace.
+    fn report(self) -> Vec<TraceEvent<Self::Label, Self::TimeStamp>>;
```
jschneider-bensch, line 28: This means it can only be instantiated if we have an allocator, right?

jschneider-bensch (commented):
The tests kind of tell me how to use it, but it would be nice to have some more documentation in a README or a commented example of how to use the macros.

test-utils/src/tracing.rs
keks: Note that right now we also only have implementations for `std::time::Instant`, which doesn't exist in `core` and `alloc` - but at least I was able to keep that out of the interfaces.

test-utils/tests/tracing_macros.rs
```diff
@@ -0,0 +1,74 @@
...
+
+use libcrux_test_utils::tracing::{EventType, MutexTrace, Trace as _, TraceEvent};
+
+mod str_labels {
+
+    use super::*;
+
+    lazy_static::lazy_static! {
+        static ref TRACE_STR: MutexTrace<&'static str, Instant> = Default::default();
+    }
```
franziskuskiefer, line 14: lazy static isn't needed anymore for things like this. You could use `LazyLock` for example.

test-utils/README.md
```diff
@@ -0,0 +1,42 @@
...
+fn this_function_is_traced() {
+  // ...
+}
+```
+
+The macro is called `trace_span` because it traces a start and end. There also
+are on-the-fly tracing facilities.
+
+After the code in question ran, the trace can be inspected. Due to the use
+of  interior mutability, there is no generic way to get a reference to the
```
franziskuskiefer, line 37: That's a little unfortunate. But how is related to interior mutability? In the usages here it seems like this is because we have a static refence that we can't consume/drop?

test-utils/README.md
```diff
@@ -0,0 +1,42 @@
+# `libcrux-test-utils`
+
+This crate contains tools for testing, tracing, profiling and benchmarking.
+
+## Tracing (`tracing`)
+
+This module provides tools for annotating functions such that calls are traced.
```
franziskuskiefer, line 7: What's the story for putting this into production code?
We need a way to enable this, while making it no-ops in regular builds.

franziskuskiefer (commented):
Thanks for the readme!
I left a couple thoughts.

test-utils/README.md
keks: Ah, yes, there are two things here.

I think for basically all trace inspection tasks it would be enough to have references to the trace entries. However, the interior mutability types mean that we can't get references into the data, so we have to use owned values. And yes, the only way to get an owned value from a static reference is to clone it.

Actually I suppose we could also use a `Mutex<Option<T>>` and then `.take()` that, or create a new `T` and use `mem::swap`, but I'm not sure that would be a lot better. This is only for testing, after all the benchmarks ran, so I am not sure we need to be super performant here.

test-utils/tests/trace.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    assert_eq!(open.ty, EventType::SpanOpen);
+    assert_eq!(open.label, "test block");
+
+    assert_eq!(close.ty, EventType::SpanClose);
+    assert_eq!(close.label, "test block");
+
+    let run_time = close.at - open.at;
+    assert!(run_time.as_micros() > 2300);
+    assert!(run_time.as_micros() < 3700);
```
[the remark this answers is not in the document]
keks: That's strange, that means that the timings are off nearly a full millisecond. Can you tweak these numbers and let me know what window of deviation works reliable on your machine?

test-utils/src/tracing.rs
jschneider-bensch: That makes sense, thanks for elaborating!

test-utils/README.md
franziskuskiefer: Yeah it's not a real issue, more an annoyance that it seems necessary.

test-utils/tests/trace.rs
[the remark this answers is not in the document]
franziskuskiefer: This works for me. But adding times here seems dangerous in general? Running it on a slower machine may still fail.

```diff
diff --git a/test-utils/tests/trace.rs b/test-utils/tests/trace.rs
index 3abfb5b4..77897835 100644
--- a/test-utils/tests/trace.rs
+++ b/test-utils/tests/trace.rs
@@ -27,7 +27,7 @@ fn test_refcell_trace_entries() {
 
     let run_time = close.at - open.at;
     assert!(run_time.as_micros() > 2300);
-    assert!(run_time.as_micros() < 3700);
+    assert!(run_time.as_micros() < 3900);
 }
 
 #[test]
@@ -67,7 +67,11 @@ fn test_trace_report(trace: &(impl Clone + Trace<Label = &'static str, TimeStamp
 
     let run_time = close.at - open.at;
     assert!(run_time.as_micros() > 2300);
-    assert!(run_time.as_micros() < 3700);
+    assert!(
+        run_time.as_micros() < 3900,
+        "runtime: {}",
+        run_time.as_micros()
+    );
 }
 
 fn test_nested_trace_report(
@@ -104,5 +108,9 @@ fn test_nested_trace_report(
 
     let run_time_outer = close_outer.at - open_outer.at;
     assert!(run_time_outer.as_micros() > 4300);
-    assert!(run_time_outer.as_micros() < 5700);
+    assert!(
+        run_time_outer.as_micros() < 6400,
+        "run_time_outer: {}",
+        run_time_outer.as_micros()
+    );
 }
diff --git a/test-utils/tests/tracing_macros.rs b/test-utils/tests/tracing_macros.rs
index 8975cc43..f619ed92 100644
--- a/test-utils/tests/tracing_macros.rs
+++ b/test-utils/tests/tracing_macros.rs
@@ -33,7 +33,11 @@ mod str_labels {
 
         let run_time = close.at - open.at;
         assert!(run_time.as_micros() > 9300);
-        assert!(run_time.as_micros() < 10700);
+        assert!(
+            run_time.as_micros() < 13000,
+            "runtime: {}",
+            run_time.as_micros()
+        );
     }
 }
 
@@ -68,6 +72,6 @@ mod enum_labels {
 
         let run_time = close.at - open.at;
         assert!(run_time.as_micros() > 9300);
-        assert!(run_time.as_micros() < 10700);
+        assert!(run_time.as_micros() < 13000);
     }
 }
```

franziskuskiefer (changes requested):
Let's not compare absolute times to avoid breakage and then get this merged.

franziskuskiefer:
@keks let's get this in

keks:
I just remembered you said we can't force-push or CI gets hickups... Should we try anyway or start to resolve this right now?

franziskuskiefer (commented):
> I just remembered you said we can't force-push or CI gets hickups... Should we try anyway or start to resolve this right now?

CI doesn't care. But you broke the review process with the force push because there's no diff to what has last been reviewed.

So please don't force push during the review process but only afterwards to clean up the history or such.

keks:
Ah, now I got what you meant. I didn't quite understand your point last time you mentioned this. Got it, will do. This time, I just didn't think about it.

test-utils/tests/trace.rs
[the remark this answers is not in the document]
franziskuskiefer: It looks like this is still open?

franziskuskiefer (approved):

PR #881, opened 2025-03-12
Update C extraction toolchain to current main

franziskuskiefer:
Update C extraction
- cg directories (extracted by `boring.sh`) contain C++ code now that's compatible with MSVC
- bump F* to 2025.02.17
- bump ocaml to 5.1.1
- update eurydice and karamel to current main
- update charon to a recent version (current main is broken)

Note that the `eurydice_glue.h` is currently different from upstream due to https://github.com/AeneasVerif/eurydice/pull/159. This fails the test jobs on Ubuntu and Windows.


.docker/c/install.sh
```diff
@@ -6,18 +6,20 @@ curl https://sh.rustup.rs -sSf | bash -s -- -y
...
+curl -L https://github.com/FStarLang/FStar/releases/download/v2025.02.17/fstar-v2025.02.17-Linux-x86_64.tar.gz \
     --output Fstar.tar.gz
 tar --extract --file Fstar.tar.gz
 rm -f Fstar.tar.gz
 
-curl -L https://github.com/FStarLang/binaries/raw/master/z3-tested/z3-4.8.5-x64-ubuntu-16.04.zip --output z3.zip
-unzip z3.zip
-rm -rf z3.zip
-mv z3-4.8.5-x64-ubuntu-16.04/ z3
+curl -L https://raw.githubusercontent.com/FStarLang/FStar/refs/heads/master/.scripts/get_fstar_z3.sh \
```
jschneider-bensch, line 17: Just to note: This was the case before, but we get a specific F* release, but this script is from F* `master`, and AFAICT gets a specific release of z3. Granted, that release seems not to have changed in 4 years, but there could be a point of divergence here.

jschneider-bensch (approved):
Look okay to me!
We will be in a bad CI state until https://github.com/AeneasVerif/eurydice/pull/159 get's merged, then?

.docker/c/install.sh
franziskuskiefer: Good point! Let me pin this to a specific commit.

PR #888, opened 2025-03-22
Type-based secret independence

karthikbhargavan:
This is an initial PR to include type-based secret independence in Libcrux.

The new crate `secrets` introduces types and functions for secret integers.
A future PR will provide a more comprehensive crate with secure erasure and constant-time ops.
To enable secret independence checking, it provides the feature `check-secret-independence`.
If this flag is not set, the crate implements all integers transparently, with (hopefully) no performance cost.

The secrets crate has been tested in libcrux-ml-kem.
Specifically, this PR also adapts the code within vector/portable to use secret integers.
To check secret independence, call `cargo check --features "check-secret-independence"`


secrets/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+license.workspace = true
+homepage.workspace = true
+edition.workspace = true
+repository.workspace = true
+readme.workspace = true
+description = "Libcrux utilities for working with secrets"
+exclude = ["/proofs"]
+
+[dependencies]
+hax-lib = { version = "0.2.0", git = "https://github.com/cryspen/hax/" }
```
keks, line 14: I think v0.2 of hax-lib is published - does this need to be a git dependency?

secrets/src/traits.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[cfg(target_arch = "aarch64")]
+impl Scalar for core::arch::aarch64::uint32x4_t {}
+
+#[cfg(target_arch = "aarch64")]
+impl Scalar for core::arch::aarch64::uint64x1_t {}
+
+#[cfg(target_arch = "aarch64")]
+impl Scalar for core::arch::aarch64::uint64x2_t {}
+*/
```
keks, line 113: Is this block supposed to be commented out?

secrets/src/traits.rs
```diff
@@ -0,0 +1,133 @@
+pub trait Classify {
```
keks, line 1: I think ideally the traits in this file could use some comments.

secrets/src/int/secret_integers.rs
```diff
@@ -0,0 +1,235 @@
...
+    type Declassified = T;
+    fn declassify(self) -> T {
+        self.0
+    }
+}
+
+impl<'a, T: Scalar> ClassifyRef for &'a T {
+    type ClassifiedRef = &'a Secret<T>;
+    fn classify_ref(self) -> &'a Secret<T> {
+        unsafe { core::mem::transmute(self) }
```
keks, line 38: Uses of unsafe should always have a SAFETY comment, regardless how obviously correct they are, IMHO

secrets/src/int/secret_integers.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+impl<'a, T: Scalar> DeclassifyRef for &'a Secret<T> {
+    type DeclassifiedRef = &'a T;
+    fn declassify_ref(self) -> &'a T {
+        unsafe { core::mem::transmute(self) }
+    }
+}
+
+pub fn classify_mut_slice<T: Scalar>(x: &mut [T]) -> &mut [Secret<T>] {
```
keks, line 49: Should this be a trait?

secrets/src/int/secret_integers.rs
```diff
@@ -0,0 +1,235 @@
...
+
+impl<U, T: Shl<U>> Shl<U> for Secret<T>
+where
+    T::Output: Into<T>,
+{
+    type Output = Secret<T>;
+    fn shl(self, rhs: U) -> Self::Output {
+        (self.0.shl(rhs).into()).into()
+    }
+}
```
keks, line 110: More out of curiosity than because I think that would be needed for this to merge, but would we also want to have an impl that is something like this:

```rust
impl<U, T: Shl<U>> Shl<Secret<U>> for Secret<T>
where
    T::Output: Into<T>,
```
i.e. where the shift count operant also is secret?

secrets/src/int.rs
```diff
@@ -0,0 +1,103 @@
...
+impl_cast_ops!(U8);
+impl_cast_ops!(U16);
+impl_cast_ops!(U32);
+impl_cast_ops!(U64);
+impl_cast_ops!(U128);
+impl_cast_ops!(I8);
+impl_cast_ops!(I16);
+impl_cast_ops!(I32);
+impl_cast_ops!(I64);
+impl_cast_ops!(I128);
```
keks, line 103: Not sure this is a big problem, but it does allow `1024.as_u8()`, which I guess isn't worse than before because we always can write `1024 as u8`. but still we do provide helper functions that do illegal things here. I suppose we could split this up into all the valid or invalid traits, or we leave this as is and *just don't call the bad functions*. Not sure what the best way is here.

keks (changes requested):
I think the general thing looks good, but it's still a bit rough around the edges. I am not sure everything I noted needs be addressed now, though.

secrets/src/traits.rs
karthikbhargavan: It needs additions to hax-lib. I created a PR:
https://github.com/cryspen/hax/pull/1373

secrets/Cargo.toml
karthikbhargavan: Switched to workspace

secrets/src/traits.rs
karthikbhargavan: Added some.

secrets/src/int/secret_integers.rs
karthikbhargavan: I added some comments. Two points:
(1) It should be safe because Secret has a transparent representation
(2) These unsafe functions should never be used in production because they are only enabled when "check-secret-independence" is used.

secrets/src/int/secret_integers.rs
karthikbhargavan: Added a comment: we would like hax to be able to process this function and it is easier if it were a function than a trait method.

secrets/src/int/secret_integers.rs
karthikbhargavan: The value we shift things by is explicitly not secret, because we cannot guarantee constant time behavior for it.

secrets/src/int.rs
karthikbhargavan: I was trying to exactly mimic `as` here using a function, but am open to any alternatives.

libcrux-ml-kem/src/vector/portable/ntt.rs
```diff
@@ -16,7 +17,7 @@ use super::vector_type::*;
                                                Spec.Utils.is_i16b (b+3328) ${vec}_future.f_elements.[j])) /\
                                     Spec.Utils.ntt_spec ${vec}.f_elements (v $zeta) (v $i) (v $j) ${vec}_future.f_elements"#))]
 pub(crate) fn ntt_step(vec: &mut PortableVector, zeta: i16, i: usize, j: usize) {
```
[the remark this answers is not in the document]
karthikbhargavan: Same as above.

secrets/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+mod traits;
```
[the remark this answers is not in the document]
karthikbhargavan: Added.

secrets/Cargo.toml
keks: That one also has git as a source :) But it looks like that hasn't prevented us from publishing, so maybe that's fine.

keks:
I left a bunch of suggestions that make comments doc comments, and one "outer doc comment" and "inner doc comment" (I think that was the intention there, but please double-check because I'm not 100% sure.

Otherwise looks good I think!

keks (changes requested):
Not sure what happened, let's try this again

karthikbhargavan:
Can we merge this?

libcrux-ml-kem/Cargo.toml
```diff
@@ -31,6 +31,7 @@ libcrux-platform = { version = "0.0.2", path = "../sys/platform" }
 libcrux-sha3 = { version = "0.0.2", path = "../libcrux-sha3" }
 libcrux-intrinsics = { version = "0.0.2", path = "../libcrux-intrinsics" }
 hax-lib.workspace = true
+libcrux-secrets = { version = "0.0.2-beta.3", path = "../secrets" }
```
franziskuskiefer, line 34: This version doesn't seem right? The workspace currently has `0.0.2`. But you may want to start the secrets crate off with `0.0.1-beta.1` or something like that.

libcrux-ml-kem/hax.py
```diff
@@ -74,6 +74,24 @@ def __call__(self, parser, args, values, option_string=None) -> None:
             env=hax_env,
         )
 
+        # Extract intrinsics
```
franziskuskiefer, line 77:
```suggestion
        # Extract libcrux-secrets
```

secrets/src/int/secret_integers.rs
```diff
@@ -0,0 +1,276 @@
...
+impl<'a, T: Scalar> ClassifyRef for &'a T {
+    type ClassifiedRef = &'a Secret<T>;
+    fn classify_ref(self) -> &'a Secret<T> {
+        unsafe { core::mem::transmute(self) }
+    }
+}
+
+// Declassify a reference to a scalar
+// Note: this is safe since the `Secret` type is `repr(transparent)`, so
+//       the memory represnetation of the public and secret values is the same
```
franziskuskiefer, line 51:
```suggestion
//       the memory representation of the public and secret values is the same
```

secrets/src/int/secret_integers.rs
```diff
@@ -0,0 +1,276 @@
...
+impl<T: Scalar> Declassify for Secret<T> {
+    type Declassified = T;
+    fn declassify(self) -> T {
+        self.0
+    }
+}
+
+// Classify a reference to a scalar
+// Note: this is safe since the `Secret` type is `repr(transparent)`, so
+//       the memory represnetation of the public and secret values is the same
```
franziskuskiefer, line 41:
```suggestion
//       the memory representation of the public and secret values is the same
```

secrets/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+[dependencies]
+hax-lib.workspace = true
+
+[features]
+check-secret-independence = []
+default = []
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(hax)', 'cfg(eurydice)'] }
```
franziskuskiefer, line 21: Do we actually use these configs?

secrets/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+readme.workspace = true
+description = "Libcrux utilities for working with secrets"
+exclude = ["/proofs"]
+
+[dependencies]
+hax-lib.workspace = true
+
+[features]
+check-secret-independence = []
+default = []
```
franziskuskiefer, line 18: This appears unused?

franziskuskiefer (commented):
Just a few nits

libcrux-ml-kem/Cargo.toml
karthikbhargavan: If we are merging this to libcrux-main, do we need a non-beta version?

secrets/Cargo.toml
karthikbhargavan: We do use `cfg(hax)` in our code, but maybe that isn't your question?

libcrux-ml-kem/Cargo.toml
karthikbhargavan: I switched to 0.0.2, but we can change it if needed. I don't think there is anything very controversial in this version of secrets.

franziskuskiefer (dismissed):
Clearing the review request until the C extraction issues are resolved.

karthikbhargavan:
Meeting with Jonathan to understand how to get this past Eurydice. Keen to merge soon.

karthikbhargavan:
@protz  is looking into this.

PR #896, opened 2025-04-01
chacha20poly1305: xchacha20 support

franziskuskiefer:
Adding xchacha.
This part is unverified, but uses the verified chacha-poly implementation.


keks (changes requested):
Can this mention in the docs (readme and/or rustdoc) that this is not verified?

keks:
(otherwise looks good :)

keks:
Did you mean to add the ML-KEM F* code here? :thinking:

franziskuskiefer:
😮  that merge apparently went wrong. Let me fix that.

PR #897, opened 2025-04-01
c: fixup cmake files and dependencies

franziskuskiefer:
This is currently blocking other things from being merged.

### DIFF
--- a/.github/workflows/mlkem-c.yml
+++ b/.github/workflows/mlkem-c.yml
@@ -165,7 +165,13 @@ jobs:
       - uses: actions/download-artifact@v4
         with:
           name: c-extraction
-          
+
+      - name: Set CC and CXX to Clang on Ubuntu
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          echo "CC=clang" >> $GITHUB_ENV
+          echo "CXX=clang++" >> $GITHUB_ENV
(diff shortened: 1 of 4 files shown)


keks (approved):
I think this looks good. I am not entirely sure why the library updates are needed, but since CI passes I trust that they are fine.

franziskuskiefer:
> I think this looks good. I am not entirely sure why the library updates are needed, but since CI passes I trust that they are fine.

This is only run on the merge queue.
The updates are necessary because that's what is failing (they also weren't using ranges for the minimum cmake version).

keks:
Should be related to this line in the workflow, but I am not sure what is going on in there: https://github.com/cryspen/libcrux/actions/runs/14198966199/workflow#L169-L172

The specific file seems to be some version of this: https://github.com/google/benchmark/blob/main/test/benchmark_min_time_flag_iters_test.cc

Probably imported here: https://github.com/cryspen/libcrux/blob/330abc7d3f766e0e4105058fe39e2a5a18870b7e/libcrux-ml-kem/c/CMakeLists.txt#L153-L156

Maybe bump that version as well?

franziskuskiefer:
Oh well, gcc ... I try switching to clang on CI. gcc lto isn't great.

PR #899, opened 2025-04-01
Drop F* files, modify CI accordingly

### LINKED ISSUE #898: F*: drop F* files
With PR https://github.com/cryspen/libcrux/pull/874, we now extract `libcrux-intrinsics` via hax in two different manners: one for ml-kem, another one for ml-dsa.

Having checked in F* files in the repo is incompatible with such two extraction for one given target.

Chatting with @franziskuskiefer, the only option we see here is thus to drop checked-in F* files, for now.

W95Psp:
Fixes #898


.github/workflows/mlkem-hax.yml
```diff
@@ -104,6 +87,11 @@ jobs:
       - name: ⤵ Install FStar
         run: nix profile install github:FStarLang/FStar/v2025.02.17
 
+      - uses: actions/download-artifact@v4
+        with:
+          name: fstar-extractions
+          path: .
```
wysiwys, line 93: When running this job on the merge queue, since we now rely on the `fstar-extractions` artifact here, should the `extract` job above also be run on the `merge_group` trigger, to ensure that this artifact exists and is up to date?

wysiwys (commented):
This looks good to me in general! However, there is one question about whether to regenerate the `fstar-extractions` artifact on the merge queue in `mlkem-hax.yml`.

.github/workflows/mlkem-hax.yml
W95Psp: Oh, right, absolutely, thanks that makes sense!

wysiwys (approved):
Looks good to me!

PR #906, opened 2025-04-03
ML-DSA: proofs: serialize_6

### LINKED ISSUE #887: Verify ML-DSA: avx2/encoding/commitment: serialize_6

W95Psp:
Fixes #887


W95Psp:
I spent about an hour looking at this today, I found the problem.

The serialization functions are mostly bit permutations.
Each function is essentially a sequence of compositions of intrinsics, which in this context are really just bit permutations as well.

So what I want to do here, is to flatten those composition of permutations into one permutation.

I do so by normalization in F*.
However, F* normalizes too eagerly and I end up with huge intermediate terms, causing exponential norm time in the number of intrinsics.

The strategy I will implement tomorrow is to normalize one composition at a time, bottom up. Instead of a exponential time in the number of intrinsics, my hypothesis is that we'll get a linear time.

I will spend some time tomorrow morning to implement that and will report the result here.

W95Psp:
The solution works as expected: it divides by 2 or 3 the norm time for serialize_4, and most of serialize_6 is going though now.
I'm tryining right now with everything in serialize_6, also I'm doing some clean up.

W95Psp:
The normalization is not a problem any longer, the strategy works fine.

From F* point of view, `serialize_6` produced a bad permutation however: I added new tests, so that I cover the entirerty of the intrinsics used so far, and found a stupid bug in one of my models, and another stupid bug in my Rust glue code I use for tests.

karthikbhargavan:
It may be that serialize_6 is running out of memory on the GH action. Let's debug and get this in next week.

PR #916, opened 2025-04-09
fix(F*): rename instance ids, following cryspen/hax#1391

W95Psp:
This PR updates some (hardcoded, we have no better way currently, see https://github.com/cryspen/hax/issues/1398) instance identifiers.
Those changed because of cryspen/hax#1391 which is updating the rustc pin: instance identifiers are hashes, so subtle changes to the AST make them change.

Note: th failure of hax-related jobs here is expected. I will bypass hax CI to get cryspen/hax#1391 in, then we will merge that PR.
I'm waiting on your approval @franziskuskiefer to bypass hax CI on cryspen/hax#1391, so that we can merge this PR right after.


karthikbhargavan:
Can we open an issue, if none exists, that we need to find a way to address instances in annotations, and then link that issue here?

karthikbhargavan (commented):
We need to also test this on bertie and pq11, both of which are somewhat critical right now. Let's please not merge if the proofs there break.

W95Psp:
I filed https://github.com/cryspen/hax/issues/1398 :)

franziskuskiefer (approved):

PR #917, opened 2025-04-10
Update eurydice_glue.h

msprotz:


karthikbhargavan (approved):
This looks like a non-breaking change and should work (if CI wills.)

PR #920, opened 2025-04-14
add xwing-06 support

### LINKED ISSUE #918: Add xwing-06 kem
Add xwing draft 06 and clean up other combiners in Kem.

https://datatracker.ietf.org/doc/draft-connolly-cfrg-xwing-kem/

keks:
Update to the hybrid kems.

- Drop support for XWingKyberDraft02, XWingKemDraft02, and X25519Kyber768Draft00
- Add support for [XWingKemDraft06](https://datatracker.ietf.org/doc/html/draft-connolly-cfrg-xwing-kem-06)
- Add key gen and encaps derandomized functions

Continues work from #919.

Closes #918


keks:
I gave this another review and I think the changes this PR make look good. I have opinions on the libcrux-kem crate API/architecture, but addressing these should be out of scope for this PR. Let's get this in and then iterate.

libcrux-kem/src/kem.rs
```diff
@@ -576,54 +474,101 @@ impl PublicKey {
...
+        }
+    }
+
+    /// Encapsulate a shared secret to the provided `pk` and return the `(Key, Enc)` tuple.
+    pub fn encapsulate_derand(&self, seed: &[u8]) -> Result<(Ss, Ct), Error> {
+        match self {
+            PublicKey::X25519(pk) => {
+                let new_sk = X25519PrivateKey::try_from(seed)?; // clamps
+                let new_pk =
+                    libcrux_ecdh::secret_to_public(libcrux_ecdh::Algorithm::X25519, &new_sk)?;
```
jschneider-bensch, line 502:
```suggestion
                    x25519_secret_to_public(&new_sk)?;
```

AFAICT this also means we don't need the `impl TryFrom<Vec<u8>>` for X25519 public keys.

libcrux-kem/src/kem.rs
```diff
@@ -576,54 +474,101 @@ impl PublicKey {
...
                 ))
             }
 
-            #[cfg(feature = "kyber")]
-            PublicKey::XWingKyberDraft02(XWingKemDraft02PublicKey { pk_m, pk_x }) => {
-                let seed = mlkem_rand(rng)?;
-                let (ct_m, ss_m) = kyber768::encapsulate(pk_m, seed);
-                let (ek_x, ct_x) = libcrux_ecdh::x25519_key_gen(rng)?;
+            PublicKey::XWingKemDraft06(XWingKemDraft06PublicKey { pk_m, pk_x }) => {
+                // let seed = mlkem_rand(rng)?;
```
jschneider-bensch, line 555:
```suggestion
```

libcrux-kem/src/kem.rs
```diff
@@ -914,53 +784,157 @@ pub fn key_gen(
...
             Ok((
-                PrivateKey::XWingKemDraft02(XWingKemDraft02PrivateKey {
-                    sk_m,
-                    sk_x,
-                    pk_x: X25519PublicKey(pk_x.0.clone()),
+                PrivateKey::XWingKemDraft06(XWingKemDraft06PrivateKey { seed }),
+                PublicKey::XWingKemDraft06(XWingKemDraft06PublicKey {
+                    pk_m: kp_m.pk().into(),
+                    // unwrap is ok here because it comes from the secret to pub above
+                    pk_x: pk_x.try_into().unwrap(),
```
jschneider-bensch, line 799: No need to `try_into()` here, I think.

libcrux-kem/src/kem.rs
```diff
@@ -914,53 +784,157 @@ pub fn key_gen(
...
             Ok((
-                PrivateKey::XWingKyberDraft02(XWingKemDraft02PrivateKey {
-                    sk_m,
-                    sk_x,
-                    pk_x: X25519PublicKey(pk_x.0.clone()),
+                PrivateKey::XWingKemDraft06(XWingKemDraft06PrivateKey { seed }),
+                PublicKey::XWingKemDraft06(XWingKemDraft06PublicKey {
+                    pk_m: kp_m.pk().into(),
+                    // unwrap is ok here because it comes from the secret to pub above
+                    pk_x: pk_x.try_into().unwrap(),
```
jschneider-bensch, line 873: No need to `try_into()` here.

libcrux-kem/src/kem.rs
```diff
@@ -914,53 +784,157 @@ pub fn key_gen(
...
+        ss_m: &[u8],
+        ss_x: &[u8],
+        ct_x: &[u8],
+        pk_x: &[u8],
+    ) -> XWingSharedSecret {
+        // label:
+        // \./
+        // /^\
+        // 5c2e2f2f5e5c
+        let mut input = ss_m.to_vec();
```
jschneider-bensch, line 916: All the inputs have a known length (32 bytes) so we could do this without `Vec` in a follow-up.

libcrux-ecdh/src/p256_internal.rs
```diff
@@ -32,6 +32,14 @@ impl TryFrom<&[u8]> for PublicKey {
     }
 }
 
+impl TryFrom<Vec<u8>> for PublicKey {
```
jschneider-bensch, line 35: Do we need this `impl`? Would be one more thing to guard if we at some point go full `no_std` here.

libcrux-ecdh/src/x25519.rs
```diff
@@ -31,17 +31,29 @@ impl TryFrom<&[u8]> for PublicKey {
     }
 }
 
+impl TryFrom<Vec<u8>> for PublicKey {
```
jschneider-bensch, line 34: Same question: Do we need this `impl`? I think if we use `x25519_secret_to_public` in `libcrux-kem` (see below), there's no need for it. Or did I miss something?

jschneider-bensch (dismissed):
I think it looks good!
I just have a few nits.

libcrux-kem/src/kem.rs
keks: With the current API design this is difficult to enforce, because a lot of the values come from the algorithm-agnostic enums before they are passed in, and we just call `.as_ref()` on them. I don't like this API-style personally, but I think as long as we use it, it doesn't make sense to do something clever here.

libcrux-ecdh/src/p256_internal.rs
keks: No we don't. I first deleted it, but then readded it when I realized that the one in X25519 is needed. I removed both since I now use x25519_secret_to_public (thanks for the hint)

libcrux-kem/src/kem.rs
jschneider-bensch: The length of the inputs here is part of the XWing spec (which could change, I guess), so I think we could assume them here and make it the caller's problem to check that. But yeah, that's just moving things around, when we really want a different overall structure.

jschneider-bensch (approved):

PR #922, opened 2025-04-16
Add `no_std` support to additional libcrux crates

wysiwys:
This PR makes the following crates `no_std`-compatible by default:
- `libcrux-traits`
- `libcrux-sha2`
- `libcrux-p256`

It also adds `no_std` support for the following crates (which can be activated with `default-features = false`):
- `libcrux-ml-kem`
- `libcrux-kem`
- `libcrux-ecdh`

Version numbers are also updated where needed.


keks (approved):
This looks good, thanks!

PR #924, opened 2025-04-18
Make sure Charon doesn't translate debug assertions

Nadrieril:
Charon used to disable debug assertions by default. As of https://github.com/AeneasVerif/charon/pull/660, it will stop doing that. 

I don't know whether you want the debug assertions in your generated code. Trying to compile ml-kem with debug assertions on doesn't work because of some missing function names (triggered by e.g. [this code](https://github.com/cryspen/libcrux/blob/0cc077121ef292dbc42308388552d6468b3567ce/libcrux-sha3/src/generic_keccak.rs#L102-L107)). The correct solution may be to update `eurydice-glue.h` so that this new code compiles.

In the meantime, this ensures debug assertions stay disabled after https://github.com/AeneasVerif/charon/pull/660. You may revert this if you want debug assertions in your generated code.


franziskuskiefer (approved):
Thanks for the heads up.
I don't think we want any asserts in the generated code for now.

Nadrieril:
I think you might have to re-approve?

PR #926, opened 2025-04-22
psq: drop flamegraph

franziskuskiefer:


keks (approved):
My understanding is that the file was added to the repo in error, and we are now removing it. If that is correct, let's merge this.

PR #931, opened 2025-04-23
blake2: add benchmarks

### LINKED ISSUE #838: Add Blake2 benchmarks

franziskuskiefer:
Fixes #838


jschneider-bensch (approved):
This looks good to me, thanks!

I wondered if it made sense to also measure the variable digest length variants. Looking at Blake2X, though, I think measuring only the constant digest length variants is fine, since variable length is achieved by concatenating outputs from constant length instances of Blake2.

PR #935, opened 2025-04-28
WIP: AES-GCM

karthikbhargavan:
This is a Work-in-Progress PR that adds support for AES-GCM.
It is inspired by existing work in HACL* for this primitive.
The PR contains three implementations of AES and three implementations of GF128.

AES:
* a bitsliced implementation that uses u16x8 
* an Intel implementation that uses AES-NI
* an armv8 implementation that uses Neon AES

GF128:
* a U128 implementation that eagerly reduces during multiplication (We assume that Rust/LLVM will know how to translate u128 operations to u16/u32 etc.)
* an Intel implementation that uses clmul instructions for multiplication and a separate reduction algorithm
* an Armv8 implementation that uses vmull_p64 instructions for multiplication and the same reduction algorithm as Intel

Many things remain to be done before this can be ready for a merge:
* It needs to be comprehensively tested with the NIST KATs
* We need to test the portable implementation on a small cortex M device
* The bitsliced implementation of AES is consciously trying to use minimal memory, and so relies on u16s.
   However, it can be sped up by 30% or so if we use u32s and process 2 blocks at a time
* The SIMD implementations process 1 block at a time, but could be sped up by 3X by processing 4 blocks at a time.
* The Neon implementation is slightly less efficient in order to share code with the other implementations. We should evaluate whether this is worth improving.
* We should apply the secret independence checks to this implementation
* For proofs, we need a spec for AES to link all three implementations against


franziskuskiefer:
Closing for now. We'll open another one when we get closer to merging this.

PR #936, opened 2025-04-28
nix: Fix the ml-dsa job

Nadrieril:
`hax.py` was replaced with `hax.sh`, the flake needs to be updated accordingly.


franziskuskiefer:
The dependency may go away again with #940

franziskuskiefer:
@Nadrieril do you still want to add perl? It shouldn't be needed anymore. Using the new script is still needed though.

karthikbhargavan (approved):

Nadrieril:
> @Nadrieril do you still want to add perl?

If it's no longer needed we can remove it, it's not particularly important though.

PR #937, opened 2025-04-28
ML-DSA: proofs: `montgomery_multiply`: flatten circuit

### LINKED ISSUE #778: Verify ML-DSA: portable arithmetic up to montgomery-multiply

W95Psp:
This PR makes F* normalize `montgomery_mulitply` so that it becomes a flat circuit.

This PR also adds a file `Libcrux_mld_dsa.Simd.Avx2.Arithmetic.fst`, which contains the fully expanded version of `montgomery_mulitply`, and a proof of equivalence.

Fixes https://github.com/cryspen/libcrux/issues/942.

### DIFF
--- a/libcrux-intrinsics/src/avx2.rs
+++ b/libcrux-intrinsics/src/avx2.rs
@@ -13,6 +13,7 @@ pub type Vec256Float = __m256;
 #[hax_lib::opaque]
 #[inline(always)]
 pub fn mm256_storeu_si256_u8(output: &mut [u8], vector: Vec256) {
+    #[cfg(not(hax))]
     debug_assert_eq!(output.len(), 32);
     unsafe {
         _mm256_storeu_si256(output.as_mut_ptr() as *mut Vec256, vector);
@@ -22,6 +23,7 @@ pub fn mm256_storeu_si256_u8(output: &mut [u8], vector: Vec256) {
 #[hax_lib::opaque]
 #[inline(always)]
 pub fn mm256_storeu_si256_i16(output: &mut [i16], vector: Vec256) {
+    #[cfg(not(hax))]
     debug_assert_eq!(output.len(), 16);
     unsafe {
         _mm256_storeu_si256(output.as_mut_ptr() as *mut Vec256, vector);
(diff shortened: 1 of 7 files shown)


W95Psp:
Let's see what CI says, but I think this PR is ready for review

libcrux-intrinsics/src/avx2.rs
```diff
@@ -13,6 +13,7 @@ pub type Vec256Float = __m256;
 #[hax_lib::opaque]
 #[inline(always)]
 pub fn mm256_storeu_si256_u8(output: &mut [u8], vector: Vec256) {
+    #[cfg(not(hax))]
     debug_assert_eq!(output.len(), 32);
```
karthikbhargavan, line 17: Is this because hax does not yet handle debug asserts well? If so can we point to an issue here?

libcrux-ml-dsa/src/simd/avx2/arithmetic.rs
```diff
@@ -56,6 +56,11 @@ pub(super) fn montgomery_multiply_by_constant(lhs: Vec256, constant: i32) -> Vec
     res
 }
 
+#[hax_lib::fstar::before(
```
karthikbhargavan, line 59: It would be wonderful to get to a readable annotation here.
I always disliked the tactic calls in F* and would hate to expose even more inscrutable syntax here.

karthikbhargavan (approved):
Left a couple of comments. This is a nice proof, and should serve as a model for the rest of the proofs here.

libcrux-intrinsics/src/avx2.rs
W95Psp: It's actually because hax does handle debug asserts and translate them: we end up with a lot of garbadge code that is useless for verification.

karthikbhargavan:
Hax should ideally translate this to an fstar assert so that it can be a
useful hint for verification. Asserts like this which capture the
developer's expectation or intuition are invaluable and we shoulnt
encourage discarding them. For this pr, lets leave a comment pointing to a
hax issue for better handling of debug asserts

W95Psp:
Maybe I document that on the first debug_assert?

karthikbhargavan:
yep just document this reasoning on the first debug assert.

libcrux-ml-dsa/src/simd/avx2/arithmetic.rs
W95Psp: Sure, I can add a `postprocess_with` macro

libcrux-intrinsics/src/avx2.rs
W95Psp: I added documentation to explain exactly why those are not wanted in hax

PR #940, opened 2025-04-30
ML-DSA: F*: extract on all platforms 

### LINKED ISSUE #939: ML-DSA: F*: doesn't extract on non-intel platforms
The F* extraction of ML-DSA uses plain `libcrux-intrinsics`, thus works only on x86.
We want to restore extraction on non-intel platforms.

W95Psp:
Fixes #939.

This PR makes libcrux-intrinsics use minicore at extract-time, so that we fix #939.

Without the requirement of #939, I would say that this is not the best thing to do: minicore is supposed to be a faithful model of core.
With the extraction now working against core+minicore instead of just core, we have to be extra careful about the locations of items in minicore if we still want minicore to be a model of core.


W95Psp:
The merge failed here due to some intrinsics being added I believe. I'm fixing that now

karthikbhargavan (approved):
Thanks! I tested on linux and MAC and it works.
If CI is happy, let's merge.

PR #951, opened 2025-05-05
mlkem: drop separate eurydice operations trait

franziskuskiefer:
To avoid duplicating the big trait, we can duplicate the small one. Also build the `Repr` trait only in the `hax` config.


karthikbhargavan (approved):
Ah cool, if this works with. both hax and eurydice, this is much better.

karthikbhargavan (dismissed):
Hm, spoke too soon. I can help fix the CI tomorrow and then we can merge if it works with all the tools.

franziskuskiefer:
Ah, I didn't test on arm. Will fix.

karthikbhargavan (approved):
Looks good.

karthikbhargavan:
If Eurydice is happy, this looks like a good change to make.

PR #954, opened 2025-05-09
Franziskus/mlkem mut

karthikbhargavan:
WIP: tranforming ML-KEM to use &muts everywhere.

### DIFF
--- a/libcrux-intrinsics/src/arm64.rs
+++ b/libcrux-intrinsics/src/arm64.rs
@@ -66,6 +66,11 @@ pub fn _vsubq_s16(lhs: int16x8_t, rhs: int16x8_t) -> int16x8_t {
 }
 
 #[inline(always)]
+pub fn _vnegq_s16(vec: int16x8_t) -> int16x8_t {
+    unsafe { vnegq_s16(vec) }
+}
+
+#[inline(always)]
 pub fn _vmulq_n_s16(v: int16x8_t, c: i16) -> int16x8_t {
     unsafe { vmulq_n_s16(v, c) }
 }
--- a/libcrux-intrinsics/src/arm64_extract.rs
+++ b/libcrux-intrinsics/src/arm64_extract.rs
@@ -83,6 +83,11 @@ pub fn _vsubq_s16(lhs: _int16x8_t, rhs: _int16x8_t) -> _int16x8_t {
 }
 
 #[inline(always)]
+pub fn _vnegq_s16(vec: _int16x8_t) -> _int16x8_t {
+    unimplemented!()
+}
+
+#[inline(always)]
 pub fn _vmulq_n_s16(v: _int16x8_t, c: i16) -> _int16x8_t {
     unimplemented!()
 }
--- a/libcrux-intrinsics/src/avx2.rs
+++ b/libcrux-intrinsics/src/avx2.rs
@@ -211,6 +211,12 @@ pub fn mm256_set1_epi16(constant: i16) -> Vec256 {
     unsafe { _mm256_set1_epi16(constant) }
 }
 
+#[hax_lib::opaque]
+#[inline(always)]
+pub fn mm256_set1_epi8(constant: i8) -> Vec256 {
+    unsafe { _mm256_set1_epi8(constant) }
+}
+
 #[inline(always)]
 pub fn mm256_set_epi16(
     input15: i16,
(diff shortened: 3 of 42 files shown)


karthikbhargavan:
@franziskuskiefer @jschneider-bensch 

This branch needs some fixed in order to extract F*, and then lax and typecheck it.
To test the current status run `hax.py_extract &> res` and inspect the errors in `res`.
We will be handling these errors incrementally.

In commit [29899aa](https://github.com/cryspen/libcrux/pull/954/commits/29899aa1a61f50024d8bc4ba957c2862c533cded)
I fixed hash_functions.rs to lax-check as an example, so now the first error you will see is in `ind_cpa.rs`.

The general rules for the transformation are as follows. 
For each function in a module:
* if an argument has been turned from an array (e.g. `[u8; LEN]`) to a slice (`&[u8]` or `&mut [u8]`), then we need a pre-condition stating what its length is: `#[hax_lib::requires(input.len() == LEN)]`
* if an output (e.g. `[u8; LEN]`) has been turned into a mutable slice input (`out:&mut [u8]`), then we need both a pre-condition as described above and a post-condition saying that the length did not change (unfortunate to have to do this, but needed for now): `#[hax_lib::ensures(|_| fstar!("Seq.length ${out}_future == Seq.length ${out}"))]`
* if an output has been turned into a mutable input, and if the function already has a post-condition stated in term of the result (e.g. `ensures(| result | fstar!("... result ...."))`) then we need to change each reference to result to refer to the future version of the output: `ensures(| _ | fstar!("... ${output}_future ...."))`

You can see some of these patterns in hash_functions.rs already.
Once you fix these, then `hax.py_extract` will no longer show errors for your module (even if there are errors in other modules). You can then open the extracted file in `proofs/fstar/extraction` and try to lax-check it.

Sometimes, the specs will have to be changed more substantially, in which case, comment them out, leave a note for the proofs team, and just try to make sure that the extraction works, and if possible that the output lax-checks.

karthikbhargavan:
In a follow-up commit [fa5d435](https://github.com/cryspen/libcrux/pull/954/commits/fa5d43511ac3245e66a2eaf0415d28d8248016a4) I fixed the spec for Spec.Utils.fsti to match the new signature of these functions.

Now all the post-conditions of hash-functions are also fixed for the new style.

jschneider-bensch:
Still relevant, proofs still need to be adapted.

jschneider-bensch:
In lieu of a full rebase, I locally fixed the `hax_lib` dependency version and worked through all errors given by `hax.py extract`, changing `|result| ...` closures and adding length pre- & postconditions as described above.

Now that this is done we get further into the extraction pipeline and hit large number of different hax errors.

jschneider-bensch:
After the rebase, I'm now working on fixing `DirectAndMut` issues.

However, I'm also seeing a lot of weird errors, e.g.:
```
error: [HAX0003] (DirectAndMut) The mutation of this &mut is not allowed here.
   --> libcrux-ml-kem/src/vector/portable/serialize.rs:733:5
    |
733 |       hax_lib::fstar::after(
    |  _____^
734 | |         "
...   |
754 | | "
755 | |     )
    | |_____^
    |
```
and
```
error: [HAX0013] (FStar backend) The following code snippet could not be parsed as valid F*:
` ` `

` ` `
    --> libcrux-ml-kem/src/ind_cpa.rs:1195:2
     |
1195 |   #[hax_lib::ensures(|_|
     |  __^
1196 | |     fstar!(r#"${decrypted}_future == (Spec.MLKEM.ind_cpa_decrypt_unpacked $K $ciphertext
1197 | |         (Libcrux_ml_kem.Polynomial.to_spec_vector_t #$K #$:Vector ${secret_key}.f_secret_as_ntt))
1198 | |         /\ length &{decrypted}_future == length $decrypted"#)
1199 | | )]
     | |__^
     |
```
So maybe I'm hitting a bad state in the backend somewhere.

franziskuskiefer:
> However, I'm also seeing a lot of weird errors, e.g.:

The first one is probably a bad error. @karthikbhargavan can you help out on how to rewrite the F* blobs to take into account the `&mut` style here?

The second one is a typo that I fixed in https://github.com/cryspen/libcrux/pull/954/commits/2a59630cf42cb58de95f1d7369f7276a1e6c8904.

jschneider-bensch:
Kyber APIs need fixing still.

karthikbhargavan:
I'll take a look at this later today.

karthikbhargavan:
I investigated the weird error about `DirectAndMut`. In these cases, what is happening is that the precondition (or postcondition) is referring to a function like `${Vector::montgomery_multiply_by_constant}` which used to be a pure function but is now a function that takes an `&mut` argument. This kind of function (with mutable inputs) cannot be used in this form in pre- or post-conditions, so we'd need to edit these function calls. I'll think about a good style to do these preconditions and experiment with them.

karthikbhargavan:
I took this branch further. Now there are some issues in avx2 which are "real" errors, and I am happy to guide on fixes.

karthikbhargavan:
After a bunch of edits I got this back to laxing mode for ml-kem.
I don't have the energy to try full verification right now, so will set this aside till early next week.

karthikbhargavan:
A bunch more proof/annotation fixes this weekend and it is slowly getting there.
Most of the edits are still to do with lengths of slices, but sometimes it also needs more interventions.

jschneider-bensch:
Closing this, for now.

Issue #966, opened 2025-05-17
[SHA3] Write tests for Sha3


franziskuskiefer:
To get familiar with the sha3 crate, write tests for all APIs.
The test coverage for different APIs is limited right now, and there are plenty.

There are APIs for three different platforms:
- portable
- neon (aarch64)
- avx2 (x86_64)

And APIs for
- [ ] sha224
- [ ] sha256
- [ ] sha384
- [ ] sha512
- [ ] shake128
- [ ] shake256
- [ ] General Xof
  - [ ] shake128
  - [ ] shake256
- [ ] Incremental (full blocks) for shake128/256
  - [ ] squeeze next block
  - [ ] squeeze next 3 blocks
  - [ ] squeeze next 5 blocks

franziskuskiefer:
Yes, they are doing the same thing. That's not very useful 😄 I presume they did different things at some point.
You can look at [this version](https://github.com/cryspen/libcrux/blob/62ecf781152dec1212d7b248c2a1427f26db0b48/libcrux-sha3/tests/sha3.rs) (it's on a branch only right now) for how this could look.

Parrot7483:
I only did the portable tests for now. The SIMD modules have different API's. Will come back once we need to test SIMD implementations.

franziskuskiefer:
@Parrot7483 do you intend to add the remaining tests here or shall someone else take over?

parrot7483:
I can add add the others.

franziskuskiefer:
@Parrot7483 do you plan to look at this?

PR #981, opened 2025-05-19
Add tests for hash functions

Parrot7483:
PR for #966


libcrux-sha3/tests/sha3.rs
```diff
@@ -1,3 +1,343 @@
...
+    use super::test_vectors;
+    use libcrux_sha3::portable::incremental::Xof;
+    use libcrux_sha3::portable::{incremental, sha224, sha256, sha384, sha512, shake128, shake256};
+
+    #[test]
+    fn sha3_224() {
+        let mut digest = [0u8; 28];
+
+        sha224(&mut digest, test_vectors::EMPTY);
+        assert_eq!(hex::encode(&digest), test_vectors::sha3_224::EMPTY);
```
franziskuskiefer, line 121: For all of these: `digest` is a reference already. So no need for the additional borrow.

libcrux-sha3/tests/sha3.rs
```diff
@@ -1,3 +1,343 @@
...
+    }
+
+    #[test]
+    fn sha3_shake128() {
+        let mut digest = [0u8; 42];
+
+        shake128(&mut digest, test_vectors::HELLO);
+        assert_eq!(
+            hex::encode(&digest),
+            &test_vectors::shake128::HELLO_FIVE_BLOCKS[..84]
```
franziskuskiefer, line 179: If you use constants for these numbers, you're less likely to run into issues.
Not too important here for the tests, but may still spare you some broken tests and tells the reader what's going on. Here this could be something like
```rust
const DIGEST_LEN: usize = 42;
const STRING_LEN: usize = DISTEST_LEN * 2;
```

libcrux-sha3/tests/sha3.rs
```diff
@@ -1,3 +1,343 @@
...
+        assert_eq!(
+            hex::encode(&digest),
+            &test_vectors::shake128::HELLO_FIVE_BLOCKS[..64]
+        );
+    }
+
+    #[test]
+    fn sha3_shake256_absorb() {
+        let mut state = incremental::Shake256Xof::new();
+        state.absorb_final(b"Hello, ");
```
franziskuskiefer, line 320: A test with `absorb` and an empty `absorb_final` would be good as well.
```rust
state.absorb(b"Hello, ");
state.absorb_final(&[]);
```

libcrux-sha3/tests/sha3.rs
```diff
@@ -1,3 +1,343 @@
...
+        state.squeeze(&mut digest);
+        assert_eq!(
+            hex::encode(&digest),
+            &test_vectors::shake256::HELLO_FIVE_BLOCKS[..64]
+        );
+    }
+}
+
 #[test]
 fn sha3_kat_oneshot() {
```
franziskuskiefer, line 342: As you noted already, the two tests down here do the same.
Since you added more, and more systematically, you could drop both of them.

libcrux-sha3/tests/sha3.rs
```diff
@@ -1,3 +1,343 @@
...
+                  25e643a4552d6971cf4efb277a7d1d10095";
+
+        pub const FOX_71: &str =
+            "2f671343d9b2e1604dc9dcf0753e5fe15c7c64a0d283cbbf722d411a0e36f6ca1\
+                  d01d1369a23539cd80f7c054b6e5daf9c962cad5b8ed5bd11998b40d5734442bed798f6e5c915";
+    }
+}
+
+// Portable implementation tests
+mod portable {
```
franziskuskiefer, line 111: What's the plan for testing the simd variants? Do you want to do a separate PR for them?

libcrux-sha3/src/lib.rs
```diff
@@ -355,6 +355,12 @@ pub mod portable {
             absorb_final::<1, u64, 168, 0x1fu8>(&mut s.state, &[data0], 0, data0.len());
         }
 
+        /// Squeeze first block
+        #[inline(always)]
+        pub fn shake128_squeeze_first_block(s: &mut KeccakState, out0: &mut [u8]) {
```
franziskuskiefer, line 360: Is this only for testing or did you want to make things symmetrical with shake256?

Some background: These very specific APIs only exist for use in ML-KEM and ML-DSA because they have a very specific usage pattern of SHA3, which allows them to be faster. The shake256 variant of this function is needed there. The shake128 is not. Hence only the shake256 variant exists.

The same is true for the functions that you added for shake256.

Note that changing functions here will also require changes to the extracted C code (this will only fail on the merge queue when the C code is tested). See https://github.com/cryspen/libcrux/blob/main/libcrux-ml-kem/TOOLS.md for the C extraction, or use the docker image (for example `docker run -v $PWD/:/home/user/libcrux --rm ghcr.io/cryspen/libcrux-c:latest bash -c "cd libcrux/libcrux-ml-kem && ./boring.sh"`).

franziskuskiefer (changes requested):
Thanks, looks pretty good already. See comments inline.

libcrux-sha3/tests/sha3.rs
Parrot7483: You can do this bug it will give you a warning: Cannot multiply `usize` by `usize` [[E0369](https://doc.rust-lang.org/error_codes/E0369.html)]

libcrux-sha3/src/lib.rs
Parrot7483: Understand. I will remove these then. 

Yes, I wanted both to be symmetrical.

libcrux-sha3/tests/sha3.rs
franziskuskiefer: There's a typo in my code. But otherwise that works fine.

Parrot7483:
I always press that bright green button on github and mess up my git 🙃

PR #982, opened 2025-05-19
Add CI workflow for building & testing PSQ

jschneider-bensch:


.github/workflows/psq-build-test.yml
```diff
@@ -0,0 +1,120 @@
...
+          - macos-13 # Intel mac
+          - macos-latest # macos-14 m1
+          - ubuntu-latest
+          - windows-latest
+        exclude:
+          - bits: 32
+            os: "macos-latest"
+          - bits: 32
+            os: "macos-13"
+          - bits: 32 # FIXME: Linking isn't working here yet for hacl #42
```
franziskuskiefer, line 35: We could drop this? This is Rust only and should work?

.github/workflows/psq-build-test.yml
```diff
@@ -0,0 +1,120 @@
...
+      - name: 🏃🏻 Asan MacOS
+        if: ${{ matrix.os == 'macos-latest' }}
+        run: RUSTDOCFLAGS=-Zsanitizer=address RUSTFLAGS=-Zsanitizer=address cargo +nightly-2024-12-14 test --release --target aarch64-apple-darwin
+
+
+      # We get false positives here.
+      # TODO: Figure out what is going on here
+      # - name: 🏃🏻 Asan Linux
+      #   if: ${{ matrix.bits == 64 && matrix.os == 'ubuntu-latest' }}
+      #   run: RUSTDOCFLAGS=-Zsanitizer=address RUSTFLAGS=-Zsanitizer=address cargo +nightly test --release --target x86_64-unknown-linux-gnu
```
franziskuskiefer, line 96:
```suggestion
```

.github/workflows/psq-build-test.yml
```diff
@@ -0,0 +1,120 @@
+name: PSQ - Build & Test
+
+on:
+  merge_group:
```
franziskuskiefer, line 4: We should clean up all the jobs. But running everything on the merge queue and main again is a bit of a waste. We shouldn't do the same for new workflows.

.github/workflows/psq-build-test.yml
jschneider-bensch: Indeed, yes. I was working off the ECDH workflow and this is also in there still.

.github/workflows/psq-build-test.yml
jschneider-bensch: Right, okay to just run this on PR and dispatch?

Issue #945, opened 2025-04-28
ML-DSA: verify avx2 `ntt`


W95Psp:
- [x] prepare core-models so that it typechecks (https://github.com/cryspen-ext/core-models/pull/6)
 - [ ] flatten `butterfly_2`
 - [ ] flatten `butterfly_4`
 - [ ] flatten `butterfly_8`
 - [ ] write a spec  for the "two-points-wise operation" (mongomery_mul > sub > add)
 - [ ] Test SMT approach for `ntt_at_layer_0`:
    - [ ] spec for each butterfly function "I change index i and j to value X and Y, everything else remains the same"
    - [ ] write spec for ntt_at_layer_0
    - [ ] ask SMT
 - [ ] If SMT approach doesn't work, then
    - [ ] lift seq to FunArrays
    - [ ] normalize ntt_at_layer_0 into a circuit of two-levels nested matches of size 8

karthikbhargavan:
This is essentially done. Will open new issues for any leftover work.

PR #987, opened 2025-05-21
ml-dsa: avx2: intro. `montgomery_multiply_aux`

W95Psp:
This commit splits the function `montgomery_multiply`, introducing `montgomery_multiply_aux`. That latter function takes explicitely `field_modulus` and `inverse_of_modulus_mod_montgomery_r`.

This is useful in `ntt.rs` where those two are precomputed.

This commit also uses `montgomery_multiply_aux` instead of redefining the same logic in `ntt.rs`.

This is a step towards the proof of AVX2 NTT, documented in the issue #945


franziskuskiefer (approved):
Thanks.
Please reference the issue this is working towards in the description before merging.

karthikbhargavan:
This helps in the proof of NTT and hence partially addresses https://github.com/cryspen/libcrux/issues/945

Issue #944, opened 2025-04-28
ML-DSA: verify portable `ntt`


W95Psp:

W95Psp:
@karthikbhargavan and I inverstigated a bit this.

Notes:
 - the indirection via the `values` field of `Coefficients` will probably makes us hit https://github.com/FStarLang/FStar/issues/3800 to a certain degree. Would be much nicer to have IndexMut on coefficients. (see https://github.com/cryspen/hax/issues/644#issuecomment-2090950211)
 - we should add an interpretations for Rust arrays (or F* sequences) as lists, defining lifts and simplification lemmas just as we did for bitvec<->int_vec
 - then we can see e.g. ntt_at_layer_0 as a circuit
 - it is less clear for `outer_3_plus`: there is a loop, and we need `montgomery_multiply_by_constant`, subtract and add to be circuits already

karthikbhargavan:
Can likely be improved, but addressed for now in #989

PR #989, opened 2025-05-21
Range Proofs and Proof Speedups for ML-DSA (Portable) NTT

karthikbhargavan:
Addresses #944

### DIFF
--- a/libcrux-ml-dsa/src/simd/portable/arithmetic.rs
+++ b/libcrux-ml-dsa/src/simd/portable/arithmetic.rs
@@ -21,7 +21,6 @@ pub fn add(lhs: &mut Coefficients, rhs: &Coefficients) {
     let _lhs0 = lhs.clone();
     hax_lib::fstar!("reveal_opaque (`%$add_pre) ($add_pre)");
     hax_lib::fstar!("reveal_opaque (`%$add_post) ($add_post)");
-        
 
     for i in 0..lhs.values.len() {
         hax_lib::loop_invariant!(|i: usize| {
@@ -99,7 +98,7 @@ pub(crate) fn get_n_least_significant_bits(n: u8, value: u64) -> u64 {
 #[hax_lib::requires(fstar!(r#"Spec.Utils.is_i64b (8380416 * pow2 32) value "#))]
 #[hax_lib::ensures(|result| fstar!(r#"Spec.Utils.is_i32b (8380416 + 4190209) result /\
                 (Spec.Utils.is_i64b (8380416 * pow2 31) value ==> Spec.Utils.is_i32b 8380416 result) /\
-                v result % 8380417 == (v value * 8265825) % 8380417"#))]
+                Spec.MLDSA.Math.(mod_q (v result) == mod_q (v value * 8265825))"#))]
 pub(crate) fn montgomery_reduce_element(value: i64) -> FieldElementTimesMontgomeryR {
     let t = get_n_least_significant_bits(MONTGOMERY_SHIFT, value as u64)
         * INVERSE_OF_MODULUS_MOD_MONTGOMERY_R;
@@ -191,7 +190,7 @@ pub(crate) fn montgomery_reduce_element(value: i64) -> FieldElementTimesMontgome
             (v $value * 8265825) % 8380417;
         }"#
     );
-
+    hax_lib::fstar!(r#"reveal_opaque (`%Spec.MLDSA.Math.mod_q) (Spec.MLDSA.Math.mod_q)"#);
     res
 }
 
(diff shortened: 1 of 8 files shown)


libcrux-ml-dsa/src/simd/portable/invntt.rs
```diff
@@ -200,25 +315,80 @@ fn invert_ntt_at_layer_2(re: &mut [Coefficients; SIMD_UNITS_IN_RING_ELEMENT]) {
...
+    Libcrux_ml_dsa.Simd.Portable.Ntt.is_i32b_polynomial (8 * v $FIELD_MAX) ${re}
+"#))]
+#[hax_lib::ensures(|_| fstar!(r#"
+    Libcrux_ml_dsa.Simd.Portable.Ntt.is_i32b_polynomial (16 * v $FIELD_MAX) ${re}_future
+"#) )]
 fn invert_ntt_at_layer_3(re: &mut [Coefficients; SIMD_UNITS_IN_RING_ELEMENT]) {
     const STEP: usize = 8; // 1 << LAYER;
     const STEP_BY: usize = 1; // step / COEFFICIENTS_IN_SIMD_UNIT;
 
+    // std::println!("layer 3:{}", re[0].values[0].abs());
```
franziskuskiefer, line 391: I stop doing this, this may still be cleanup. Just note that there's more.

```suggestion
```

W95Psp (approved):
The F* looks good to me!

libcrux-ml-dsa/src/simd/traits.rs
```diff
@@ -30,6 +30,16 @@ pub(crate) trait Repr: Copy + Clone {
 pub(crate) mod specs {
```
franziskuskiefer, line 30: Thanks for moving all the proof code in here. Since this is huge by now, let's move it into a separate file to keep things a little cleaner.

libcrux-ml-dsa/src/simd/traits.rs
```diff
@@ -30,6 +30,16 @@ pub(crate) trait Repr: Copy + Clone {
...
+    pub(crate) const MONT_R: u32 = 8380417;
+
+    pub(crate) const FIELD_MAX: u32 = 8380416;
+
+    pub(crate) const FIELD_MID: u32 = 4190208;
+
+    pub(crate) const NTT_BASE_BOUND: u32 = FIELD_MID;
+
     type SIMDContent = [i32; COEFFICIENTS_IN_SIMD_UNIT];
     // Avoiding a recursive bundle
```
franziskuskiefer, line 44: This sounds like it should get a hax bug for improving this? Duplicating stuff like this just because it's in a nested module isn't great.

libcrux-ml-dsa/src/simd/portable/ntt.rs
```diff
@@ -201,20 +324,81 @@ fn ntt_at_layer_2(re: &mut [Coefficients; SIMD_UNITS_IN_RING_ELEMENT]) {
...
+                (Seq.index ${re}_future i).f_values))
+"#))]
 fn outer_3_plus<const OFFSET: usize, const STEP_BY: usize, const ZETA: i32>(
     re: &mut [Coefficients; SIMD_UNITS_IN_RING_ELEMENT],
 ) {
+    #[cfg(hax)]
+    let orig_re = re.clone();
+
     for j in OFFSET..OFFSET + STEP_BY {
+        hax_lib::loop_invariant!(|j: usize| fstar!(
```
franziskuskiefer, line 365: It looks like ml-dsa is lacking a `unused_qualifications` warning/deny. Can you add one to `lib.rs`. The qualification here is unused. You should either drop the `use` statement for it, or the `hax_lib` here.

libcrux-ml-dsa/src/simd/portable/ntt.rs
```diff
@@ -1,73 +1,144 @@
...
 use super::vector_type::Coefficients;
 use crate::simd::traits::{COEFFICIENTS_IN_SIMD_UNIT, SIMD_UNITS_IN_RING_ELEMENT};
 
+#[cfg(hax)]
+use crate::simd::traits::specs::*;
+#[cfg(hax)]
+pub(crate) const NTT_BASE_BOUND: u32 = FIELD_MID;
+
+#[inline(always)]
+#[hax_lib::fstar::options("--z3rlimit 300 --split_queries always")]
```
franziskuskiefer, line 13: Is there a reason not to import `use hax_lib::{*, fstar::*}`?

libcrux-ml-dsa/src/simd/portable/ntt.rs
```diff
@@ -76,6 +147,9 @@ fn ntt_at_layer_0(re: &mut [Coefficients; SIMD_UNITS_IN_RING_ELEMENT]) {
         zeta_2: i32,
         zeta_3: i32,
     ) {
+        hax_lib::fstar!(
+            "reveal_opaque (`%Spec.Utils.is_i32b_array_opaque) (Spec.Utils.is_i32b_array_opaque)"
+        );
```
franziskuskiefer, line 152: Let's make sure there's some separation between proofs and coder everywhere.

```suggestion
        );

```

libcrux-ml-dsa/src/simd/portable/arithmetic.rs
```diff
@@ -350,10 +369,12 @@ pub(super) fn power2round(t0: &mut Coefficients, t1: &mut Coefficients) {
 // TODO: Revisit this function when doing the range analysis and testing
```
franziskuskiefer, line 369: Is this still an open todo?

franziskuskiefer (approved):
Thanks, generally looks good with a couple nits.

libcrux-ml-dsa/src/simd/traits.rs
karthikbhargavan: The big zeta functions will likely move to the Spec module, so let's leave that for a future PR.
I would prefer the pre and post functions to remain close to the trait, and they are not very big.

libcrux-ml-dsa/src/simd/traits.rs
karthikbhargavan: This bit is not a hax bug.  Rust allows circular dependencies between modules but F* does not and we have to deal with it in some way.  hax would correctly inject a recursive bundle here, and I prefer not to have recursive bundles for proof ergonomics, and so we try to avoid trivial circular dependencies in our own code wherever possible, especially when writing specs.

libcrux-ml-dsa/src/simd/portable/ntt.rs
karthikbhargavan: fixed, thanks.

libcrux-ml-dsa/src/simd/portable/ntt.rs
karthikbhargavan: Done

libcrux-ml-dsa/src/simd/portable/arithmetic.rs
karthikbhargavan: Still todo until we analyze all call-sites

Issue #373, opened 2024-07-08
Verify ML-DSA


franziskuskiefer:

karthikbhargavan:
The plan for the ML-DSA proofs are as follows:

1.  Prove the arithmetic code (by porting them from ML-KEM where possible)
2. Prove the serialization code (by adapting and reusing the tactic from ML-KEM)
3. Prove the portable NTT code
4. Prove the optimized AVX2 NTT code
5. Set the pre- and post-conditions for the SIMD trait
6. Prove the generic algorithms

W95Psp:
To verify `simd::avx2::encoding`, we need to model the following intrinsics:
 - [ ] mm256_abs_epi32
 - [ ] mm256_add_epi32
 - [ ] mm256_add_epi64
 - [ ] mm256_and_si256
 - [ ] mm256_blend_epi32
 - [ ] mm256_bsrli_epi128
 - [ ] mm256_castsi256_ps
 - [x] mm256_castsi256_si128
 - [ ] mm256_cmpeq_epi32
 - [ ] mm256_cmpgt_epi32
 - [x] mm256_extracti128_si256
 - [ ] mm256_loadu_si256_i32
 - [ ] mm256_loadu_si256_u8
 - [ ] mm256_madd_epi16
 - [ ] mm256_movemask_ps
 - [ ] mm256_mul_epi32
 - [x] mm256_mullo_epi16_shifts
 - [x] mm256_mullo_epi32
 - [ ] mm256_or_si256
 - [ ] mm256_permute2x128_si256
 - [x] mm256_permutevar8x32_epi32
 - [ ] mm256_set1_epi32
 - [x] mm256_set_epi16
 - [x] mm256_set_epi32
 - [ ] mm256_set_epi64x
 - [x] mm256_set_epi8
 - [ ] mm256_set_m128i
 - [ ] mm256_setzero_si256
 - [ ] mm256_shuffle_epi32
 - [x] mm256_shuffle_epi8
 - [ ] mm256_sign_epi32
 - [ ] mm256_slli_epi32
 - [ ] mm256_slli_epi64
 - [x] mm256_sllv_epi32
 - [ ] mm256_srai_epi32
 - [x] mm256_srli_epi64
 - [x] mm256_srlv_epi32
 - [ ] mm256_srlv_epi64
 - [ ] mm256_storeu_si256_i32
 - [ ] mm256_sub_epi32
 - [ ] mm256_testz_si256
[...]

PR #990, opened 2025-05-21
mldsa: hide hax clones a little better

franziskuskiefer:
A small quality of life improvement for cloning mutable input variables for hax.

#373


karthikbhargavan:
Nice idea. Would be nice to have something that is acceptable and evocative to Rust developers.
And we should put this into hax_lib.
Maybe `hax_lib::ghost(x = y)`?

franziskuskiefer:
> And we should put this into hax_lib.
> Maybe `hax_lib::ghost(x = y)`?

yeah I agree. this would be better placed in hax itself. `ghost` sounds good.
I just wanted something quick to see how it looks. It's not ideal, but better already because only one line and clearly marked as a hax thing.

franziskuskiefer:
Closing in favour of https://github.com/cryspen/hax/issues/1167

PR #996, opened 2025-05-23
sha2/ed25519: Minor API improvements

franziskuskiefer:
A few minor API improvements on a sha2 and ed25519.

@keks I don't know if you want to get them into the releases or not. I'm fine either way.

cc cryspen/home#287

### DIFF
--- a/sha2/src/hacl.rs
+++ b/sha2/src/hacl.rs
@@ -3,6 +3,7 @@
 #![allow(non_camel_case_types)]
 #![allow(unused_assignments)]
 #![allow(unreachable_patterns)]
+#![allow(dead_code)]
 
 //! This module contains generated hacl code.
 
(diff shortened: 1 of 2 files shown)


sha2/src/hacl.rs
```diff
@@ -3,6 +3,7 @@
 #![allow(non_camel_case_types)]
 #![allow(unused_assignments)]
 #![allow(unreachable_patterns)]
+#![allow(dead_code)]
```
jschneider-bensch, line 6: This is about some unused type aliases, right? Should we annotate them explicitly, rather than using this general `allow`?

jschneider-bensch (commented):
Looks fine to me

sha2/src/hacl.rs
franziskuskiefer: It's generated code. So I'd prefer just have it here on top. When we touch more code, we can do that (or actually clean it up).

jschneider-bensch (approved):

Issue #964, opened 2025-05-15
[CI] Update benchmarks to `benchmark-upload-and-plot-action@v3`


wysiwys:

franziskuskiefer:
- #946 
- #956

PR #1003, opened 2025-05-26
Upgrade benchmarking actions in ML-KEM and ML-DSA workflows

wysiwys:
For the ML-KEM and ML-DSA benchmarks, this PR updates the workflow config and benchmark names as required to upgrade the action versions (see #964).

Before merging this PR, the old benchmark results should be combined and placed in the new location in the `gh-pages` branch.


franziskuskiefer (approved):
lgtm

PR #1006, opened 2025-05-27
Ensure iNTT inputs are reduced

jschneider-bensch:


franziskuskiefer:
lgtm, can you add some description for future selfs?
I don't see a performance impact, so that's good

jschneider-bensch:
Need to update C extraction, still

PR #1012, opened 2025-06-05
[ML-KEM] Provide idiomatic Rust API & PQCP common API

jschneider-bensch:
This PR attempts to provide an idiomatic Rust API for ML-KEM, as well as the PQCP common APIs.

- [x] introduce default feature `c-apis` for old APIs
  - [x] Add deprecation warning to old APIs
- [ ] Implement idiomatic Rust API for packed key generation, encapsulation, decapsulation. These will always perform key validation before encapsulation and decapsulation.
  - [x] ML-KEM-512
  - [ ] ML-KEM 768
  - [ ] ML-KEM 1024
- [ ] Implement idiomatic Rust API for unpacked key generation, encapsulation, decapsulation. These will perform key validation once on deserialization and then perform encapsulation and decapsulation without further validation.
  - [x] ML-KEM-512
  - [ ] ML-KEM 768
  - [ ] ML-KEM 1024  
- [ ] Provide PQCP API on top of idiomatic API
  - [x] ML-KEM-512 **???**  I am not sure about `crypto_kem_sk_from_seed` and `crypto_kem_pk_from_sk`, specifically whether they should work for unpacked or packed keys.
  - [ ] ML-KEM 768
  - [ ] ML-KEM 1024
- [ ] Update all doc comments
- [ ] Update examples, benchmarks and tests to use new APIs

### DIFF
--- a/libcrux-ml-kem/tests/self.rs
+++ b/libcrux-ml-kem/tests/self.rs
@@ -36,6 +36,30 @@ macro_rules! impl_consistency {
     };
 }
 
+macro_rules! impl_consistency_new {
+    ($name:ident, $key_gen:expr, $encaps:expr, $decaps:expr) => {
+        #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test::wasm_bindgen_test)]
+        #[test]
+        fn $name() {
+            let randomness = random_array();
+            let key_pair = $key_gen(randomness);
+            let randomness = random_array();
+            let (ciphertext, shared_secret) =
+                key_pair.public_key().encapsulate(randomness).unwrap();
+            let shared_secret_decapsulated =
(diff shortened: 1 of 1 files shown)


dconnolly:
This is lovely to see 🎉

keks:
One note and some questions:

From the C implementation linked in the issue I understand that the two unclear methods should operate on unpacked keys.

What I don't understand is how "rusty" we want the PQCP API to be. Should we return values or write into an &mut? Should we use array references like the C API does or use strong types? Do we plan to extract this to C as well? Is the goal that we exactly get the PQCP API if we extract to C?

keks:
One more open question: It seems the unpacked structs are specific to the backend, but the PCQP needs to commit to the types. Which one should we use?

jschneider-bensch:
Closing this, as PQCP APIs were added with #1165.

PR #1013, opened 2025-06-07
Sha3 extract

karthikbhargavan:


franziskuskiefer:
> If I understand correctly, the main changes here are to switch from:
> 
> * generic store_block
> * generic squeeze
> * generic keccak
>   towards a style (necessary for now) where for each of the above 3, we have three versions, for 1, 2, and 4 outputs.
> 
> This is needed primarily because of the limitations of hax in handling a polymorphic number of mutable buffers. We should merge back these versions when hax can handle a better style.

yes. and I moved them into separate modules to keep things somewhat nice.
F* extracts for me with this, and I added a CI job for it (not pushed yet, blocked by #1022).

I also added traits for load and store as these are not functions on the item elements, but the state/bytes.

libcrux-sha3/examples/sha3.rs
```diff
@@ -0,0 +1,73 @@
...
+                hex::encode(digest)
+            }
+            "shake256" => {
+                let mut digest = vec![0u8; args.out_len];
+                shake256_ema(&mut digest, data);
+                hex::encode(digest)
+            }
+            unknown_variant => {
+                eprintln!(
+                "Unsupported or unknown variant: \"{}\".\nSupported variants are: sha3-224, sha3-256, sha3-384, sha3-512, shake128, shake256.\n",
```
jschneider-bensch:
```suggestion
                "Unsupported or unknown variant: \"{}\".\nSupported variants are: sha224, sha256, sha384, sha512, shake128, shake256.\n",
```
I think these should match the discriminants of the `match`.

libcrux-sha3/src/generic_keccak.rs
```diff
@@ -1,564 +1,221 @@
...
-#[cfg_attr(hax, hax_lib::opaque)]
-pub(crate) struct KeccakXofState<
-    const PARALLEL_LANES: usize,
-    const RATE: usize,
-    STATE: KeccakStateItem<PARALLEL_LANES>,
-> {
-    inner: KeccakState<PARALLEL_LANES, STATE>,
+/// Simd128 specific implementations.
+///
+/// We need to work around hax limitations and therefore need to implement `squeeze`
```
jschneider-bensch, line 17: Is there a specific issue that could be linked here?

libcrux-sha3/src/generic_keccak/simd128.rs
```diff
@@ -0,0 +1,92 @@
...
+) {
+    let mut s = KeccakState::<2, _uint64x2_t>::new();
+    let data_len = data[0].len();
+    for i in 0..data_len / RATE {
+        s.absorb_block::<RATE>(data, i * RATE);
+    }
+    let rem = data_len % RATE;
+    s.absorb_final::<RATE, DELIM>(data, data_len - rem, rem);
+
+    debug_assert!(out0.len() == out1.len());
```
jschneider-bensch: This should go to the beginning of the function.

Similarly, it should be asserted that `data[0].len() == data[1].len()`.

libcrux-sha3/src/avx2.rs
jschneider-bensch, line 1: To keep things consistent, should this module be named `simd256`? Same question for `neon`.

keks:
This has requested reviews, but the first sentence is that it's not ready for review. Which one is it? :)

karthikbhargavan:
Edited the description, which was out of date.

keks:
Alright, I'll take a look today!

.github/workflows/mlkem-c.yml
keks, line 1: Why are we disabling C ML-KEM extraction here? And if that is really needed, should we maybe remove the lines instead of commenting them out? Also why just PR, not push?

libcrux-sha3/benches/sha3.rs
```diff
@@ -33,10 +33,11 @@ macro_rules! impl_comp {
                     BenchmarkId::new("portable", fmt(*payload_size)),
                     payload_size,
                     |b, payload_size| {
-                        b.iter_batched(
+                        b.iter_batched_ref(
                             || randombytes(*payload_size),
                             |payload| {
-                                let _d: [u8; digest_size($libcrux)] = hash($libcrux, &payload);
+                                let _d: [u8; digest_size($libcrux)] =
+                                    core::hint::black_box(hash($libcrux, payload));
```
keks, line 40: You can just let the closure return the call to `hash`, criterion adds the black box itself: https://docs.rs/criterion/0.6.0/src/criterion/bencher.rs.html#335-343

libcrux-sha3/benches/sha3.rs
```diff
@@ -48,11 +49,11 @@ macro_rules! impl_comp {
                     BenchmarkId::new("simd128", fmt(*payload_size)),
                     payload_size,
                     |b, payload_size| {
-                        b.iter_batched(
+                        b.iter_batched_ref(
                             || randombytes(*payload_size),
                             |payload| {
                                 let mut digest = [0u8; digest_size($libcrux)];
-                                neon::$rust_fun(&mut digest, &payload);
+                                core::hint::black_box(neon::$rust_fun(&mut digest, payload));
```
keks, line 56: Same here

libcrux-sha3/benches/sha3.rs
```diff
@@ -87,23 +88,23 @@ fn shake256(c: &mut Criterion) {
...
+                    payload0,
+                    payload1,
+                    payload2,
+                    payload3,
                     &mut digest0,
                     &mut digest1,
                     &mut digest2,
                     &mut digest3,
-                );
+                ));
```
keks, line 100: return value instead of black box, see above.

libcrux-sha3/src/generic_keccak.rs
```diff
@@ -1,564 +1,221 @@
...
-    // Buffer inputs on absorb.
-    buf: [[u8; RATE]; PARALLEL_LANES],
+/// Simd256 specific implementations.
+#[cfg(feature = "simd256")]
+pub(crate) mod simd256;
 
-    // Buffered length.
-    buf_len: usize,
+/// Portable specific implementations.
+pub(crate) mod portable;
```
keks, line 27: Maybe the comments can be updated that these still are generic implementations, but specific w.r.t. N

libcrux-sha3/src/traits.rs
```diff
@@ -24,25 +15,71 @@ pub(crate) fn set_ij<const N: usize, T: KeccakStateItem<N>>(
...
+    fn squeeze4<const RATE: usize>(
+        &self,
+        out0: &mut [u8],
+        out1: &mut [u8],
+        out2: &mut [u8],
+        out3: &mut [u8],
+        start: usize,
+        len: usize,
+    );
 }
```
keks: I think the "only implement one and panic when a different one is called" is a bit unfortunate. Is there a reason these aren't three fully separate traits and each specific one just implements the one that t supports?

keks: I suppose alternatively we could have `out: [&mut [u8]; N]`, but that's not pretty and I am not sure how hax likes that

libcrux-sha3/src/generic_keccak/simd128.rs
```diff
@@ -0,0 +1,92 @@
...
+        out0: &mut [u8],
+        out1: &mut [u8],
+        start: usize,
+    ) {
+        self.keccakf1600();
+        self.squeeze2::<RATE>(out0, out1, start, RATE);
+    }
+
+    #[inline(always)]
+    pub(crate) fn squeeze_first_block<const RATE: usize>(&self, out0: &mut [u8], out1: &mut [u8]) {
```
jschneider-bensch, line 58:
```suggestion
    /// Write out the first block of Keccak output.
    ///
    /// This function MUST NOT be called after any of the other `squeeze_*` functions have been called, since that would result in a duplicate output block.
    pub(crate) fn squeeze_first_block<const RATE: usize>(&self, out0: &mut [u8], out1: &mut [u8]) {
```

It should be documented that the `squeeze_first_*` functions must not be called after any of the other `squeeze` functions, since that would write out an old Keccak state. The names are a hint, but since we don't enforce it otherwise, I think we should be more explicit in documentation.

libcrux-sha3/src/generic_keccak/simd256.rs
```diff
@@ -0,0 +1,106 @@
...
+) {
+    let mut s = KeccakState::<4, Vec256>::new();
+    let data_len = data[0].len();
+    for i in 0..data_len / RATE {
+        s.absorb_block::<RATE>(data, i * RATE);
+    }
+    let rem = data_len % RATE;
+    s.absorb_final::<RATE, DELIM>(data, data_len - rem, rem);
+
+    debug_assert!(out0.len() == out1.len());
```
jschneider-bensch: See above, the `debug_assert` should go to the beginning of the function. Further, it currently doesn't cover `out2`,`out3` and `data` lengths.

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -0,0 +1,220 @@
...
+            buf: [Self::zero_block(); PARALLEL_LANES],
+            buf_len: 0,
+            sponge: false,
+        }
+    }
+
+    /// Absorb
+    ///
+    /// This function takes any number of bytes to absorb and buffers if it's not enough.
+    /// The function assumes that all input slices in `blocks` have the same length.
```
jschneider-bensch:
```suggestion
    /// The function assumes that all input slices in `inputs` have the same length.
```

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -0,0 +1,220 @@
...
+        let input_consumed = self.fill_buffer(inputs);
+
+        if input_consumed > 0 {
+            let mut borrowed = [[0u8; RATE].as_slice(); PARALLEL_LANES];
+            // We have a full block in the local buffer now.
+            #[allow(clippy::needless_range_loop)]
+            for i in 0..PARALLEL_LANES {
+                borrowed[i] = &self.buf[i];
+            }
+
```
jschneider-bensch, line 102:
```suggestion
            let borrowed: [&[u8]; PARALLEL_LANES] =
                core::array::from_fn(|lane| &self.buf[lane][..]);
```
Doing it like this could avoid stating one more loop invariant down the line.

libcrux-sha3/src/simd/portable.rs
```diff
@@ -0,0 +1,167 @@
...
+
+#[inline(always)]
+pub(crate) fn load_block<const RATE: usize>(state: &mut [u64; 25], blocks: &[u8], start: usize) {
+    debug_assert!(start + RATE <= blocks.len() && RATE % 8 == 0);
+
+    // First load the block, then xor it with the state
+    // Note: combining the two loops below reduces performance for large inputs,
+    //       so we knowingly use two loops: one for loading, one for xor
+    let mut state_flat = [0u64; 25];
+
```
jschneider-bensch, line 44: If it doesn't impact perf, this initialization could also be done with `core::array::from_fn` to save some loop invariant lines on verification.

libcrux-sha3/src/neon.rs
```diff
@@ -0,0 +1,137 @@
+use crate::generic_keccak::simd128::keccak2;
+
+/// A portable SHA3 224 implementation.
```
jschneider-bensch: These doc comments are out of date.

libcrux-sha3/src/portable.rs
```diff
@@ -0,0 +1,187 @@
...
+/// A portable SHA3 224 implementation.
+#[inline(always)]
+pub fn sha224(digest: &mut [u8], data: &[u8]) {
+    keccak1::<144, 0x06u8>(data, digest);
+}
+
+/// A portable SHA3 256 implementation.
+#[inline(always)]
+pub fn sha256(digest: &mut [u8], data: &[u8]) {
+    keccak1::<136, 0x06u8>(data, digest);
```
jschneider-bensch, line 20: It would be nice to have constants for all the rates and delimiters. This would make it harder to use the wrong constant in some place and/or make it easier to spot.

.github/workflows/mlkem-c.yml
franziskuskiefer: This code is dead and will be removed. But replaced with a different version of the C code. There's no need to re-run things on push since it's run on the merge queu.

libcrux-sha3/benches/sha3.rs
franziskuskiefer: I know, but that doesn't really do anything. Criterion can't really handle these things. So this is just as bad as criterion without. We'll have to rewrite them at some point.

libcrux-sha3/src/generic_keccak.rs
franziskuskiefer: Support for mutability. But there's no single issue for this as far as I can tell.

libcrux-sha3/src/generic_keccak/xof.rs
franziskuskiefer: This version requires a bounds check because of the `lane` indexing. The other one doesn't. I don't know if it makes a (big) performance difference. But the current version generates way nicer asm. https://godbolt.org/z/oTxjM1eMd

libcrux-sha3/src/portable.rs
franziskuskiefer: Sure, can you file a followup?

libcrux-sha3/src/simd/portable.rs
franziskuskiefer: Similar to the other one will this probably introduce range checks. If you can find a version with `from_fn` that doesn't, we can change it.

libcrux-sha3/src/traits.rs
franziskuskiefer: This is the entire point of this PR. We need to get rid of generically sized outputs because we don't have mutability support in hax.

libcrux-sha3/src/generic_keccak.rs
franziskuskiefer: I thought having the module called generic is enough?

libcrux-sha3/src/traits.rs
franziskuskiefer: But I split it into 3 for you ;)

franziskuskiefer (approved):

libcrux-sha3/src/generic_keccak/xof.rs
jschneider-bensch: Good point, thanks for providing the assembly example!

jschneider-bensch (approved):
Looks good from my side!

libcrux-sha3/src/generic_keccak.rs
keks: Ah I think I misunderstood, never mind

keks (approved):
This looks good now!

PR #1028, opened 2025-06-20
`fix(F*/ml-{kem,dsa})`: refresh hardcoded super instance hashes

W95Psp:
This PR follows https://github.com/cryspen/hax/pull/1514#issuecomment-2989394599: it only updates super bounds hash that we have inlined in the Rust code because of https://github.com/cryspen/hax/issues/1313.


W95Psp:
I just forced merged the PR on hax.
(this is required since there is a dependency cycle in CI checks between libcrux CI and hax CI)

karthikbhargavan (approved):
It is unfortunate that our proofs are not robust with these identifier changes, but we have a task to fix this in the coming months: https://github.com/cryspen/hax/issues/1519

karthikbhargavan:
This is failing CI. I need to debug.

franziskuskiefer (approved):

Issue #974, opened 2025-05-19
Update C extraction


franziskuskiefer:
We need to update the C extraction scripts for ML-KEM and ML-DSA to catch up with recent tooling changes to be able to drop manual intervention

- [x] ~~Test and adapt to new changes in https://github.com/AeneasVerif/eurydice/pull/192
  This doesn't appear to be ready. See https://github.com/AeneasVerif/eurydice/issues/280#issuecomment-3309168401~~
- [x] Full extraction for testing (#1072)
- [x] Update extraction to work with current eurydice glue file https://github.com/AeneasVerif/eurydice/pull/166
- [x] Reduce CI for `mlkem/c` to build and test in debug mode (this code is unused and we shouldn't spend too much time maintaining and running it)
- [x] https://github.com/FStarLang/karamel/issues/637
- [x] https://github.com/AeneasVerif/eurydice/pull/258

djmdjm:
Yes, we've used the C header only version for a while, getting this back would be great - thanks

keks:
@franziskuskiefer I think this is done, right?

djmdjm:
I saw that the PR was merged but I haven't had a chance to test it for
OpenSSH yet. Don't let that stop you from closing this issue if you prefer
though.

franziskuskiefer:
Thanks @djmdjm let us know if the new version works for you or if there's anything else you need here.

For the general issue, the main part that's still open here is to make synchronising with upstream eurydice easier. We still use a custom version of the `eurydice_glue.h` file. That's not great and breaks CI on the tooling side. I started doing the split [here](https://github.com/cryspen/libcrux/compare/main...franziskus/c-glue-split) but didn't get around to doing this upstream. When we have individual files, we can pick the ones we need here in libcrux. Once that's done, we can close this issue.

djmdjm:
Last week I finally got around to updating OpenSSH to use libcrux v0.0.4. it worked fine, thanks for doing this.

PR #1046, opened 2025-07-02
c: add mlkem C extraction

franziskuskiefer:
Add a C extraction of ML-KEM again now that `cg` is C++.

This includes
- splitting out the common parts (`mlkem.cmake`, tests, benches, intrinsics)
- add C extraction to CI

This new C code requires C++20 when built with C++.
This works towards #974.

cc @djmdjm


libcrux-ml-kem/c/fix-symcrypt-includes.patch
keks, line 1: I was trying to find out why this was needed and isn't anymore, but I couldn't find a place where this is used. Has this been outdated before this PR?

franziskuskiefer: This was for local testing with symcrypt. I don't think it's needed anymore.
But cc'ing @protz in case he wants to keep that around somehow.

PR #1047, opened 2025-07-02
ML-DSA: Proofs: remaining avx2 deserialize functions

W95Psp:
This PR adds missing proofs on avx2 deserialize functions.
Now the low-level proofs on MLDSA avx2 enconding are completed.

Helps with https://github.com/cryspen/libcrux/issues/373.


libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
```diff
@@ -180,74 +180,134 @@ pub(crate) fn serialize(simd_unit: &Vec256, serialized: &mut [u8], gamma1_expone
...
         19 => serialized.len() == 20,
         _ => false,
-    })]
+})]
+#[hax_lib::ensures(|result| fstar!("deserialize_post $gamma1_exponent $serialized ${out}_future"))]
 pub(crate) fn deserialize(serialized: &[u8], out: &mut Vec256, gamma1_exponent: usize) {
-    match gamma1_exponent as u8 {
-        17 => deserialize_when_gamma1_is_2_pow_17(serialized, out),
-        19 => deserialize_when_gamma1_is_2_pow_19(serialized, out),
+    // TODO: why `gamma1_exponent as u8`, not `gamma1_exponent`?
```
jschneider-bensch: Matching on `usize` was an issue for C extraction at some point I believe, but I don't know if that's still a thing at the moment. cc @franziskuskiefer

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
```diff
@@ -180,74 +180,134 @@ pub(crate) fn serialize(simd_unit: &Vec256, serialized: &mut [u8], gamma1_expone
...
+       result.(mk_int ((i / bytes) * 32 + i % bytes))) /\
+    (forall (i: nat{i < 256}).
+       i % 32 >= bytes ==> Core_models.Abstractions.Bit.Bit_Zero? result.(mk_int i))
+"#
+)]
+#[hax_lib::fstar::before(r#"[@@ "opaque_to_smt"]"#)]
 #[hax_lib::requires(serialized.len() == 18)]
-fn deserialize_when_gamma1_is_2_pow_17(serialized: &[u8], out: &mut Vec256) {
+#[hax_lib::ensures(|result| fstar!("deserialize_unsigned_post (mk_int 17) $serialized $result"))]
+fn deserialize_when_gamma1_is_2_pow_17_unsigned(serialized: &[u8]) -> Vec256 {
```
jschneider-bensch: Can you put a small comment noting when you changed things from the `&mut` style to returning an owned type for the sake of a proof? That way we know in the future that this is an explicit exception to the general style, motivated by what the proof needs.

I'll run some measurements to verify that it doesn't have an impact on perf or stack (which I assume it does not).

jschneider-bensch (commented):
I think it looks good from the Rust side, just want to run some quick measurements before approving.

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
W95Psp: Ah, make sense.
I git blamed, and you're right, that's something related to the toolchain: 9bdda4c03597de873dd5e4e26b229aa7f6d73246.
On the commit it's about hax: if it is really just about hax, we should revert that.

Maybe I can push a PR and see if C extraction breaks?

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
W95Psp: Sure. I could also just put that back as a `&mut`, it probably has no consequences!

(I did that non-`&mut` style because it felt a bit wierd to me I think, no big reasons)

W95Psp: Done, we're back to a `&mut` style!

libcrux-ml-dsa/src/simd/avx2/encoding/gamma1.rs
W95Psp: Ah! I know what it is: with ahx + F* we previously could not can't match on `usize`s!
I closed https://github.com/cryspen/hax/issues/464 just now, since this is now possible.

I removed that comment line, and I pushing a fix PR.

W95Psp: https://github.com/cryspen/libcrux/pull/1056

libcrux-ml-dsa/src/simd/avx2/encoding/t0.rs
```diff
@@ -87,10 +87,23 @@ pub(crate) fn serialize(simd_unit: &Vec256, out: &mut [u8]) {
...
+       i % 32 >= bytes ==> Core_models.Abstractions.Bit.Bit_Zero? result.(mk_int i))
+"#
+)]
+#[hax_lib::fstar::before(r#"[@@ "opaque_to_smt"]"#)]
 #[hax_lib::requires(serialized.len() == 13)]
-pub(crate) fn deserialize(serialized: &[u8], out: &mut Vec256) {
-    debug_assert_eq!(serialized.len(), 13);
-
+#[hax_lib::ensures(|result| fstar!("deserialize_unsigned_post $serialized $result"))]
+fn deserialize_unsigned(serialized: &[u8]) -> Vec256 {
```
jschneider-bensch: Is it also possible to change this back to `&mut` as well?

karthikbhargavan: Hm, this one cannot be an &mut because the input and output have different types?

W95Psp:
Thanks for the reviews!
I agree @karthikbhargavan, I did that for simplicity and because everything is in F*, in contrast with everything in Rust.

I fixed the last `&mut` @jschneider-bensch, thanks for pointing it, I was missing it :)

I refreshed the C snapshots.
Putting it to merge queue now.

Issue #1043, opened 2025-06-30
[PSQ] TLSCodec as wire format


jschneider-bensch:

jschneider-bensch:
Resolved in #1048

PR #1048, opened 2025-07-02
PSQv2

### LINKED ISSUE #1040: [PSQ] Builder pattern for PSQ handshake state

### LINKED ISSUE #1041: [PSQ] Handshake state machine API
- `read_message`
- `write_message`
- `is_initiator`
- `is_handshake_finished`
- `into_transport_mode`
- ...

### LINKED ISSUE #1042: [PSQ] Transport state machine
See e.g. https://docs.rs/snow/latest/snow/struct.TransportState.html 
- `read_message`
- `write_message`
- ...

### LINKED ISSUE #1044: [PSQ] Implement protocol
Implement PSQ protocol variants.

### LINKED ISSUE #1063: [PSQ] Query mode

### LINKED ISSUE #1064: [PSQ] Registration mode

### LINKED ISSUE #1067: [PSQ] Benchmark v2

### LINKED ISSUE #1069: [PSQ] Implement Replay/DoS mitigation

jschneider-bensch:
This PR contains the initial version for PSQv2 with query and registration modes. It addresses the following issues
- #1040: We have a `Builder` that takes own longterm key pairs and peer longterm public keys and builds a `QueryInitiator`, `RegistrationInititator` or `Responder`, which all `impl Protocol`.
- #1041: The `Protocol` trait defines `read_message`, `write_message` and `is_receiver` as a uniform API for Initiators and Responder. In a slight difference to the `snow` API, `read_message` returns the pair of `(bytes_deserialized, out_bytes_written)` instead of just `out_bytes_written`.
- #1043: All messages are serialized with TLSCodec, and TLSCodec is used internally as well to ensure we have unique encodings for key derivation inputs. At the moment this does more copies than necessary, since we have the same types for serialization and deserialization. The reason is that we can't use the `Write` oriented API in `tls_codec` to write to a `&mut [u8]` at the moment, and not all base types in `tls_codec` implement the right combinations of `Serialize(Bytes)` and `Deserialize(Bytes)`. I opened #1066 to track our use of copies.
- #1063 & #1064: The `Responder` state machine is as follows:
 ```
                Initial -- read_message() --> RespondRegistration -- write_message() --> ToTransport
                |     ^
read_message() |     | write_message()
                v     |
               RespondQuery
```
The two `Initiator`s exercise the two legs of this state machine, with the `RegistrationInitiator` looking like this:
```
                Initial -- write_message() --> Waiting -- read_message() --> Done
```
The `ToTransport/Done` states will then be those states where `into_transport_mode` will give you a `TransportState` struct that implements the transport state machine (cf. #1042).
- #1069: I've chosen a (hopefully) simple approach here, where the Responder keeps a running list of recently seen initiator ephemeral keys, which can be seen after the first message deserialization. Before any cryptographic operations are performed the responder checks whether the ephemeral key was contained in the recently seen keys, and if so it does not process the message further. The number of keys that are stored can be given to the responder builder, since it might depend on the host machine at which point looking through this list is more expensive than fully processing the message.
- #1042: I've added a `Transport` struct which also implements the `Protocol` trait. Initiator and receiver implement `IntoTransport`, which consumes them and gives a `Transport`.
---

### Still missing
- [ ] Documentation
- [ ] More testing
- [ ] Placing v1 behind a feature and making this the default

Fixes #1040
Fixes #1041 
Fixes #1042
Fixes #1063 
Fixes #1064 
Fixes #1069  
Fixes #1067

### CODE REVIEW

File: libcrux-psq/src/protocol.rs [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+
+mod ecdh;
+mod initiator;
+mod keys;
+mod responder;
+mod session;
+mod signature;
+mod transcript;
+
+#[derive(Debug, PartialEq)]
```
franziskuskiefer: This really shouldn't be derived 😊

libcrux-psq/src/protocol/transcript.rs
```diff
@@ -0,0 +1,86 @@
...
+        initiator_longterm_pk: PublicKey,
+        responder_pq_pk: Option<libcrux_ml_kem::mlkem768::MlKem768PublicKey>,
+        pq_encaps: Option<libcrux_ml_kem::mlkem768::MlKem768Ciphertext>,
+    }
+
+    Transcript::add_hash::<TX1_DOMAIN_SEP>(
+        Some(*tx0),
+        &Transcript1Inputs {
+            initiator_longterm_pk: initiator_longterm_pk.clone(),
+            responder_pq_pk: responder_pq_pk.clone(),
```
franziskuskiefer: We don't need to copy things here.

libcrux-psq/src/protocol/transcript.rs
```diff
@@ -0,0 +1,86 @@
...
+impl Transcript {
+    fn new(initial_input: &impl SerializeBytes) -> Self {
+        Self::add_hash::<TX0_DOMAIN_SEP>(None, initial_input)
+    }
+
+    fn add_hash<const DOMAIN_SEPARATOR: u8>(
+        old_transcript: Option<Transcript>,
+        input: &impl SerializeBytes,
+    ) -> Transcript {
+        let mut domain_separated_input = vec![DOMAIN_SEPARATOR];
```
franziskuskiefer: pre-allocating vectors makes things faster.

libcrux-psq/src/protocol/transcript.rs
```diff
@@ -0,0 +1,86 @@
...
+    tx0: &Transcript,
+    initiator_longterm_pk: &PublicKey,
+    responder_pq_pk: &Option<libcrux_ml_kem::mlkem768::MlKem768PublicKey>,
+    pq_encaps: &Option<libcrux_ml_kem::mlkem768::MlKem768Ciphertext>,
+) -> Transcript {
+    #[derive(TlsSerializeBytes, TlsSize)]
+    struct Transcript1Inputs {
+        initiator_longterm_pk: PublicKey,
+        responder_pq_pk: Option<libcrux_ml_kem::mlkem768::MlKem768PublicKey>,
+        pq_encaps: Option<libcrux_ml_kem::mlkem768::MlKem768Ciphertext>,
```
franziskuskiefer: We can add `#[tls_codec(with = "mlkem_pk_codec")]` etc. here for now, or implement that directory in mlkem.

libcrux-psq/src/protocol/session.rs
```diff
@@ -0,0 +1,62 @@
...
+    ProtocolMode,
+};
+
+/// The length of a session ID in bytes.
+pub const SESSION_ID_LENGTH: usize = 32;
+
+/// The length of a sessin key in bytes.
+pub const SESSION_KEY_LENGTH: usize = 32;
+
+#[derive(Debug, PartialEq)]
```
franziskuskiefer: Don't derive these on anything that's key material (except for public keys).

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,218 @@
...
+    },
+}
+
+#[derive(TlsSerializeBytes, TlsDeserializeBytes, TlsSize)]
+// XXX: Determine what should go in here
+pub(crate) struct InitiatorInnerPayload {}
+
+impl InitiatorPre {
+    pub(crate) fn first_message(
+        mode: ProtocolMode,
```
franziskuskiefer: This should probably be a reference. In the registration case you don't want to copy all the keys around.

franziskuskiefer (commented):
Looks like this needs to be rebased onto main.

libcrux-psq/src/protocol/transcript.rs
jschneider-bensch: Trying to implement it in ML-KEM directly, now.

libcrux-psq/src/protocol/transcript.rs
jschneider-bensch: I think it makes sense to instead use the incremental hashing API here.

libcrux-psq/src/protocol/transcript.rs
jschneider-bensch: I've implemented that in ML-KEM now.

jschneider-bensch:
I've now implemented registration mode as well, and addressed your comments. In registration mode, the inner initiator payload is passed out for an application to validate, as discussed.
The PR is still missing a lot of documentation, but I would be happy to receive feedback on the overall structure, if possible.

libcrux-psq/src/protocol/ecdh.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub(crate) fn new(rng: &mut impl CryptoRng) -> Self {
+        Self(
+            libcrux_ecdh::generate_secret(libcrux_ecdh::Algorithm::X25519, rng)
+                .expect("Insufficient Randomness"),
+        )
+    }
+}
+
+impl From<&PrivateKey> for PublicKey {
+    fn from(value: &PrivateKey) -> Self {
```
franziskuskiefer: I think API wise I'd expect a `public` or `into_public` on the `PrivateKey`. Then the caller doesn't need to figure out the type of the `PublicKey` first.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -0,0 +1,243 @@
...
+use tls_codec::{DeserializeBytes, SerializeBytes, TlsSerializeBytes, TlsSize};
+
+use super::{
+    ecdh::{self, PrivateKey, PublicKey, SharedSecret},
+    session::{SessionKey, SESSION_ID_LENGTH},
+    transcript::{self, Transcript},
+};
+
+#[derive(TlsSerializeBytes, TlsSize)]
+pub struct AEADKey(pub [u8; KEY_LEN]);
```
franziskuskiefer: Should the raw key really be publicly accessible?

libcrux-psq/src/protocol.rs
```diff
@@ -0,0 +1,333 @@
...
+        let aad_initiator = b"Test Data I";
+        let aad_responder = b"Test Data R";
+
+        let responder_ecdh_sk = PrivateKey::new(&mut rng);
+        let responder_ecdh_pk = PublicKey::from(&responder_ecdh_sk);
+
+        let (initiator_pre, initiator_msg) = QueryInitiator::query(
+            &responder_ecdh_pk,
+            ctx.as_slice(),
+            aad_initiator.as_slice(),
```
franziskuskiefer: These two are slices already.

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        let (tx0, k0) = derive_k0(
+            responder_longterm_ecdh_pk,
+            &initiator_ephemeral_ecdh_pk,
+            &initiator_ephemeral_ecdh_sk,
+            ctx,
+            false,
+        );
+
+        let ciphertext = k0.serialize_encrypt(&InitiatorOuterPayload::Query, &aad);
```
franziskuskiefer: `aad` is a reference already.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -0,0 +1,243 @@
...
+                KEY_LEN,
+            )
+            .unwrap()
+            .try_into()
+            .unwrap(),
+        )
+    }
+
+    pub(crate) fn serialize_encrypt(&self, payload: &impl SerializeBytes, aad: &[u8]) -> Vec<u8> {
+        let mut ciphertext = vec![0u8; payload.tls_serialized_len() + 16];
```
franziskuskiefer: If we want to follow snow APIs, we should have the caller always pass in the output bytes to write into. That's usually nicer for networking to avoid coyping things around.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -0,0 +1,243 @@
...
+        let payload_serialized = payload.tls_serialize().unwrap();
+
+        let _ = encrypt(
+            self.as_ref(),
+            &payload_serialized,
+            &mut ciphertext,
+            aad,
+            &[0; NONCE_LEN],
+        )
+        .expect("Encryption Error");
```
franziskuskiefer: Let's make sure that every `unwrap` has a note on why it can't panic.

libcrux-psq/src/protocol/responder.rs
```diff
@@ -0,0 +1,190 @@
...
+                tx1,
+                k1,
+                pq: pq_encaps.is_some(),
+            },
+        )
+    }
+
+    pub fn decrypt_outer(
+        responder_longterm_ecdh_sk: &PrivateKey,
+        responder_longterm_ecdh_pk: &PublicKey,
```
franziskuskiefer: Putting them into a simple key pair wrapper would be nice.

libcrux-psq/src/protocol/responder.rs
```diff
@@ -0,0 +1,190 @@
...
+        } = initiator_msg;
+        let (tx0, k0) = derive_k0(
+            responder_longterm_ecdh_pk,
+            &initiator_ephemeral_ecdh_pk,
+            responder_longterm_ecdh_sk,
+            ctx,
+            true,
+        );
+
+        let received_payload = k0.decrypt_deserialize(&ciphertext, &aad);
```
franziskuskiefer: `aad` is a reference already. Clippy should tell you these things as well.

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,244 @@
...
+    responder_pq_pk: Option<&'keys MlKem768PublicKey>,
+    tx1: Transcript,
+    k1: AEADKey,
+}
+
+#[derive(TlsSerializeBytes, TlsDeserializeBytes, TlsSize)]
+#[repr(u8)]
+pub enum InitiatorOuterPayload {
+    Query,
+    Registration {
```
franziskuskiefer: Given the huge difference in size here, these should probably get boxed. The other option could be that these are all just references.

libcrux-psq/src/protocol.rs
```diff
@@ -0,0 +1,333 @@
...
+            &mut rng,
+        );
+
+        let initiator_msg_wire = initiator_msg
+            .tls_serialize()
+            .expect("Failed to serialize initiator message");
+        let initiator_msg_deserialized = Message::tls_deserialize_exact_bytes(&initiator_msg_wire)
```
franziskuskiefer: We could add a zero-copy version to TLS codec to reduce the number of copies. For things like MLS this didn't really make a difference so far, so I didn't. But we should keep an eye on this and see if we think it makes sense to add it.

libcrux-psq/src/protocol.rs
```diff
@@ -0,0 +1,333 @@
...
+            .expect("Failed to deserialize initiator message");
+
+        match Responder::decrypt_outer(
+            &responder_ecdh_sk,
+            &responder_ecdh_pk,
+            &initiator_msg_deserialized,
+            ctx.as_slice(),
+        ) {
+            (InitiatorOuterPayload::Query, tx0, k0) => {
```
franziskuskiefer: I don't like these kind of matches.

I don't know if it matters, because we want people to use the snow like API anyway (that we should add on top). But an `as_query_msg(&self) -> Option<&InitiatorOuterPayload::Query>` would make this a little nicer imo. But not too important.

libcrux-psq/src/protocol.rs
```diff
@@ -0,0 +1,333 @@
...
+                let responder_message = Responder::respond_query(
```
franziskuskiefer: But even with this match, this should be an `if let` instead, to avoid the unnecessary indentation.

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,244 @@
...
+            &tx2,
+        );
+
+        k2.decrypt_deserialize(&responder_msg.ciphertext, &responder_msg.aad)
+    }
+}
+
+impl<'keys> RegistrationInitiatorPre<'keys> {
+    pub fn registration_message(
+        initiator_longterm_ecdh_pk: &'keys PublicKey,
```
franziskuskiefer: That's a lot of arguments. Using key pairs could help a little.

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,244 @@
...
+            initiator_longterm_ecdh_sk,
+            responder_pq_pk,
+            responder_longterm_ecdh_pk,
+            aad_inner,
+            rng,
+            tx0,
+            &k0,
+        );
+
+        let payload0 = InitiatorOuterPayload::Registration {
```
franziskuskiefer: `encrypt_inner_payload` has all the information and could just return this struct.

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,244 @@
...
+                initiator_longterm_ecdh_sk,
+                initiator_longterm_ecdh_pk,
+                responder_pq_pk,
+                tx1,
+                k1,
+            },
+            Message {
+                ephemeral_ecdh_pk: initiator_ephemeral_ecdh_pk,
+                ciphertext: ciphertext_outer,
+                aad: aad_outer.to_vec(),
```
franziskuskiefer: If this is stored here, we should not pass it in as a reference, but owned.

libcrux-psq/src/protocol.rs
```diff
@@ -0,0 +1,333 @@
...
+                    responder_pq_pk,
+                    &tx0,
+                    &k0,
+                    &initiator_longterm_ecdh_pk,
+                    pq_encaps.as_ref(),
+                    &ciphertext,
+                    &aad,
+                );
+
+                // pretend we did some validation of the inner_payload here
```
franziskuskiefer: I feel like we should define some traits for doing this. Let's discuss later.

libcrux-psq/src/protocol/responder.rs
```diff
@@ -0,0 +1,190 @@
...
+            ephemeral_ecdh_pk: initiator_ephemeral_ecdh_pk,
+            ..
+        } = initiator_msg;
+        let responder_ephemeral_ecdh_sk = PrivateKey::new(rng);
+        let responder_ephemeral_ecdh_pk = PublicKey::from(&responder_ephemeral_ecdh_sk);
+        let tx2 = tx2(&tx1, &responder_ephemeral_ecdh_pk);
+        let k2 = derive_k2_registration_responder(
+            &k1,
+            &tx2,
+            &Box::new(initiator_longterm_ecdh_pk),
```
franziskuskiefer: Why box this here?

libcrux-psq/src/protocol/ecdh.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl AsRef<[u8]> for PrivateKey {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_slice()
+    }
+}
+
+impl SharedSecret {
+    pub(crate) fn derive(sk: &PrivateKey, pk: &PublicKey) -> SharedSecret {
+        SharedSecret(libcrux_ecdh::derive(Algorithm::X25519, &pk.0, &sk.0).unwrap())
```
franziskuskiefer: It looks like we can end up here without checking anything on the keys (incl. length). So this may well panic.

libcrux-psq/src/protocol/responder.rs
```diff
@@ -0,0 +1,190 @@
...
+            &initiator_ephemeral_ecdh_pk,
+            &responder_ephemeral_ecdh_sk,
+        );
+
+        let ciphertext = k2.serialize_encrypt(&ResponderRegistrationPayload {}, aad);
+
+        let responder_pq_pk = if *pq { Some(responder_pq_pk) } else { None };
```
franziskuskiefer: You could use `then_some` instead.

libcrux-psq/src/protocol/responder.rs
```diff
@@ -0,0 +1,190 @@
...
+            responder_pq_pk,
+            &k2,
+            &tx2,
+        );
+        (
+            session_state,
+            Message {
+                ephemeral_ecdh_pk: responder_ephemeral_ecdh_pk,
+                ciphertext,
+                aad: aad.to_vec(),
```
franziskuskiefer: `aad` should maybe be owned here again.

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,244 @@
...
+    pub fn complete_registration(self, responder_msg: &'keys Message) -> SessionState<'keys> {
+        let Self {
+            responder_longterm_ecdh_pk,
+            initiator_ephemeral_ecdh_sk,
+            initiator_longterm_ecdh_sk,
+            initiator_longterm_ecdh_pk,
+            responder_pq_pk,
+            tx1,
+            k1,
+        } = self;
```
franziskuskiefer: This style is always pretty noisy. Is there a reason you don't just access the fields on `self`?

libcrux-psq/src/protocol/session.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// The length of a sessin key in bytes.
+pub const SESSION_KEY_LENGTH: usize = 32;
+
+pub struct SessionKey {
+    pub(crate) identifier: [u8; SESSION_ID_LENGTH],
+    pub(crate) key: AEADKey,
+}
+
+pub struct SessionState<'keys> {
```
franziskuskiefer: How is this session state being loaded/stored? And how are the referenced keys being filled in when this happens.

libcrux-psq/src/protocol/keys.rs
jschneider-bensch: No, I used this for debugging and left it by accident.

libcrux-psq/src/protocol/ecdh.rs
jschneider-bensch: I added a `to_public` to `PrivateKey`.

libcrux-psq/src/protocol/keys.rs
jschneider-bensch: So you mean the application ultimately will pass in output bytes?

libcrux-psq/src/protocol/initiator.rs
jschneider-bensch: Right, I've decided to box the registration payload.

libcrux-psq/src/protocol.rs
jschneider-bensch: I've added something like you describe :+1:

libcrux-psq/src/protocol/initiator.rs
jschneider-bensch: I've made it so the application can provide an optional `aad` by reference, now.

libcrux-psq/src/protocol/responder.rs
jschneider-bensch: I'm not sure why I boxed this here. I think I changed back and forth a number of times on where we only have references and where we have owned types.

libcrux-psq/src/protocol/ecdh.rs
jschneider-bensch: I'm returning a `Result` here now.

libcrux-psq/src/protocol/responder.rs
jschneider-bensch: I think it's still not consistent what's owned and what not everywhere. I think I had some trouble with serialization when there were named lifetimes.

libcrux-psq/src/protocol/initiator.rs
jschneider-bensch: It thought it might make it more readable, but I went with shorter names in many places now and changed access in the way you proposed.

libcrux-psq/src/protocol/session.rs
jschneider-bensch: I haven't implemented session storage yet, but the idea was to use `tls_codec` for that as well. But indeed I wonder how the references should be handled

libcrux-psq/src/protocol.rs
jschneider-bensch: I've created #1066 so we can work on minimizing copies overall.

libcrux-psq/src/lib.rs
```diff
@@ -10,6 +10,10 @@ use std::array::TryFromSliceError;
 #[derive(Debug)]
 /// PSQ Errors.
 pub enum Error {
+    /// An error during serialization.
+    Serialization,
+    /// The Initator message was stale
```
franziskuskiefer:
```suggestion
    /// The Initiator message was stale
```

libcrux-psq/src/protocol.rs
```diff
@@ -0,0 +1,347 @@
...
+pub(crate) fn write_output(payload: &[u8], out: &mut [u8]) -> Result<usize, Error> {
+    let payload_len = payload.len();
+    if out.len() < payload_len {
+        return Err(Error::OutputBufferShort);
+    }
+    out[..payload_len].copy_from_slice(payload);
+    Ok(payload_len)
+}
+
+// pub struct TransportMessage {
```
franziskuskiefer: We should either remove or uncomment this code.

libcrux-psq/src/protocol/transcript.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    fn add_hash<const DOMAIN_SEPARATOR: u8>(
+        old_transcript: Option<&Transcript>,
+        input: &impl SerializeBytes,
+    ) -> Transcript {
+        let mut hasher = libcrux_sha2::Sha256::new();
+        hasher.update(&[DOMAIN_SEPARATOR]);
+        hasher.update(old_transcript.tls_serialize().unwrap().as_slice());
+        hasher.update(input.tls_serialize().unwrap().as_slice());
```
franziskuskiefer: Please add a note to everything that can panic. So here we need a reason why the `unwrap`s are safe.

libcrux-psq/src/protocol/ecdh.rs
```diff
@@ -0,0 +1,85 @@
+//! KEM type wrappers
+//!
+//! This module provides wrappers around KEM types, assuming a DH-KEM
+//! style API (for now). In the future this module should become
+//! obsolete, since we can use libcrux KEM trait implementers
+//! directly. (cf. https://github.com/cryspen/libcrux/issues/1035)
```
franziskuskiefer: That issue is closed. Is the comment outdated or shall this get addressed?

libcrux-psq/src/protocol/ecdh.rs
```diff
@@ -0,0 +1,85 @@
...
+/// A wrapper around a KEM public key.
+pub struct PublicKey(Vec<u8>);
+
+impl AsRef<[u8]> for PublicKey {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_slice()
+    }
+}
+
+#[derive(Clone)]
```
franziskuskiefer: I was wondering where we needed to clone this, which led me down the rabbit hole of changing the state style a little, avoiding unnecessary copies.

libcrux-psq/src/protocol/api.rs
```diff
@@ -0,0 +1,154 @@
...
+use libcrux_ml_kem::mlkem768::{MlKem768KeyPair, MlKem768PublicKey};
+
+use super::{ecdh::KEMKeyPair, initiator::RegistrationInitiator};
+
+#[derive(Debug)]
+pub enum Error {
+    BuilderState,
+    Serialize(tls_codec::Error),
+    Deserialize(tls_codec::Error),
+    CryptoError,
```
franziskuskiefer, line 26: I guess we could pass this one out as well. But that requires wrapping all of the errors here again. So I'm not sure if it's worth it.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -0,0 +1,256 @@
...
+        Ok(AEADKey(
+            libcrux_hkdf::expand(
+                libcrux_hkdf::Algorithm::Sha256,
+                prk,
+                info.tls_serialize().map_err(|e| Error::Serialize(e))?,
+                KEY_LEN,
+            )
+            .map_err(|_| Error::CryptoError)?
+            .try_into()
+            .expect("HKDF expands to `KEY_LEN` bytes"),
```
franziskuskiefer: Since this function is already returning an error, I'd prefer returning an error rather than unwrapping.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -0,0 +1,256 @@
...
+        [],
+        key.tls_serialize().map_err(|e| Error::Serialize(e))?,
+    )
+    .map_err(|_| Error::CryptoError)?;
+
+    Ok(libcrux_hkdf::expand(
+        libcrux_hkdf::Algorithm::Sha256,
+        prk,
+        SESSION_KEY_INFO,
+        SESSION_ID_LENGTH,
```
franziskuskiefer: Same, this should be an error not a panic.

libcrux-psq/src/protocol/api.rs
```diff
@@ -0,0 +1,154 @@
...
+
+pub struct TransportState;
+
+pub trait IntoTransport {
+    fn into_transport_mode(self) -> TransportState;
+    fn is_handshake_finished(&self) -> bool;
+}
+
+pub trait HandshakeState {
+    fn is_initiator(&self) -> bool;
```
franziskuskiefer: Do we use this function? It amy be handy indeed. I just wonder where we use it.

libcrux-psq/src/protocol/api.rs
```diff
@@ -0,0 +1,154 @@
...
+        self.outer_aad = outer_aad;
+        self
+    }
+
+    pub fn longterm_ecdh_keys(mut self, longterm_ecdh_keys: &'a KEMKeyPair) -> Self {
+        self.longterm_ecdh_keys = Some(longterm_ecdh_keys);
+        self
+    }
+
+    pub fn longterm_pq_keys(mut self, longterm_pq_keys: &'a MlKem768KeyPair) -> Self {
```
franziskuskiefer: This won't allow crypto agility. We should at least use a kem key pair here as well to allow using different algorithms.

libcrux-psq/src/protocol/api.rs
```diff
@@ -0,0 +1,154 @@
...
+    // builders
+    pub fn build_query_initiator(self) -> Result<QueryInitiator<'a>, Error> {
+        let Some(responder_ecdh_pk) = self.peer_longterm_ecdh_pk else {
+            return Err(Error::BuilderState);
+        };
+
+        QueryInitiator::new(responder_ecdh_pk, self.context, self.outer_aad, self.rng)
+    }
+
+    pub fn build_registration_initiator(self) -> Result<RegistrationInitiator<'a, Rng>, Error> {
```
franziskuskiefer, line 231: I thought you wanted to make this one initiator? What's the reason to have two different ones after all?

libcrux-psq/src/protocol/api.rs
```diff
@@ -0,0 +1,154 @@
...
+            self.rng,
+        )
+    }
+
+    pub fn build_responder(self) -> Result<Responder<'a, Rng>, Error> {
+        let Some(longterm_ecdh_keys) = self.longterm_ecdh_keys else {
+            return Err(Error::BuilderState);
+        };
+
+        let Some(longterm_pq_keys) = self.longterm_pq_keys else {
```
franziskuskiefer: So we can't have a non pq-responder? Shouldn't this just be used, like in the initiator?

libcrux-psq/src/protocol/api.rs
```diff
@@ -0,0 +1,154 @@
...
+
+pub struct Builder<'a, Rng: CryptoRng> {
+    rng: &'a mut Rng,
+    context: &'a [u8],
+    inner_aad: &'a [u8],
+    outer_aad: &'a [u8],
+    longterm_ecdh_keys: Option<&'a KEMKeyPair>,
+    longterm_pq_keys: Option<&'a MlKem768KeyPair>,
+    peer_longterm_ecdh_pk: Option<&'a PublicKey>,
+    peer_longterm_pq_pk: Option<&'a MlKem768PublicKey>,
```
franziskuskiefer: We talked about putting the keys into a ciphersuite. Is that still the plan.

libcrux-psq/src/util.rs
```diff
@@ -0,0 +1,32 @@
+pub(crate) mod mlkem_pk_codec {
```
franziskuskiefer: Do we still need this file?

libcrux-psq/src/protocol/transcript.rs
jschneider-bensch: Right, I missed those. They should error, if serialization fails.

libcrux-psq/src/protocol/ecdh.rs
jschneider-bensch: Indeed, now that the KEM traits are merged I will try to use that.

libcrux-psq/src/protocol/api.rs
jschneider-bensch: We don't currently use it, I just inlcuded it to match the `snow` API.

libcrux-psq/src/protocol/api.rs
jschneider-bensch: As far as I understood, one of the ideas of query mode was to ask for ciphersuite support on the responder side. Having one initiator would then mean the initiator needs a way to dynamically update it's ciphersuite which I thought was more complicated than necessary.

libcrux-psq/src/protocol/api.rs
jschneider-bensch: Yes, I did it like this because we don't have ciphersuites yet and then the responder should be able to answer a PQ-requesting client as well as classical clients. I'll update this as we introduce ciphersuite.

libcrux-psq/src/util.rs
jschneider-bensch: No, you're right.

libcrux-psq/src/protocol/api.rs
jschneider-bensch: For now, I didn't change this.

libcrux-psq/src/protocol/api.rs
jschneider-bensch: I've removed it, so the same trait (now `Protocol`) makes sense for the `Transport` struct.

jschneider-bensch:
I've now implemented the transformation to a secure channel `Transport` state.

libcrux-ml-kem/src/lib.rs
```diff
@@ -90,9 +90,6 @@ pub(crate) mod hax_utils;
 // This is being tracked in https://github.com/hacspec/hacspec-v2/issues/27
 pub(crate) mod constants;
 
-#[cfg(all(feature = "alloc", feature = "incremental"))]
```
franziskuskiefer, line 93: This is where the CI fails. Looks like this is used in some feature combination.

libcrux-psq/tests/api.rs
```diff
@@ -0,0 +1,208 @@
+use libcrux_ml_kem::mlkem768;
+use libcrux_psq::protocol::{
+    api::{IntoTransport, Protocol},
+    ecdh::KEMKeyPair,
+    *,
+};
+use rand::RngCore;
+
+#[test]
+fn query() {
```
franziskuskiefer: Let's split this into a query and a registration test case file.

libcrux-psq/src/protocol/initiator.rs
```diff
@@ -0,0 +1,323 @@
...
+
+        self.state = RegistrationInitiatorState::ToTransport(ToTransportState { tx2, k2 }.into());
+
+        Ok((bytes_deserialized, out_bytes_written))
+    }
+}
+
+impl<'a, Rng: CryptoRng> IntoTransport for RegistrationInitiator<'a, Rng> {
+    fn into_transport_mode(mut self) -> Result<Transport, Error> {
+        let RegistrationInitiatorState::ToTransport(state) = take(&mut self.state) else {
```
franziskuskiefer: No need to `take` the state here. You consume `self` anyway.

```suggestion
    fn into_transport_mode(self) -> Result<Transport, Error> {
        let RegistrationInitiatorState::ToTransport(state) = self.state else {
```

libcrux-psq/src/protocol/keys.rs
```diff
@@ -0,0 +1,290 @@
...
+    Ok(
+        libcrux_hkdf::expand(Algorithm::Sha256, prk, SESSION_KEY_INFO, SESSION_ID_LENGTH)
+            .map_err(|_| Error::CryptoError)?
+            .try_into()
+            .map_err(|_| Error::CryptoError)?, // We don't expect this to fail, unless HDKF gave us the wrong output length
+    )
+}
+
+// skCS = KDF(K2, "shared secret" | tx2)
+pub(super) fn derive_session_key(k2: &AEADKey, tx2: &Transcript) -> Result<SessionKey, Error> {
```
franziskuskiefer: We could consume the key and transcript here to avoid reusing it.

libcrux-psq/src/protocol/responder.rs
```diff
@@ -0,0 +1,375 @@
...
+                let out_bytes_written = write_output(initiator_inner_payload.0.as_slice(), out)?;
+                Ok((bytes_deserialized, out_bytes_written))
+            }
+        }
+    }
+}
+
+impl<'a, Rng: CryptoRng> IntoTransport for Responder<'a, Rng> {
+    fn into_transport_mode(mut self) -> Result<Transport, Error> {
+        let ResponderState::ToTransport(state) = take(&mut self.state) else {
```
franziskuskiefer: Same here.

```suggestion
    fn into_transport_mode(self) -> Result<Transport, Error> {
        let ResponderState::ToTransport(state) = self.state else {
```

libcrux-psq/benches/psq_v2.rs
```diff
@@ -0,0 +1,279 @@
...
+use libcrux_ml_kem::mlkem768::{self, MlKem768KeyPair};
+use libcrux_psq::protocol::{
+    api::{Builder, Protocol},
+    ecdh::KEMKeyPair,
+    initiator::QueryInitiator,
+    responder::Responder,
+};
+use rand::CryptoRng;
+
+fn query<const PQ: bool>(c: &mut Criterion) {
```
franziskuskiefer, line 21: Let's add registration as well to benchmark the transport channel as well.

libcrux-psq/src/protocol/ecdh.rs
franziskuskiefer: This is still open?

libcrux-psq/src/protocol/api.rs
franziskuskiefer: As discussed, let's add a simple wrapper for now and add the ciphersuite in a follow-up.

libcrux-psq/src/protocol/ecdh.rs
jschneider-bensch: I've removed the comment. At the moment we plan to have X25519 only for this KEM, but I'll leave the wrapper here in case we want to extend that at some point.

libcrux-psq/src/protocol/api.rs
```diff
@@ -0,0 +1,268 @@
...
+
+#[derive(TlsDeserialize, TlsSize)]
+struct TransportMessage {
+    ciphertext: VLBytes,
+    tag: [u8; 16],
+}
+
+impl Protocol for Transport {
+    fn write_message(&mut self, payload: &[u8], out: &mut [u8]) -> Result<usize, Error> {
+        let mut ciphertext = vec![0u8; payload.len()];
```
franziskuskiefer, line 74: Should we set a max payload size here? For noise the max message length is 65535 bytes.

libcrux-psq/src/protocol/session.rs
```diff
@@ -0,0 +1,12 @@
+use super::keys::AEADKey;
+
+/// The length of a session ID in bytes.
+pub const SESSION_ID_LENGTH: usize = 32;
+
+/// The length of a sessin key in bytes.
+pub const SESSION_KEY_LENGTH: usize = 32;
+
+pub struct SessionKey {
```
franziskuskiefer, line 10: We'll need a way to store and retrieve these keys.
There are also the public keys we may want to bind to this in some way. We can do that in a followup. But let's note down what exactly is missing here.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -0,0 +1,292 @@
...
+        ))
+    }
+
+    fn increment_nonce(&mut self) {
+        let mut buf = [0u8; 16];
+        buf[16 - NONCE_LEN..].copy_from_slice(self.1.as_slice());
+        let mut nonce = u128::from_be_bytes(buf);
+        nonce += 1;
+        let buf = nonce.to_be_bytes();
+        self.1.copy_from_slice(&buf[16 - NONCE_LEN..]);
```
franziskuskiefer, line 59: We have 4 bytes for the counter (16-12). So doing this on a `u32` sounds more reasonable. And even if unlikely, we need to throw an error if the counter wraps.

libcrux-psq/benches/psq_v2.rs
```diff
@@ -0,0 +1,755 @@
...
+                        .write_message(registration_payload_responder, &mut msg_channel)
+                        .unwrap();
+
+                    let (_len_i_deserialized, _len_i_payload) = initiator
+                        .read_message(&msg_channel, &mut payload_buf_initiator)
+                        .unwrap();
+
+                    let mut initiator = initiator.into_transport_mode().unwrap();
+                    let _ = initiator
+                        .write_message(b"Transport Message", &mut msg_channel)
```
franziskuskiefer: This looks like it doesn't take any significant time (~359ns) for me. Shall we use a somewhat bigger payload to see a less noisy number? But we should also figure out what usual payload sizes are here in order to measure the right thing.

libcrux-psq/src/protocol/keys.rs
jschneider-bensch: I'm not sure I understand regarding the counter: `NONCE_LEN` is 12 for ChaChaPoly, so we should have 12 bytes for the counter, no?

But I agree about throwing an error on wrap.

libcrux-psq/src/protocol/api.rs
jschneider-bensch: We should only be limited by the AEAD's maximum message length here, right? I think for our (hacl-rs) ChaChaPoly thats `u32::MAX`, but if we want to match noise, that's fine too I think.

libcrux-psq/src/protocol/session.rs
jschneider-bensch: I've made #1077 and linked it here.

PR #1053, opened 2025-07-08
Add and implement KEM traits

keks:
This PR adds a number of KEM traits and implements them for all our KEMs. All of them are derand-style.

Traits:
- `libcrux-traits::kem::arrayref::Kem`: has const generics for all the lengths. Contains associated functions (i.e. methods have no receiver). Functions take referenves to arrays of the correct length as arguments. Outputs are written to mutable references.
- `libcrux-traits::kem::slice::Kem`: has no const generics. Generally similar to the `arrayref` variant, but takes slices instead of array references. Error types contain variants for length mismatches.
- `libcrux-traits::kem::owned::Kem`: similar to arrayref, but instead of writing to mutable references, outputs are returned as arrays. Has a blanket impl for all arrayref::Kem.
- `libcrux-traits::kem::secrets::Kem`: similar to owned, but also classifies all returned values and consumes classified values. This is a step to get closer to secret independence. Has a blanket impl for all owned.

slice does not have a blanket impl for all arrayref because that doesn't work:
- the const generics needed to describe the arrayref::Kem are not use in the type or the slice::Kem
- there could be multiple implementations for arrayref::Kem, and we can only impl slice::Kem once

That's why there is a macro that impls slice::Kem for any type that impls arrayref::Kem.

## What is not in here?

- This PR does not remove existing code, but maybe it should.
- There are no randomized APIs, but we should be able to add randomized traits and blanket impl them for these traits
  - Maybe this won't work for similar reasons as with slice, but with the RAND const generics. in that case we'll have to do macros again.

---

This might still be a little bit messy and not super well documented, but I hope it's good enough for a first look.


curve25519/src/lib.rs
```diff
@@ -29,3 +29,51 @@ trait Curve25519 {
...
+
+impl libcrux_traits::kem::arrayref::Kem<SK_LEN, PK_LEN, PK_LEN, SHK_LEN, SK_LEN, SK_LEN>
+    for X25519
+{
+    fn keygen(
+        ek: &mut [u8; SK_LEN],
+        dk: &mut [u8; PK_LEN],
+        rand: &[u8; SK_LEN],
+    ) -> Result<(), libcrux_traits::kem::arrayref::KeyGenError> {
+        dk.copy_from_slice(rand);
```
jschneider-bensch, line 41: In `libcrux_ecdh::x25519::generate_secret` we do rejection sampling to avoid an all zero key. We should probably return an error here as well, if `rand` is all zeroes.

curve25519/src/lib.rs
```diff
@@ -29,3 +29,51 @@ trait Curve25519 {
...
+        Ok(())
+    }
+
+    fn encaps(
+        ct: &mut [u8; PK_LEN],
+        ss: &mut [u8; SHK_LEN],
+        ek: &[u8; PK_LEN],
+        rand: &[u8; SK_LEN],
+    ) -> Result<(), libcrux_traits::kem::arrayref::EncapsError> {
+        let mut eph_dk = *rand;
```
jschneider-bensch, line 53: Analogous to `keygen` we should error if `rand` is zero.

curve25519/src/lib.rs
jschneider-bensch, line 1: Would it be out of scope to change the constant names to align more with those defined in the KEM trait?

libcrux-kem/src/kem.rs
```diff
@@ -883,6 +890,135 @@ mod xwing {
 
     use super::*;
 
+    const MLKEM768_EK_LEN: usize = libcrux_ml_kem::mlkem768::MlKem768PublicKey::len();
+    const MLKEM768_DK_LEN: usize = libcrux_ml_kem::mlkem768::MlKem768PrivateKey::len();
+    const MLKEM768_CT_LEN: usize = libcrux_ml_kem::mlkem768::MlKem768Ciphertext::len();
+    const MLKEM768_SS_LEN: usize = 32;
```
jschneider-bensch: Not that these would be likely to change at all, but we do expose `libcrux_ml_kem::{SHARED_SECRET_SIZE, ENCAPS_SEED_SIZE, KEY_GENERATION_SEED_SIZE}`.

libcrux-kem/src/kem.rs
```diff
@@ -883,6 +890,135 @@ mod xwing {
...
+    const MLKEM768_RAND_KEYGEN_LEN: usize = 64;
+    const MLKEM768_RAND_ENCAPS_LEN: usize = MLKEM768_SS_LEN;
+
+    const X25519_EK_LEN: usize = 32;
```
jschneider-bensch: I think these could also be taken from `curve25519`.

libcrux-ml-kem/Cargo.toml
```diff
@@ -31,11 +31,12 @@ libcrux-platform = { version = "0.0.2", path = "../sys/platform" }
 libcrux-sha3 = { version = "0.0.3", path = "../libcrux-sha3" }
 libcrux-intrinsics = { version = "0.0.3", path = "../libcrux-intrinsics" }
 libcrux-secrets = { version = "0.0.3", path = "../secrets" }
+libcrux-traits = { version = "0.0.3", path = "../traits" }
 hax-lib.workspace = true
 
 [features]
 # By default all variants and std are enabled.
-default = ["default-no-std", "std"]
+default = ["default-no-std", "std", "kyber"]
```
jschneider-bensch: I don't think we want to have `kyber` in the default features. Left here by accident?

libcrux-ml-kem/src/lib.rs
```diff
@@ -171,3 +174,69 @@ cfg_kyber! {
...
+            fn keygen(
+                ek: &mut [u8; CPA_PKE_PUBLIC_KEY_SIZE],
+                dk: &mut [u8; SECRET_KEY_SIZE],
+                rand: &[u8; KEY_GENERATION_SEED_SIZE],
+            ) -> Result<(), libcrux_traits::kem::owned::KeyGenError> {
+                let key_pair = generate_key_pair(*rand);
+                ek.copy_from_slice(key_pair.pk());
+                dk.copy_from_slice(key_pair.sk());
+
+                // TODO: validate that the key is valid (how?)
```
jschneider-bensch: Key validation is described in [section 7](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf#section.7) of the standard as something you should do before encapsulation and decapsulation. For this we have `mlkem$param::validate_public_key` and `mlkem$param::validate_private_key` (which also takes the ciphertext as an argument just to check that it has a type with the right length). AFAIK we have decided that these should not be automatically included in the standard encaps/decaps calls, but it should be up to the library user to call them before encaps/decaps.

Regarding validation related to keygen: If you did not generate the key pair yourself, but read it from somewhere you can also optionally do a key pair check, but I don't think we implement that anywhere at the moment.

traits/src/kem/slice.rs
```diff
@@ -0,0 +1,215 @@
...
+                "the provided decapsulation key has the wrong length"
+            }
+        };
+
+        f.write_str(text)
+    }
+}
+
+#[cfg(feature = "error_in_core")]
+/// Here we implement the Error trait. This has only been added to core relatively recently, so we
```
jschneider-bensch, line 224: Can you copy this comment to the feature definition in `Cargo.toml` as well?

traits/src/kem/slice.rs
```diff
@@ -0,0 +1,215 @@
...
+        match value {
+            arrayref::DecapsError::InvalidCiphertext => DecapsError::InvalidCiphertext,
+            arrayref::DecapsError::InvalidDecapsKey => DecapsError::InvalidDecapsKey,
+            arrayref::DecapsError::Unknown => DecapsError::Unknown,
+        }
+    }
+}
+
+#[macro_export]
+/// Implements [`Kem`] for any [`arrayref::Kem`]. Hides the impl in a module of the given name.
```
jschneider-bensch:
```suggestion
/// Implements [`Kem`] for any [`arrayref::Kem`].
```
Or did I miss it somehow?

p256/src/impl_kem.rs
```diff
@@ -0,0 +1,80 @@
...
+        ek: &mut [u8; EK_LEN],
+        dk: &mut [u8; DK_LEN],
+        rand: &[u8; RAND_KEYGEN_LEN],
+    ) -> Result<(), KeyGenError> {
+        if !super::p256::validate_private_key(rand.as_slice()) {
+            return Err(KeyGenError::InvalidRandomness);
+        }
+
+        dk.copy_from_slice(rand.as_slice());
+        if !super::p256::dh_initiator(ek, dk) {
```
jschneider-bensch, line 27: Not for this PR, but looking at the differences between this and `libcrux_ecdh::p256_internal::key_gen`, it looks like that will run `validate_scalar` on the same scalar three times. :O

jschneider-bensch (changes requested):
I think the traits look pretty good, I have mostly minor comments.
But indeed, as you mentioned it would be good to have more documentation in the `traits` crate, as well as tests that exercise the impls.

libcrux-ml-kem/src/lib.rs
keks: So no checks in keygen an encaps, other than rand != [0]?

jschneider-bensch: Yeah, I think that would be in line with what we do right now.

traits/src/kem/tests.rs
jschneider-bensch, line 1: Nice! But we're not using it anywhere at the moment, right?

libcrux-ml-kem/src/lib.rs
```diff
@@ -171,3 +174,80 @@ cfg_kyber! {
...
+                KEY_GENERATION_SEED_SIZE,
+                SHARED_SECRET_SIZE,
+            > for $variant
+        {
+            fn keygen(
+                ek: &mut [u8; CPA_PKE_PUBLIC_KEY_SIZE],
+                dk: &mut [u8; SECRET_KEY_SIZE],
+                rand: &[u8; KEY_GENERATION_SEED_SIZE],
+            ) -> Result<(), libcrux_traits::kem::owned::KeyGenError> {
+                if rand.iter().fold(0, crate::or) == 0 {
```
jschneider-bensch: In `curve25519` we disallow zero randomness because it results in a bad key, but in ML-KEM I don't think there's a reason to disallow all zero randomness for keygen or encaps.

traits/src/kem/tests.rs
keks: I inject it in the macro for implementing the slice variant. Not super clean, but it did the trick...

keks: I could put it into a second macro or rename the macro, not sure what the best way to do that is.

libcrux-ml-kem/src/lib.rs
keks: Hm, I figured zero randomness is always a problem because it suggests that no randomness has been generated. If the zero key itself is the problem, then we should first clamp and then check that the key is valid eg [here](https://github.com/cryspen/libcrux/blob/bdf3bbb83d2c97a8c6886bb4d3c992e5f42c28a4/libcrux-ecdh/src/x25519.rs#L45-L50).

traits/src/kem/tests.rs
jschneider-bensch: Ah, sorry, I just missed it then! I think doing it like that is fine. Maybe we can document that the macro adds a test.

keks:
Looks like CI has been failing due to rate limiting, retrying after a bit of a wait

PR #1057, opened 2025-07-15
Add AEAD traits

keks:
This PR adds the AEAD traits.


traits/src/aead/arrayref.rs
```diff
@@ -0,0 +1,58 @@
+//! This module contains the trait and related errors for an Authenticated
+//! Encryption with Associated Data (AEAD) scheme that takes array references
+//! as arguments and writes outputs to mutable array references.
+
+/// An Authenticated Encryption with Associated Data (AEAD) scheme. This trait
+/// is low-level and is mostly used for implementing other, more usable APIs.
+pub trait Aead<const KEY_LEN: usize, const TAG_LEN: usize, const NONCE_LEN: usize> {
+    /// Encrypt a plaintext message, producing a ciphertext and an authentication tag.
+    /// The `ciphertext` argument must have length `plaintext.len() + TAG_LEN`.
```
franziskuskiefer: Why does the tag need to go into the ciphertext? That's what the separate `tag` argument is there for? Same goes for decrypt.

traits/src/aead/arrayref.rs
```diff
@@ -0,0 +1,58 @@
...
+        ciphertext: &[u8],
+        tag: &[u8; TAG_LEN],
+    ) -> Result<(), DecryptError>;
+}
+
+/// Error that can occur during encryption.
+#[derive(Debug, PartialEq, Eq)]
+pub enum EncryptError {
+    /// The ciphertext buffer has the wrong length.
+    WrongCiphertextText,
```
franziskuskiefer: This sounds like the ciphertext was wrong, not it's length. Same for the decryption error.

traits/src/aead/arrayref.rs
```diff
@@ -0,0 +1,58 @@
+//! This module contains the trait and related errors for an Authenticated
+//! Encryption with Associated Data (AEAD) scheme that takes array references
+//! as arguments and writes outputs to mutable array references.
+
+/// An Authenticated Encryption with Associated Data (AEAD) scheme. This trait
+/// is low-level and is mostly used for implementing other, more usable APIs.
+pub trait Aead<const KEY_LEN: usize, const TAG_LEN: usize, const NONCE_LEN: usize> {
```
franziskuskiefer, line 11: What about keygen?

keks: I had removed that because in the derand world it's a little redundant; it would only copy bytes from the rand buffer into they key buffer.

traits/src/aead/arrayref.rs
keks: Yeah the comment is wrong, I'll fix that

traits/src/aead/secrets.rs
```diff
@@ -0,0 +1,136 @@
+//! This module contains the trait and related errors for an Authenticated
+//! Encryption with Associated Data (AEAD) scheme that takes array references
+//! as arguments and returns outputs as arrays or vectors.
```
wysiwys: This description should add a line about the `libcrux-secrets` types used (since currently this description is the same as the one for the `owned` trait)

```suggestion
//! This module contains the trait and related errors for an Authenticated
//! Encryption with Associated Data (AEAD) scheme that takes array references
//! as arguments and returns outputs as arrays or vectors. Secret values use
//! the types from [libcrux-secrets](libcrux-secrets).
```

traits/src/aead/secrets.rs
```diff
@@ -0,0 +1,136 @@
...
+/// here. Check the documentation of the types implementing this trait to make sure which inputs
+/// are valid.
+pub trait Aead<const KEY_LEN: usize, const TAG_LEN: usize, const NONCE_LEN: usize> {
+    /// Encrypt a plaintext message, producing a ciphertext and an authentication tag.
+    fn encrypt<const MSG_LEN: usize>(
+        key: &[U8; KEY_LEN],
+        nonce: &[u8; NONCE_LEN],
+        aad: &[u8],
+        plaintext: &[U8; MSG_LEN],
+    ) -> Result<([u8; MSG_LEN], [u8; TAG_LEN]), EncryptError>;
```
wysiwys, line 22: From a user's perspective, for functions like these that return a tuple, a type with more semantic information could be returned here instead (like a struct with `ciphertext` and `tag` fields), to eliminate any confusion

traits/src/aead/slice.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+}
+
+#[cfg(feature = "error_in_core")]
+/// Here we implement the Error trait. This has only been added to core relatively recently, so we
+/// are hiding that behind a feature.
+mod error_in_core {
+    impl core::error::Error for super::EncryptError {}
+    impl core::error::Error for super::DecryptError {}
```
wysiwys, line 157: These error types are missing a `core::fmt::Display` implementation (which is needed by `core::error::Error`), which currently causes a build error when `error_in_core` is enabled

wysiwys (changes requested):
This looks good to me! There is one build error for the new `error_in_core` feature flag that appears to be due to a missing `Display` implementation. Other than that, I added one nit and a more general comment.

traits/src/aead/secrets.rs
keks: Yes! Let's put that on the doc as an option for the next iteration. I think for this round we should just stick with the style we already merged for the KEM.

traits/src/aead/slice.rs
keks: Ah, you are right! And I suppose we should also check that this builds on CI with the features enabled

keks:
I'll rebase after the review

.github/workflows/traits-build.yml
```diff
@@ -0,0 +1,106 @@
...
+        if: ${{ matrix.bits == 32 &&  matrix.os == 'ubuntu-latest' }}
+        run: |
+          rustup target add i686-unknown-linux-gnu
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib g++-multilib
+
+      - name: 🛠️ Setup Ubuntu x64
+        if: ${{ matrix.bits == 64 &&  matrix.os == 'ubuntu-latest' }}
+        run: |
+          rustup target add aarch64-unknown-linux-gnu
```
wysiwys: For compilation on Ubuntu x86-64, this target does not appear to have been provided to `rustc` (instead, it uses the target `x86_64-unknown-linux-gnu`)

.github/workflows/traits-build.yml
```diff
@@ -0,0 +1,106 @@
...
+env:
+  CARGO_TERM_COLOR: always
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    if: ${{ github.event_name != 'merge_group' }}
```
wysiwys, line 20: On CI, will we check that this workflow succeeds using a required status check? 

If we skip this job on the merge queue, it could block the merge queue from completing (since the `build` job would be skipped rather than succeeding). To address this, a status check job at the end could report success if the `build` job is skipped.

.github/workflows/traits-build.yml
keks: Huh, I copied this from some other workflow (kem-build-test I think) and this was already there. Turns out it's basically everywhere. I removed them in all the places where it's redundant.

I suppose we could build for aarch64 on linux (and maybe windows), but since we can't run the tests, this is of limited utility. We should look into arm runners or qemu running in the future.

Cf. https://chat.google.com/room/AAQA6lfp4Zw/TFaQnhrdNLE/TFaQnhrdNLE

.github/workflows/traits-build.yml
```diff
@@ -0,0 +1,110 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        bits: [32, 64]
+        os:
+          - macos-13 # Intel mac
+          - macos-latest # macos-14 m1
```
franziskuskiefer: This is actually macos-15 since yesterday.

.github/workflows/traits-build.yml
```diff
@@ -0,0 +1,110 @@
...
+      # - name: ⬆ Upload build
+      #   uses: ./.github/actions/upload_artifacts
+      #   with:
+      #     name: build_${{ matrix.os }}_${{ matrix.bits }}
+
+      # We get false positives here.
+      # TODO: Figure out what is going on here
+      # - name: 🏃🏻 Asan Linux
+      #   if: ${{ matrix.bits == 64 && matrix.os == 'ubuntu-latest' }}
+      #   run: RUSTDOCFLAGS=-Zsanitizer=address RUSTFLAGS=-Zsanitizer=address cargo +nightly test --release --target x86_64-unknown-linux-gnu
```
franziskuskiefer: snip?

chacha20poly1305/src/impl_aead_trait.rs
franziskuskiefer, line 1: I'm not keen on the `impl_` prefix in file names. But fine with me for now.

traits/src/aead/arrayref.rs
franziskuskiefer: Yeah, I thought behind a `cfg(features = "rand")`. But I'm fine without for now. For the external API then we should probably offer it.

traits/src/aead/arrayref.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// An Authenticated Encryption with Associated Data (AEAD) scheme. This trait
+/// is low-level and is mostly used for implementing other, more usable APIs.
+///
+/// Some implementors of this trait may impose stronger restrictions on the inputs than described
+/// here. Check the documentation of the types implementing this trait to make sure which inputs
+/// are valid.
+pub trait Aead<const KEY_LEN: usize, const TAG_LEN: usize, const NONCE_LEN: usize> {
+    /// Encrypt a plaintext message, producing a ciphertext and an authentication tag.
+    /// The arguments `plaintext` and `ciphertext` must have the same length.
```
franziskuskiefer, line 13: Should it really have the same length or would it be ok to allow something that's long enough, i.e. might be longer?

traits/src/aead/owned.rs
```diff
@@ -0,0 +1,104 @@
...
+/// Some implementors of this trait may impose stronger restrictions on the inputs than described
+/// here. Check the documentation of the types implementing this trait to make sure which inputs
+/// are valid.
+pub trait Aead<const KEY_LEN: usize, const TAG_LEN: usize, const NONCE_LEN: usize> {
+    /// Encrypt a plaintext message, producing a ciphertext and an authentication tag.
+    fn encrypt<const MSG_LEN: usize>(
+        key: &[u8; KEY_LEN],
+        nonce: &[u8; NONCE_LEN],
+        aad: &[u8],
+        plaintext: &[u8; MSG_LEN],
```
franziskuskiefer, line 18: Using an array for the msg/ctxt, this is severely limited in size (it must fit on the stack). Are we sure this API makes sense?

traits/src/aead/owned.rs
```diff
@@ -0,0 +1,104 @@
...
+    ) -> Result<[u8; MSG_LEN], DecryptError>;
+
+    /// Decrypt a ciphertext, verifying its authenticity.
+    fn decrypt_to_vec(
+        key: &[u8; KEY_LEN],
+        nonce: &[u8; NONCE_LEN],
+        aad: &[u8],
+        ciphertext: &[u8],
+        tag: &[u8; TAG_LEN],
+    ) -> Result<alloc::vec::Vec<u8>, DecryptError>;
```
franziskuskiefer, line 47: Should `alloc` go behind a feature?

traits/src/aead/slice.rs
```diff
@@ -0,0 +1,206 @@
...
+pub enum EncryptError {
+    /// The ciphertext buffer has the wrong length.
+    WrongCiphertextLength,
+    /// The plaintext is too long for this algorithm or implementation.
+    PlaintextTooLong,
+    /// The AAD is too long for this algorithm or implementation.
+    AadTooLong,
+    WrongKeyLength,
+    WrongTagLength,
+    WrongNonceLength,
```
franziskuskiefer: These should get some doc comment as well.

traits/src/aead/slice.rs
```diff
@@ -0,0 +1,206 @@
...
+    InvalidTag,
+    /// The plaintext buffer has the wrong length.
+    WrongPlaintextLength,
+    /// The plaintext is too long for this algorithm or implementation.
+    PlaintextTooLong,
+    /// The AAD is too long for this algorithm or implementation.
+    AadTooLong,
+    WrongKeyLength,
+    WrongTagLength,
+    WrongNonceLength,
```
franziskuskiefer: These as well.

secrets/src/int/public_integers.rs
```diff
@@ -56,9 +56,35 @@ impl<'a, T> DeclassifyRef for &'a T {
     }
 }
 
+// TODO: hax doesn't like this, so comment this out for now
```
franziskuskiefer: If we don't need it in hax processed code, we could just put it behind a `[cfg(not(hax))]` for now.

franziskuskiefer (approved):
lgtm

I left a couple comments, some of them would be nice to address, but some might also go into follow ups or discussions for the next round.

traits/src/aead/owned.rs
keks: You can turn any slice of the correct length into an array reference using TryInto. It does not need to be on the stack.

traits/src/aead/arrayref.rs
keks: I asked myself the same question and am not sure. I am happy to change the semantics here.

.github/workflows/traits-build.yml
keks: I copied this from another workflow and thought we might want to enable this here as well once we figured out the issue, so I kept it. Happy to remove it, though.

traits/src/aead/owned.rs
keks: Ah yeah, makes sense.

traits/src/aead/owned.rs
keks: It could also be a boxed array, for that matter.

keks: Sorry yeah that doesn't make sense. Not sure how that happened

keks: (I don't think the issue is that it must fit on the stack; I think the issue is that we often don't know the message lenght at compile time)

keks: Ahh right. So the idea here is that this version offers two ways to be used: For stuff like protocol messages where (a) you know the message size at compile time and (b) messages typically are quite small, you can use the on-stack owned version. These are heap-alloc free (and thus faster and available in no-std and no-alloc) and in some cases a bit more convenient than first initializing an array and then writing into it.

For cases where the length is not know, we can't return an owned dynamic value anyway, so we return a Vec, which requires alloc. So this offers the best solutions for both settings. We could say that these are too many options, not sure. But so far we did offer owned versions, so I wanted to stick with the pattern and later revise the combinations we provide.

But I'll include the concern about too large stack values in the doc.

keks: I al leaving this for now. Let's discuss whether this makes sense once we finish the current round.

traits/src/aead/arrayref.rs
keks: Let's leave this as-is for now. I left a note in the doc for a future redesign.

chacha20poly1305/src/impl_aead_trait.rs
keks: I left this for now. This would also touch upon the impl_hacl pattern we have all over the repo, which seemed like was too much for a drive-by fix.

franziskuskiefer:
Is this ready to get merged?

PR #1058, opened 2025-07-16
[CI] Add workflow to check `no_std` compatibility for some crates

### LINKED ISSUE #977: Add `no_std` workflow
Add a CI workflow that checks `no_std` compatibility for crates that support it

wysiwys:
Resolves #977

### DIFF
--- a/.github/workflows/no-std-check.yml
+++ b/.github/workflows/no-std-check.yml
@@ -15,8 +15,6 @@ jobs:
       - name: Install toolchain
         run: rustup target add thumbv7em-none-eabihf
 
-      - name: config RUST_TARGET_FLAG
-
       - name: Check no_std compatibility
         run: |
           ./no-std-check.sh
--- a/.github/workflows/no-std-check.yml
+++ b/.github/workflows/no-std-check.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  - no-std-check:
+  no-std-check:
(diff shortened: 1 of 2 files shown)


.github/workflows/no-std-check.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Install toolchain
+        run: rustup target add thumbv7em-none-eabihf
+
+      - name: config RUST_TARGET_FLAG
+        run: echo "RUST_TARGET_FLAG=--target=thumbv7em-none-eabihf" > $GITHUB_ENV
+
+      - name: Check no_std compatibility for crates that support it by default
+        run: |
+          cargo build -p libcrux-chacha20poly1305 \
+          -p libcrux-curve25519 \
```
franziskuskiefer: I wonder if there's a better way to maintain this list than in the action here. Maybe we should just put this in a bash script and run that here? Then it's easier to run locally as well.

wysiwys: :+1:  I moved the `no_std` check logic into a bash script (`no-std-check.sh`). That makes it easier to test the `no_std` check locally.

Would it be better for this bash script to be located in the workspace root directory or in another directory?

franziskuskiefer: Workspace root is fine. At some point we should be able to drop this and just call it for everything in the workspace.

franziskuskiefer (approved):
Thanks

PR #1061, opened 2025-07-16
Add `std` feature gate for `libcrux-ecdsa`

### LINKED ISSUE #983: `std` feature gate for `libcrux-ecdsa`
Currently, `libcrux-ecdsa` depends on `rand` with its default features enabled, which include `rand/std`. Let's feature-gate this behind an `std` feature for `libcrux-ecdsa`.

wysiwys:
This PR adds an `std` feature to the `libcrux-ecdsa` crate. This places the dependencies of the `libcrux-ecdsa` crate that require `std` behind the `std` feature flag.

This PR also makes these related changes:
- include this new `std` feature in the `default` features list for this crate, so that, as before, `libcrux-ecdsa` depends by default on the `rand` crate, with `rand/std` enabled
- add `rand_core/os_rng` to the dev dependencies for this crate. This is the minimum dependency needed to run a test that uses `rand_core::OsRng` (or the version re-exported by the `rand` crate). 

Resolves #983

### DIFF
--- a/ecdsa/Cargo.toml
+++ b/ecdsa/Cargo.toml
@@ -23,5 +23,6 @@ rand = ["dep:rand"]
 std = ["rand?/std"]
 
 [dev-dependencies]
+rand_core = { version = "0.9" , features = ["os_rng"] }
 serde = { version = "1.0.217", features = ["derive"] }
 serde_json = "1.0.138"
--- a/ecdsa/Cargo.toml
+++ b/ecdsa/Cargo.toml
@@ -15,11 +15,12 @@ libcrux-p256 = { version = "=0.0.3", path = "../p256", features = [
     "expose-hacl",
 ] }
 libcrux-sha2 = { version = "=0.0.3", path = "../sha2" }
-rand = { version = "0.9", optional = true }
+rand = { version = "0.9", optional = true, default-features = false }
 
 [features]
-default = ["rand"]
+default = ["rand", "std"]
 rand = ["dep:rand"]
+std = ["rand?/std"]
 
 [dev-dependencies]
 serde = { version = "1.0.217", features = ["derive"] }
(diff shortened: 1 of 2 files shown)


ecdsa/Cargo.toml
```diff
@@ -15,12 +15,14 @@ libcrux-p256 = { version = "=0.0.3", path = "../p256", features = [
...
+rand = { version = "0.9", optional = true, default-features = false }
 
 [features]
-default = ["rand"]
+default = ["rand", "std"]
 rand = ["dep:rand"]
+std = ["rand?/std"]
 
 [dev-dependencies]
+rand_core = { version = "0.9" , features = ["os_rng"] }
```
franziskuskiefer, line 26: To avoid duplicating a dependency here, we could just depennd on `libcrux-ecdsa = {version ="*", features = ["rand"]}`

wysiwys: This change would not work as expected, because the `rand` feature for this crate does not include the `os_rng` feature by default, even via the `rand/std` feature. 

However, another way to avoid adding `rand_core` as a dev dependency would be for the `std` feature to include `rand?/default` instead of `rand?/std`.

On the other hand, we don't need to include these other `rand/default` features for just the `libcrux-ecdsa` library (and not the tests in this crate), which are all for including rngs ([link](https://docs.rs/crate/rand/0.9.1)).

wysiwys: I think since the `OsRng` is only needed for tests, it makes sense to include it separately, in the dev dependencies - although I can also see that it is cleaner to just depend on `rand/default` by default in this crate.

franziskuskiefer: ok, sounds good to me. Let's leave this as is.

PR #1072, opened 2025-07-21
Add ML-KEM extraction for eurydice testing

protz:
This restores the c/ directory to a working state, after an earlier attempt to remove the files in c/ from version control resulted in further additional breakage.

I am recapping here the various messages I posted to https://hacspec.zulipchat.com/#narrow/channel/433829-Circus/topic/circus-green/with/529839729 
- some of the c/h files in the libcrux-ml-kem/c directory were removed, but some remained: I do not understand why this was done, and removed everything
- the benches and tests in c/ were replaced with a symlink to the benches and tests from cg/ -- this does not work as the code has different APIs; again, I do not understand why this was done and reverted to the previous c-only benchmarks
- the cmakelists, similarly, changed, but I do not understand why this was done; I also reverted this

There was one additional change to account for a file that was renamed, but that's about it.

This is not my preferred outcome -- what I would prefer is for files to remain under version control in c/ so that I can assess the impact of my changes via `git diff`, and for c/ to be ignored for all intents and purpose by libcrux CI. (Meaning, no CI failure if the shape of the code changes -- this is simply ignored.) But this part can be discussed in a separate issue.


karthikbhargavan:
Thanks for the PR, I'll test it on our dev machine tomorrow and then add a review.
This should hopefully help us get to a new stable state on the extracted C code.

franziskuskiefer:
This is not what we want. The code in this directory works as intended.
In particular do we not want to duplicate code, or keep unused C files around.

franziskuskiefer (changes requested):
In order to keep testing other variants of the mlkem C extraction, you should use the eurydice config file in here.

franziskuskiefer:
Closing this in favour of moving this to eurydice, as discussed in #1096.

Issue #1074, opened 2025-07-22
hax: use pinned hax instead of latest hax


W95Psp:
### Situation
Currently, libcrux relies on:
 - the Rust library `hax-lib`, pinned on a version on `crates.io`
 - the hax binary tool, built from the HEAD of branch `main` from the git repository 

The hax toolchain is coupled with the library: generally hax at commit `H` requires the crate it extracts to use `hax-lib` at commit `H`.
When hax is used with an outdated `hax-lib`, extraction or proofs may break.

### Solution
Instead of using the latest hax binary from main, use hax from the commit at which `hax-lib` is pinned in the `Cargo.toml` of libcrux.

More precisely, running `cargo metadata --format-version=1 | jq -r '.packages[] | select(.name=="hax-lib") | .version'`, we can get the version number of `hax-lib`, then build `hax` from the git checkout of hax repo with the tag corresponding to the version.

W95Psp:
Related to https://github.com/cryspen/hax/issues/1279.

franziskuskiefer:
This is blocked on working hax releases. #1075 probably has everything we want on this side already.

PR #1075, opened 2025-07-22
Use `hax` version corresponding to crate's `hax-lib` version

wysiwys:
This pull request adds a reusable workflow that retrieves the tag of the `hax-lib` dependency used in the libcrux crate. 

This reusable workflow is used to retrieve the correct version of the `hax` tool in the hax CI workflows.

Resolves https://github.com/cryspen/libcrux/issues/1074

### DIFF
--- a/.github/workflows/mldsa-hax.yml
+++ b/.github/workflows/mldsa-hax.yml
@@ -26,6 +26,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # NOTE: This always runs, even if the `hax_ref` workflow input is provided.
+  # TODO: Move this reusable workflow to a standalone action
   get-hax-ref:
     uses: ./.github/workflows/get-hax-ref.yml
 
--- a/.github/workflows/mlkem-hax.yml
+++ b/.github/workflows/mlkem-hax.yml
@@ -26,6 +26,8 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # NOTE: This always runs, even if the `hax_ref` workflow input is provided.
+  # TODO: Move this reusable workflow to a standalone action
   get-hax-ref:
     uses: ./.github/workflows/get-hax-ref.yml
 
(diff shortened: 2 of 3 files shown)


parrot7483:
CI seams to have the same problem as [this issue](https://github.com/cryspen/hax/issues/1554). Some missing _super expression.

franziskuskiefer:
@wysiwys can we get this updated to use the latest releases? Then we can get this in.

wysiwys:
@franziskuskiefer the workflows in this PR are already configured to use the `hax-lib` version in used in the `Cargo.toml` for the workspace, so these workflows will already use the latest `hax-lib` releases. 

Should I mark this PR as ready for review? The next step would be to rework the reusable workflow and add it to [cryspen/actions](https://github.com/cryspen/actions/), but since the implementation in this PR is ready to use, I think it would make sense to PR this first, and rework its functionality into a standalone action at a later point in time.

franziskuskiefer:
Yes, let's make this read to review then.

I manually triggered the workflows to run the proofs as well.
- [ML-KEM](https://github.com/cryspen/libcrux/actions/runs/16777092113)
- [ML-DSA](https://github.com/cryspen/libcrux/actions/runs/16777086445)

Currently mlkem proofs are broken on main.

.github/workflows/mldsa-hax.yml
```diff
@@ -12,10 +12,6 @@ on:
     - cron: "0 0 * * *"
 
   workflow_dispatch:
-    inputs:
```
franziskuskiefer, line 15: We should keep this input for testing.

franziskuskiefer (approved):
One nit. Otherwise lgtm

.github/workflows/mldsa-hax.yml
wysiwys: I added this workflow_dispatch input back in, so we can now use this input for testing again. One thing to note about this is that now, the `get-hax-ref` job will run even if the workflow_dispatch input is provided (although it will be ignored in this case). 

This is done because conditionally running it would complicate the workflow logic and require testing that the workflows complete successfully on the merge queue. Additionally, since we plan to use a standalone action for this behavior later, those changes would no longer be needed. I added a note about it in the workflows.

franziskuskiefer (approved):

PR #1076, opened 2025-07-22
Make file improvement

Parrot7483:
This PR does two things: 
- Make the usage of automatically finding F* modules optional. Useful while debugging when one ones to set includes explicitly. 
- Make the 'clean' command .PHONY and prevent HAX from extracting to F* when one only wants to clean.


franziskuskiefer:
@karthikbhargavan what's the state here?

parrot7483:
Was merged here https://github.com/cryspen/libcrux/pull/1101

Issue #1039, opened 2025-06-30
`Digest` trait


jschneider-bensch:
Should be implemented in
- `sha2`
- `libcrux-sha3`
- `blake2`

wysiwys:
I added an initial implementation of the `Digest` traits to this branch, with `arrayref` and `slice` APIs, implemented for `sha2`: 
- https://github.com/cryspen/libcrux/compare/wysiwys/digest-trait?expand=1

wysiwys:
This branch has now been updated with an initial draft of `blake2`, `sha2`, and `sha3` updated to use the new `Digest` traits. Some considerations:
- At this point in time, the `blake2` crate uses the same struct type to represent keyed and unkeyed hasher implementations. However, we don't yet apply the new trait APIs to the keyed versions. Because of this, it may be cleaner to use a separate type for each.
- Currently, there are separate traits for oneshot and incremental `Digest` APIs, neither of which requires the other. It would be good to investigate in which cases it makes sense to blanket impl the oneshot trait for the incremental trait.

wysiwys:
Note on digest trait APIs: The old `libcrux_traits::Digest` trait has a lot of overlap with the new trait APIs, and is only currently implemented in the `sha2` crate. Once the new trait APIs are more settled, it would make sense to look at whether the old Digest trait is still needed.

PR #1078, opened 2025-07-23
Wysiwys/digest trait

wysiwys:
This PR is an initial implementation of the new libcrux digest traits. The approach is based on the approach taken for the KEM and AEAD traits. 

- `libcrux-blake2`
- `libcrux-sha2`
- `libcrux-sha3`

Two main trait types are defined: `DigestIncremental` and `Hash`. Each of the `arrayref`, `owned`, and `slice` modules defines two traits of this kind. A blanket implementation is used to implement the `owned` traits for structs that implement the `arrayref` traits. A convenience macro is provided to implement the `slice` traits.

Notably, this implementation currently includes convenience `Hasher` structs in `libcrux_traits::digest`. These structs utilize the underlying digest traits to provide a more convenient interface for the hashing functionality. However, it is currently unclear whether this approach is suitable, since the requirements of the `libcrux-blake2`, `libcrux-sha2`, and `libcrux-sha3` crates differ quite a bit. Specifically for `libcrux-blake2`'s variable digest size hasher, the `arrayref` traits are not implemented, meaning a convenience `Hasher` type needs to be flexible enough to allow not all of the digest traits to be implemented. Currently, this is addressed by providing a separate hasher type for this case.

Other questions include:
- Blake2: This crate currently uses a shared struct for both keyed and unkeyed hashers. Should the unkeyed functionality be split out? 
- Currently, the `libcrux-blake2` crate does not provide a oneshot API. Is this something that should be automatically derived from the incremental API?


keks (changes requested):
Clara and I met yesterday to take a look at the design together and found some things that could be improved:

In order to be able to have just a single hasher struct, extract the associated type `IntermediateState` from the other traits into a base trait that the others extend. In the struct definition, constrain only to that. In the impls, constrain to the specific trait.

Blake2 supports generating hashes of any length (more or less). While this is interesting functionality, let's not overload the semantics of the plain Digest traits here. So far the slice-based version used the length of the provided mut slice as an indication for how long the output hash should be. This can be problematic if the user accidentally passes in a mut slice of the wrong length and we end up with e.g. a 4 byte key.
To prevent this, we slightly refactor the way this is implemented:
1. There is just one struct per hash (Blake2s, Blake2b), with a const generic param for the output length
2. The arrayref version is implemented just for the same length
3. The slice version checks that the output slice has the expected length and then calls the underlying implementation. We might be able to do this by converting it to a mut arrayref and use the same API as above.

There also were some small TODO comments, but these can be addressed once the design is stable.

wysiwys:
The Digest trait implementation is now updated to reflect the changes listed above. 

One additional change to this PR is that the `update()` and `reset()` methods are now part of the base Digest trait, rather than individually being part of the   `arrayref::DigestIncremental` and `slice::DigestIncremental` traits. 

I'll mark this PR as ready for review, as a first pass for the new Digest traits.

jschneider-bensch:
I promise, I'll take a look at this tomorrow morning!

traits/src/digest.rs
```diff
@@ -0,0 +1,75 @@
...
+    fn default() -> Self {
+        Self {
+            state: Default::default(),
+        }
+    }
+}
+
+impl<const N: usize, D: DigestBase + slice::Hash> Hasher<N, D> {
+    // TODO: rename
+    pub fn hash_slice(digest: &mut [u8], payload: &[u8]) -> Result<usize, slice::HashError> {
```
jschneider-bensch, line 74: Could you add doc comments, e.g. explaining the return type?

blake2/src/impl_digest_trait.rs
```diff
@@ -0,0 +1,77 @@
+use crate::impl_hacl::*;
+use libcrux_traits::digest::{arrayref, slice, DigestIncrementalBase, Hasher, UpdateError};
+
+macro_rules! impl_digest_traits {
+    ($out_size:ident, $type:ty, $blake2:ty, $hasher:ty) => {
+        impl<const $out_size: usize> DigestIncrementalBase for $type {
+            type IncrementalState = $blake2;
+
+            fn update(state: &mut Self::IncrementalState, chunk: &[u8]) -> Result<(), UpdateError> {
+                // XXX: maps all known errors returned by this function
```
keks: I don't think the XXX mark is needed here

```suggestion
                // maps all known errors returned by this function
```

keks (approved):
I think this look ready to merge after it's rebased. Just one tiny nit.

(oh, forgot to hit send on my review...)

keks (approved):
This looks good, let's get this in!

libcrux-ml-kem/c.sh
```diff
@@ -57,25 +57,30 @@ if [[ "$portable_only" = 1 ]]; then
     export LIBCRUX_DISABLE_SIMD128=1
 fi
 
+
+if [[ "$clean" = 1 ]]; then
+    cargo clean
```
franziskuskiefer, line 62: This was only cleaning C before. Let's note down somewhere that this is cleaning Rust now as well. This may lead to unexpected results because clean deletes C files that we may want to keep around.
Also, `boring.sh` cleans already itself. So that should just add clean then to this invocation.

But we can do all this in a follow up.

sha2/src/impl_digest_trait.rs
franziskuskiefer, line 1: Unrelated to C, but I still think that this naming scheme of `impl_*` is bad.

traits/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 
 extern crate alloc;
 
+// TODO: remove the old `Digest` trait defined here
```
franziskuskiefer: If this is actually something that should be done it should get an issue here, and a note on an issue (or a new one).

franziskuskiefer (approved):
Let's get this in.
But there seem to be some follow ups needed.

traits/src/lib.rs
wysiwys: Yes, it's not quite clear yet how to handle this trait. I added a note in #1039 and linked it in this comment.

sha2/src/impl_digest_trait.rs
wysiwys: I'll make a follow-up issue about this, since the current naming scheme is used by other trait implementations

wysiwys: Opened an issue here:
- https://github.com/cryspen/libcrux/issues/1107?issue=cryspen%7Clibcrux%7C1118

wysiwys:
This PR made some changes to the ML-KEM `c.sh` to prevent the script failing when charon returned successfully without producing an LLBC file. 

These changes were:
- In the charon step, extract `libcrux-secrets` before `libcrux-sha3`, instead of after
- When the `--clean` flag is provided to the script, run `cargo clean`,  in addition to cleaning up C files

The PR also ensures that the script removes all relevant `.llbc` files if they already exist (where before `./libcrux-sha3.llbc` and `./libcrux-ml-kem.llbc` were removed if they already exist, now `./libcrux-secrets.llbc` is also removed)

PR #1079, opened 2025-07-23
[PSQ] Use TLSCodec for Serialization/Deserialization in PSQv1

### LINKED ISSUE #925: Libcrux psq Default impl for Decoding `InitiatorMsg`
At the moment there is a bit of awkwardness working with the `libcrux_psq` crate generically across the supported KEM algorithms because there is no blanket implementation for the `Decode` trait. In order to accommodate this downstream libraries would need to add things like this in their usage:

```rust
impl<'p, C, T> PsqProtocol<'p, C, T>
where
    C: Authenticator,
    T: PSQ,
    InitiatorMsg<<T as PSQ>::InnerKEM>: Decode,
{ ... }
```

It would be nice to have a blanket implementation to cover the `Decode` trait like there is for `Encode`. 
For example, the `PSQ` trait could then modified and the `Decode` trait could be implemented for all `KEM`
types whose ciphertext implements `Decode`:  

```rust
pub trait PSQ: private::Seal {
    /// The underlying KEM.
    type InnerKEM: KEM<Ciphertext: Encode+Decode, SharedSecret: Encode, EncapsulationKey: Encode>;

	// ...
}

impl<T: KEM<Ciphertext: Decode>> Decode for InitiatorMsg<T> {
    fn decode(bytes: &[u8]) -> Result<(Self, usize), Error> {

        let (ct, n) = <T as KEM>::Ciphertext::decode(&bytes)?;
            // Ct::decode(libcrux_kem::Algorithm::X25519, &bytes[0..32]).unwrap();
        let mut mac = [0u8; MAC_LENGTH];
        mac.copy_from_slice(&bytes[n..n + MAC_LENGTH]);

        let (aead_mac, _len) = AeadMac::decode(&bytes[n + MAC_LENGTH..])?;

        let out = Self {
            encapsulation: Ciphertext::<T> {
                inner_ctxt: ct,
                mac,
            },
            aead_mac,
        };

        Ok((out, bytes.len()))
    }
}
```

The issue seems like it stems from some incongruity between:

```rust
// libcrux_kem::Ct
pub fn decode(alg: Algorithm, bytes: &[u8]) -> Result<Self, Error>

// libcrux_psq::Decode
pub trait Decode {
    // Required method
    fn decode(bytes: &[u8]) -> Result<(Self, usize), Error>
       where Self: Sized;
}
```

Where implementing `libcrux_psq::Decode` for `libcrux_kem::Ct` requires keeping track of the length of the decoded ciphertext, which might be better done by `Ct` itself either with a change to the `decode` constructor or by adding a `len` method to the `Ct` struct.

jschneider-bensch:
This PR replaces almost all ad hoc serialization and deserialization in PSQv1 with TLSCodec (the exception being the timestamp included in the initiator message).

To this end, I've introduced `codec` features in `libcrux-ecdh`, `ed25519` & `libcrux-kem`.

Fixes #925

### DIFF
--- a/libcrux-psq/src/classic_mceliece.rs
+++ b/libcrux-psq/src/classic_mceliece.rs
@@ -4,34 +4,68 @@
 //! `mceliece460896f`) as the underlying KEM.
 
 use classic_mceliece_rust::{
-    decapsulate_boxed, encapsulate_boxed, keypair_boxed, Ciphertext, PublicKey, SecretKey,
-    SharedSecret,
+    decapsulate_boxed, encapsulate_boxed, keypair_boxed, Ciphertext as Ct, PublicKey as Pk,
+    SecretKey, SharedSecret as Ss,
 };
 
 use libcrux_traits::kem::{KEMError, KeyPair, KEM};
+use tls_codec::{Deserialize, Serialize, Size};
 
 use crate::traits::*;
 
+/// A wrapper around the `classic_mceliece_rust` type `Ciphertext`.
+pub struct Ciphertext(Ct);
+impl Serialize for Ciphertext {
+    fn tls_serialize<W: std::io::Write>(&self, writer: &mut W) -> Result<usize, tls_codec::Error> {
+        todo!()
+    }
+}
+
+impl Deserialize for Ciphertext {
+    fn tls_deserialize<R: std::io::Read>(bytes: &mut R) -> Result<Self, tls_codec::Error>
+    where
+        Self: Sized,
+    {
+        todo!()
+    }
+}
+
+impl Size for Ciphertext {
+    fn tls_serialized_len(&self) -> usize {
+        todo!()
+    }
+}
+
+/// A wrapper around the `classic_mceliece_rust` type `PublicKey`.
+pub struct PublicKey<'a>(Pk<'a>);
+
+impl<'a> Size for PublicKey<'a> {
+    fn tls_serialized_len(&self) -> usize {
+        todo!()
+    }
+}
+impl<'a> Serialize for PublicKey<'a> {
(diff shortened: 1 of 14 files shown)


libcrux-psq/src/classic_mceliece.rs
```diff
@@ -4,35 +4,78 @@
...
+        Self: Sized,
+    {
+        let bytes_deserialized = VLBytes::tls_deserialize(bytes)?;
+        // XXX: This is the expected length of the ciphertext for mceliece460896f.
+        // If we didn't want to hardcode this, we'd have to pull in `generic-array` as a dependency and do something like
+        // ```
+        // let array = GenericArray<u8, Ciphertext::EncappedKeySize>::from(bytes_deserialized);
+        // let ciphertext =  Ciphertext::from_bytes(&array)?;
+        // ```
+        Ok(Ciphertext(Ct::from(<[u8; 156]>::try_from(
```
franziskuskiefer: Maybe pull the length out in a constant to make it even more obvious that this is a hard coded length.

libcrux-psq/src/impls.rs
```diff
@@ -8,7 +8,8 @@
 //!   KEM *does not provide post-quantum security*. We include it for
 //!   testing and benchmarking.
 
-use crate::traits::*;
+#[allow(unused_imports)] // XXX: We do use it in the macro below.
+use crate::traits::PSQ;
```
franziskuskiefer: This is only used in the test. That's why the linter complains.

franziskuskiefer (approved):
Thanks! This looks so much cleaner.
Just two nits.

libcrux-psq/src/impls.rs
jschneider-bensch: Right, I've moved it to the tests module.

PR #1080, opened 2025-07-23
Signature traits

### LINKED ISSUE #1034: `Signature` APIs
Should be implemented in
- [ ] `ed25519`
- [ ] `libcrux-ml-dsa`
- [ ] `ecdsa`
- [ ] `rsa-pss`

wysiwys:
This PR implements the Signature traits.
- `keygen()`
- `sign()`
- `verify()`

Resolves #1034 

Includes implementations for:
- [x] `ed25519` (`arrayref`, `slice`, `owned`)
- [x] `libcrux-ml-dsa` (`owned`)
- [x] `ecdsa` (`arrayref`, `slice`, `owned`)

### DIFF
--- a/libcrux-ml-dsa/src/impl_signature_trait.rs
+++ b/libcrux-ml-dsa/src/impl_signature_trait.rs
@@ -104,6 +104,28 @@ pub mod signers {
     }
     pub use impl_context;
 
+    impl From<crate::types::SigningError> for arrayref::SignError {
+        fn from(err: crate::types::SigningError) -> Self {
+            use crate::types::SigningError::*;
+            match err {
+                RejectionSamplingError => arrayref::SignError::InvalidRandomness,
+                ContextTooLongError => arrayref::SignError::InvalidArgument,
+            }
+        }
+    }
+    impl From<crate::types::VerificationError> for arrayref::VerifyError {
+        fn from(err: crate::types::VerificationError) -> Self {
+            use crate::types::VerificationError::*;
+
+            match err {
+                MalformedHintError
+                | SignerResponseExceedsBoundError
+                | CommitmentHashesDontMatchError => arrayref::VerifyError::InvalidSignature,
+                VerificationContextTooLongError => arrayref::VerifyError::LibraryError,
+            }
+        }
+    }
+
     macro_rules! impl_signature_trait {
         ($module:ident, $name:tt) => {
             mod $module {
@@ -147,13 +169,7 @@ pub mod signers {
                             randomness.declassify(),
                             signature,
                         )
-                        .map_err(|err| {
-                            use crate::types::SigningError::*;
-                            match err {
-                                RejectionSamplingError => arrayref::SignError::InvalidRandomness,
-                                ContextTooLongError => arrayref::SignError::InvalidArgument,
-                            }
-                        })
+                        .map_err(arrayref::SignError::from)
                     }
                     /// Verify a signature using a provided verification key and context.
                     fn verify(
@@ -167,18 +183,7 @@ pub mod signers {
                             T::context(),
                             signature,
                         )
-                        .map_err(|err| {
-                            use crate::types::VerificationError::*;
-
-                            match err {
-                                MalformedHintError
-                                | SignerResponseExceedsBoundError
-                                | CommitmentHashesDontMatchError => {
-                                    arrayref::VerifyError::InvalidSignature
-                                }
-                                VerificationContextTooLongError => arrayref::VerifyError::LibraryError,
-                            }
-                        })
+                        .map_err(arrayref::VerifyError::from)
                     }
 
                     fn keygen_derand(


ecdsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,65 @@
...
+        pub type $alias = Signer<libcrux_sha2::$digest_alg_name>;
+
+        impl arrayref::Sign<&Nonce, $sk_len, $sig_len> for $alias {
+            fn sign(
+                payload: &[u8],
+                private_key: &[u8; $sk_len],
+                signature: &mut [u8; $sig_len],
+                nonce: &Nonce,
+            ) -> Result<(), arrayref::SignError> {
+                crate::p256::_sign_internal(
```
keks: Hm, _sign_internal is not inlined, maybe call the specific functions directly? e.g. ecdsa_sign_p256_sha2

ecdsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,65 @@
...
+        }
+        impl arrayref::Verify<&(), $pk_len, $sig_len> for $alias {
+            fn verify(
+                payload: &[u8],
+                public_key: &[u8; $pk_len],
+                signature: &[u8; $sig_len],
+                _aux: &(),
+            ) -> Result<(), arrayref::VerifyError> {
+
+                crate::p256::_verify_internal(
```
keks: same here

ecdsa/src/impl_signature_trait.rs
keks: or maybe inline it. I feel like the matching is unnecessary indirection, but in release mode it should be optimized out. Still feels weird to have it.

ed25519/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,53 @@
+use libcrux_traits::signature::arrayref::{Sign, SignError, Verify, VerifyError};
+
+pub struct Signer;
+
+const PUBLIC_KEY_LEN: usize = 32;
+const PRIVATE_KEY_LEN: usize = 32;
+const SIGNATURE_LEN: usize = 64;
+
+impl Sign<&(), PRIVATE_KEY_LEN, SIGNATURE_LEN> for Signer {
```
keks: I think we probably should pass in () instead of &(). The former is truly zero-size. Or am I missing a reason against this?

libcrux-ml-dsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,115 @@
...
+                }
+                crate::VerificationError::CommitmentHashesDontMatchError => {
+                    arrayref::VerifyError::LibraryError
+                }
+                crate::VerificationError::VerificationContextTooLongError => {
+                    arrayref::VerifyError::LibraryError
+                }
+            }
+        }
+    }
```
keks, line 104: These matches all return the same thing. We can just have the LibraryError as the body directly :)

traits/src/signature/arrayref.rs
```diff
@@ -0,0 +1,91 @@
+pub trait Sign<SignAux, const PRIVATE_KEY_LEN: usize, const SIGNATURE_LEN: usize> {
+    // TODO: should this be called `private_key` or `sk`?
```
keks: In general I prefer using "signing key" and "verification key" (sk, vk), but I do not have super strong feelings. Maybe something we should decide once and stop bikeshedding :)

traits/src/signature/secrets.rs
```diff
@@ -0,0 +1,48 @@
...
+pub trait SignNoAux<const PRIVATE_KEY_LEN: usize, const SIGNATURE_LEN: usize> {
+    fn sign(
+        payload: &[u8],
+        private_key: &[U8; PRIVATE_KEY_LEN],
+    ) -> Result<[u8; SIGNATURE_LEN], SignError>;
+}
+
+impl<
+        'a,
+        const PRIVATE_KEY_LEN: usize,
```
keks: not borrowing the unit also means we can drop the lifetimes here I guess

traits/src/signature/slice.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum VerifyError {
+    InvalidPayloadLength,
+    InvalidPublicKeyLength,
+    InvalidSignatureBufferLength,
+    InvalidSignature,
+    LibraryError,
+}
```
keks, line 184: Both the types and the variants need comments. We should also impl core::fmt::Display for all errors.

traits/src/signature/arrayref.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum VerifyError {
+    InvalidSignature,
+    InvalidPayloadLength,
+    LibraryError,
+}
+
+// No auxiliary information
+pub trait SignNoAux<const PRIVATE_KEY_LEN: usize, const SIGNATURE_LEN: usize> {
```
keks: Hmm, I am not sure how this could be improved, but I think as a user I might be confused what  the "No  Aux" is about. One option would be to instead have "SignWithRandomness" and "SignWithRandomnessAndContext" but I am not sure that is any better :sweat_smile: 

We could also say that for now this API doesn't support signing with contexts an we just have the Sign and SignWithRandomness. This could also cover the salt and nonce.

What is the context, by the way? If it is a value that carries precomputed stuff, we could also think about a trait that has a receiver, so we can put that into the type that implements the trait.

libcrux-ml-dsa/src/impl_signature_trait.rs
wysiwys: Thanks, I removed these `From` implementations for now and just map the errors directly to `LibraryError`, to keep things simpler.

ecdsa/src/impl_signature_trait.rs
wysiwys: Agreed! I took out the indirection here, and called the specific sign and verify functions directly.

ed25519/src/impl_signature_trait.rs
wysiwys: Ah, thanks, I agree that it makes more sense to use the unit type here! It also helps simplify the blanket impls because they no longer need a lifetime for the `&()`.

traits/src/signature/arrayref.rs
wysiwys: Ah, I'd agree that something like `SIGNING_KEY` and `VERIFICATION_KEY` would be best to use here. Yes, let's unify the usage. I'll have a look and see if there's any reason to keep it as is; if not, I can just change the usage to these terms already.

traits/src/signature/arrayref.rs
wysiwys: I think using a receiver for the context would be a good way to avoid confusing or overlapping trait implementations. The auxiliary data argument in these traits is currently used to represent a wide variety of inputs, such as the `salt_len` (in verification in `libcrux-rsa`), the `nonce` (in signing in `libcrux-ecdsa`) and the `context` (in signing and verification in `libcrux-ml-dsa`). 

One possibility could be to create a separate `SignWithRandomness` trait, and then use the receiver for the auxiliary or context values that it makes sense to include, for both `Sign` and `SignWithRandomness` traits. This approach could be one way to avoid having too many overlapping traits, and also be better from an ergonomics perspective.

wysiwys:
Thanks for the review and feedback @keks! I addressed most of the feedback points, and added some comments for the other points. 

For the trait organization (specifically how to handle randomness and auxiliary data), I think there is room for improvement, too, and I think there is likely a combination of the different options from our discussion above that will address many of the concerns around ergonomics/complexity. I will take another look at the different combinations of input data we need to support, and look at the pros and cons of  different approaches.

traits/src/signature/arrayref.rs
jschneider-bensch: I also prefer "signing key" and "verification key" for signatures.

rsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,115 @@
+use super::impl_hacl::*;
+
+use libcrux_traits::signature::arrayref;
+
+// $bytes is pk_len, sk_len and sig_len
+macro_rules! impl_signature_trait {
+    ($bits:literal, $bytes:literal, $digest_alg:ident, $alias:ident) => {
+        #[allow(non_camel_case_types)]
+        // TODO: include in docs that this is an RSA signer?
```
jschneider-bensch: The docs could include a description of the `SignAux` type and its purpose.

libcrux-ml-dsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,93 @@
...
+                #[doc = concat!("An ", stringify!($module), " signer.")]
+                pub type $alias = super::Signer<$name>;
+
+                const VERIFICATION_KEY_LEN: usize =
+                    crate::ml_dsa_generic::$module::VERIFICATION_KEY_SIZE;
+                const SIGNING_KEY_LEN: usize = crate::ml_dsa_generic::$module::SIGNING_KEY_SIZE;
+                const SIGNATURE_LEN: usize = crate::ml_dsa_generic::$module::SIGNATURE_SIZE;
+
+                // XXX: implementing owned trait directly because there is no arrayref equivalent
+                // TODO: for docs, these should appear as consts in trait def
```
jschneider-bensch: Do you mean that the constants above should appear in the docs?

rsa/src/impl_signature_trait.rs
jschneider-bensch, line 1: The error conversions look sensible to me.

jschneider-bensch (commented):
I think the traits look good!

I have some comments:
- There are some `TODO`s left in different places. Are these for follow-ups, or do you want to address them in this PR?
- Could you try and add a generic test for some of the APIs, e.g. like with the KEM trait, s.t. we know the implementation works?
- Could you summarize in the PR description which scheme now implements which APIs? Or are they all covered via blanket impls?

traits/src/signature/arrayref.rs
wysiwys: Then let's go with "signing key" and "verification key". I went ahead have updated this PR to use these terms across the board.

libcrux-ml-dsa/src/impl_signature_trait.rs
wysiwys: In a way - I was looking for a way to make the constants in the rustdoc trait definition into hyperlinks to the definitions. However, it is not currently possible to generate rustdocs with that functionality, unfortunately.

wysiwys: I took out this TODO for now and will mark this conversation as resolved (although once the trait design is more settled, I think it would make sense to do another pass on the doc comments and rustdocs).

keks:
What I am wondering is whether our Signature traits should support all uses of context. We also don't support all the features of the blake2 hash (e.g. personalisation or keying of the hash).

We could pretty easily support the use of contexts known at compile time:

- Add a `trait Context { context() -> &'static [u8]; }` to the `mldsa` crate
- Then `impl<Ctx: Context> RandomizedSignature for MlDsa<Ctx>`, where we call the context function to get the value
- The context function should be marked inline(always). Maybe we can define a macro that defines a context struct and implements the trait for it, so the user doesn't forget to set it?

rsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,170 @@
...
+                    crate::DigestAlgorithm::$digest_alg,
+                    &signing_key.as_var_len(),
+                    payload,
+                    salt,
+                    signature,
+                )
+                .map_err(|e| match e {
+                    crate::Error::MessageTooLarge => arrayref::SignError::InvalidPayloadLength,
+                    _ => arrayref::SignError::LibraryError,
+
```
keks:
```suggestion
```

rsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,170 @@
...
+                    &VarLenPublicKey::try_from(verification_key.as_ref()).unwrap(),
+                    payload,
+                    salt_len,
+                    signature,
+                )
+                .map_err(|e| match e {
+                    crate::Error::InvalidSignatureLength => arrayref::VerifyError::InvalidSignatureBufferLength,
+                    crate::Error::MessageTooLarge => arrayref::VerifyError::InvalidPayloadLength,
+                    _ => arrayref::VerifyError::LibraryError,
+
```
keks:
```suggestion
```

rsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,170 @@
...
+                })
+            }
+        }
+
+        // manual implementation of sign slice trait
+        /// The [`mod@slice`] version of the Sign trait.
+        impl slice::Sign for $alias {
+
```
keks:
```suggestion
```

rsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,170 @@
...
+            type SigningKey<'a> = generic_keys::VarLenPrivateKey<'a, u8>;
+            /// Sign a payload using a provided signing key and `salt`.
+            fn sign(
+                payload: &[u8],
+                signing_key: generic_keys::VarLenPrivateKey<'_, u8>,
+                signature: &mut [u8],
+                salt: &[u8],
+            ) -> Result<(), slice::SignError> {
+                sign_varlen(
+
```
keks:
```suggestion
```

ecdsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,101 @@
...
+                );
+                if !result {
+                    return Err(arrayref::VerifyError::LibraryError);
+                }
+                Ok(())
+            }
+        }
+        libcrux_traits::impl_signature_slice_trait!($alias => SIGNING_KEY_LEN, SIG_LEN, &Nonce, nonce, &'a [u8; SIGNING_KEY_LEN]);
+        libcrux_traits::impl_verify_slice_trait!($alias => VERIFICATION_KEY_LEN, SIG_LEN, (), _aux);
+        // TODO: owned and secrets traits not appearing in docs
```
keks: What do you mean with this TODO? Does the impl not show up or the trait definitions? Is that resolved?

ecdsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,101 @@
...
+    //! [`libcrux_traits::signature`] APIs.
+
+    use libcrux_traits::signature::arrayref;
+
+    const SIG_LEN: usize = 64;
+    const VERIFICATION_KEY_LEN: usize = 64;
+    const SIGNING_KEY_LEN: usize = 32;
+
+    macro_rules! impl_signature_trait {
+    ($digest_alg_name:ident, $alias:ident, $sign_fn:ident, $verify_fn:ident) => {
```
keks: I think the indentiation here is a bit off

ed25519/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,62 @@
...
+                payload
+                    .len()
+                    .try_into()
+                    .map_err(|_| SignError::InvalidPayloadLength)?,
+                payload,
+            );
+
+            Ok(())
+        }
+    }
```
keks:
```suggestion
    }
    
```

libcrux-ml-dsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,100 @@
...
+                            verification_key,
+                            payload,
+                            context,
+                            signature,
+                        )
+                        .map_err(|_| owned::VerifyError::LibraryError)
+                    }
+                }
+
+                // TODO: secrets trait not appearing in docs
```
keks: same here

traits/src/signature/slice.rs
```diff
@@ -0,0 +1,208 @@
...
+            super::arrayref::VerifyError::InvalidSignatureBufferLength => {
+                Self::InvalidSignatureBufferLength
+            }
+            super::arrayref::VerifyError::InvalidPayloadLength => Self::InvalidPayloadLength,
+            super::arrayref::VerifyError::LibraryError => Self::LibraryError,
+        }
+    }
+}
+
+/// Implements [`Sign`] for any [`arrayref::Sign`](crate::signature::arrayref::Sign) where the
```
jschneider-bensch: The last part of this comment isn't true anymore, right?

traits/src/signature/slice.rs
```diff
@@ -0,0 +1,208 @@
...
+                .map_err($crate::signature::slice::SignError::from)
+            }
+        }
+    };
+}
+/// Implements [`Verify`] for any [`arrayref::Verify`](crate::signature::arrayref::Verify)
+#[macro_export]
+macro_rules! impl_verify_slice_trait {
+    ($type:ty => $vk_len:expr, $sig_len:expr, $verify_aux:ty, $verify_aux_param:tt) => {
+        /// The [`mod@slice`] version of the Verify trait.
```
jschneider-bensch: This doesn't link correctly for me. Maybe something like `[slice](libcrux_traits::signature::slice)` would work?

rsa/src/impl_hacl.rs
```diff
@@ -2,17 +2,39 @@ use crate::Error;
...
+    use super::*;
+
+    /// A generic RSA Private Key that is `LEN` bytes long.
+    #[repr(C)]
+    pub struct PrivateKey<const LEN: usize, PrivateKeyByte> {
+        pub(crate) pk: PublicKey<LEN>,
+        pub(crate) d: [PrivateKeyByte; LEN],
+    }
+    /// An RSA Private Key backed by slices. Use if the length is not known at compile time.
+    pub struct VarLenPrivateKey<'a, PrivateKeyByte> {
```
keks: Maybe this should also be repr(C)?

rsa/src/impl_hacl.rs
```diff
@@ -2,17 +2,39 @@ use crate::Error;
...
+        pub(crate) pk: VarLenPublicKey<'a>,
+        pub(crate) d: &'a [PrivateKeyByte],
+    }
```
keks: (disclaimer: just want to clarify what I meant earlier, not that anything needs to be changed here. This seems to work.)

What I had meant was something like
```rust
struct GenericSigningKey<PK, D> {
  pk: PK,
  d: D,
}

type PrivateKey<const LEN: usize> = GenericPrivateKey<PublicKey<LEN>, [u8; LEN]>;
type ClassifiedPrivateKey<const LEN: usize> = GenericPrivateKey<PublicKey<LEN>, [U8; LEN]>;
type VarLenPrivateKey = GenericPrivateKey<VarLenPublicKey, [u8]>;
// ...

// or even

type PrivateKey<const LEN: usize> = GenericPrivateKey<[u8; LEN], [u8; LEN]>;
type ClassifiedPrivateKey<const LEN: usize> = GenericPrivateKey<[u8; LEN], [U8; LEN]>;
type VarLenPrivateKey<'a> = GenericPrivateKey<&'a [u8], &'a [u8]>;
// ...
```

What you have seems to work though, so unless you think this is somehow better, just leave it as is.

keks (dismissed):
Nice, even RSA is looking better and better :)

I noticed some files had weird whitespace. Did you run rustfmt? I also found a bunch of TODOs that note that something doesn't show up in documentation. What exactly isn't showing up? The comments read like the trait doesn't show up, but that makes sense because the block only contains the impl.

traits/src/signature/slice.rs
wysiwys: Thanks for catching that! I updated the link to the suggestion. Could you let me know if the link works correctly on your side now?

rsa/src/impl_signature_trait.rs
wysiwys: I ran rustfmt but it is not fixing these formatting issues. I'll enable `format_macro_bodies` and see if that fixes the formatting

rsa/src/impl_hacl.rs
wysiwys: The `VarLenPrivateKey` currently doesn't strictly need to be repr(C), since the Declassify implementation is safe (it does not use `std::mem::transmute`). The declassification looks like the below:

```rust
fn declassify(self) -> Self::Declassified {
    use libcrux_secrets::DeclassifyRef;
    generic_keys::VarLenPrivateKey {
        pk: self.pk,
        d: self.d.declassify_ref(),
    }
}

```
I moved the `libcrux_secrets` trait implementations for this and the `PrivateKey` into the `generic_keys` module, so it's more convenient to view the structs alongside the trait implementations. It also wouldn't hurt to make the `VarLenPrivateKey` repr(C), but it doesn't look like there's an advantage to doing so right now.

ecdsa/src/impl_signature_trait.rs
wysiwys: Ah, yeah, so this is about specifically the rustdocs for each `$alias` type alias. It would be useful if the rustdocs contained all of the traits implemented for the `$alias`, but this does not happen by default.

A bit more detail: the traits that are explicitly implemented on the `$alias` (which in this case are the arrayref and slice traits) appear in the rustdocs for the `$alias`, but the traits that are implemented by blanket impls (in this case, these are the owned and secrets traits) don't appear there, although they do appear on the rustdoc page for the underlying type that is aliased.

rsa/src/impl_signature_trait.rs
wysiwys: Enabling this feature for `cargo fmt` did not fix these formatting issues, either, so I did a manual formatting pass over the crates, to clean up some of the whitespace and indentation.

franziskuskiefer:
Converting this to a draft for now. See #1107.

rsa/src/impl_hacl.rs
wysiwys: I'm marking this thread as resolved, since the traits are no longer implemented for RSA.

libcrux-ml-dsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    // List structs in doc comment if enabled.
+    /// A convenience struct for signature scheme functionality.
+    #[cfg_attr(feature = "mldsa44", doc = "\n - [`MlDsa44`]")]
+    #[cfg_attr(feature = "mldsa65", doc = "\n - [`MlDsa65`]")]
+    #[cfg_attr(feature = "mldsa87", doc = "\n - [`MlDsa87`]")]
+    pub struct Signer<Implementation> {
+        _phantom_data: core::marker::PhantomData<Implementation>,
+    }
```
keks: What is the reason for this wrapper struct? Could we not just define the individual structs and implement the Sign/Verify traits on them directly?

libcrux-ml-dsa/src/impl_signature_trait.rs
keks, line 1: I am not sure this is the right time to do this, but we could make `Signer` or the per-algorithm struct generic over a type that implements a new `trait Context { context()->&'static [u8]; }` and use that as the context.

traits/src/signature/arrayref.rs
```diff
@@ -0,0 +1,92 @@
...
+    type VerifyAux<'a>;
+    /// Verify a signature using a provided verification key. Required auxiliary information is provided using
+    /// the `aux` argument.
+    fn verify(
+        payload: &[u8],
+        verification_key: &[u8; VERIFICATION_KEY_LEN],
+        signature: &[u8; SIGNATURE_LEN],
+        aux: Self::VerifyAux<'_>,
+    ) -> Result<(), VerifyError>;
+}
```
keks: Why did you decide to split this into two traits? I think this is a valid way to design the traits, but it deviates from the traits we already have.

keks (commented):
I think this looks pretty good! I have a few questions on some design details that I would like to discuss before we proceed, but the fundamentals look like they are in place!

traits/src/signature/arrayref.rs
wysiwys: The main reason for this decision originally was that there would be no difference between `arrayref::Verify`, `owned::Verify`, and `secrets::Verify`, so `owned` and `secrets` just re-exported `arrayref::Verify`, rather than defining a redundant interface. However, after taking out the `secrets` traits, I believe it would make things simpler to combine the traits into one. Thanks for catching that!

libcrux-ml-dsa/src/impl_signature_trait.rs
wysiwys: We can, and that makes the docs simpler, too. I removed this to reduce how complex the trait implementation is

libcrux-ml-dsa/src/impl_signature_trait.rs
wysiwys: Added an implementation of this because I think it's a good abstraction for the context, and a good way to not exclude any sign/verify functionality. It can always be taken out later, too

ecdsa/src/impl_signature_trait.rs
wysiwys: I removed the type alias for a generic `Signer` here, and implemented the trait on a standalone signer struct.

wysiwys:
I added an initial implementation of `keygen()` to the traits in this PR.

traits/src/signature/arrayref.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Sign a payload using a provided signature key. Required auxiliary information is provided using
+    /// the `aux` argument.
+    fn sign(
+        payload: &[u8],
+        signing_key: &[U8; SIGNING_KEY_LEN],
+        signature: &mut [u8; SIGNATURE_LEN],
+        aux: Self::SignAux<'_>,
+    ) -> Result<(), SignError>;
+    /// Auxiliary information needed for verification.
+    type VerifyAux<'a>;
```
keks: I think we can drop this one now, also in the other traits

libcrux-ml-dsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// A convenience macro for implementing the [`Context`] trait.
+    /// The `$context` must be provided as a `&'static [u8]`.
+    ///
+    /// Usage:
+    /// ```rust
+    /// use libcrux_ml_dsa::signers::{Context, impl_context};
+    /// impl_context!(AppContext, b"context");
+    /// ```
+    macro_rules! impl_context {
+        ($name:ident, $context:expr) => {
```
keks: Maybe this should have a way for the user to provide a comment to the type that we define? Or we just assign one based on the byte string.

keks: If we do it based on the context, we might want to make it a literal, though.

traits/src/signature/arrayref.rs
```diff
@@ -0,0 +1,123 @@
...
+            Self::LibraryError => "indicates a library error",
+        };
+
+        f.write_str(text)
+    }
+}
+
+#[cfg(feature = "error-in-core")]
+mod error_in_core {
+
```
keks:
```suggestion
```

ecdsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,110 @@
...
+                    }
+                    Ok(())
+                }
+                fn keygen(signing_key: &mut [u8; SIGNING_KEY_LEN], verification_key: &mut [u8; VERIFICATION_KEY_LEN], randomness: [u8; RAND_KEYGEN_LEN]) -> Result<(), arrayref::KeyGenError> {
+                    use libcrux_traits::kem::arrayref::*;
+
+                    libcrux_p256::P256::keygen(verification_key, signing_key, &randomness)
+                        .map_err(|e| match e {
+                            libcrux_traits::kem::arrayref::KeyGenError::InvalidRandomness => arrayref::KeyGenError::InvalidRandomness,
+
```
keks, line 96:
```suggestion
```

ecdsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,121 @@
...
+                ) -> Result<(), arrayref::VerifyError> {
+                    let result = libcrux_p256::$verify_fn(
+                        payload.len().try_into().map_err(|_| arrayref::VerifyError::InvalidPayloadLength)?,
+                        payload,
+                        verification_key,
+                        <&[u8; 32]>::try_from(&signature[0..32]).unwrap(),
+                        <&[u8; 32]>::try_from(&signature[32..]).unwrap(),
+                    );
+                    if !result {
+                        return Err(arrayref::VerifyError::LibraryError);
```
keks, line 77: I don't think a failed signature verification should return a LibraryError

ecdsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,121 @@
...
+                ) -> Result<(), arrayref::SignError> {
+                    let result = libcrux_p256::$sign_fn(
+                        signature,
+                        payload.len().try_into().map_err(|_| arrayref::SignError::InvalidPayloadLength)?,
+                        payload,
+                        signing_key,
+                        &nonce.0,
+                    );
+                    if !result {
+                        return Err(arrayref::SignError::LibraryError);
```
keks, line 57: I don't know if this necessarily is a library error, because it can also step from invalid arguments: https://github.com/cryspen/libcrux/blob/3a5124fd4a8e949bd0c27c76a56cb64705b0eb3a/p256/src/p256.rs#L1998-L2000

ecdsa/src/impl_signature_trait.rs
wysiwys: Would it make sense to instead replace these with `UnknownError`s?

libcrux-ml-dsa/src/ml_dsa_generic/multiplexing.rs
keks, line 1: Somehow `sign_mut` existed in the instantiations but not in multiplexed. Added so we can impl arrayref

libcrux-ml-dsa/src/impl_signature_trait.rs
keks, line 1: I did a more complicate thing here for bounding the context sizes, because I didn't see a good way to return an error regarding this. This way we use the type system to ensure the error can't happen.

Dynamic contexts already were not supported in this API because we need to be able to determine the context from the type, so I hope that is fine.

keks (commented):
Left some comments on my changes

crates/primitives/signature/src/lib.rs
```diff
@@ -0,0 +1,98 @@
+#[cfg(any(feature = "ecdsa", feature = "ed25519", feature = "mldsa"))]
```
franziskuskiefer, line 1: A lib comment up here would be nice with a little info and an example on how to use this.

libcrux-ml-dsa/Cargo.toml
```diff
@@ -46,6 +48,11 @@ simd256 = ["libcrux-sha3/simd256", "libcrux-intrinsics/simd256"]
 acvp = []                                                        # expose internal API for ACVP testing
 test-utils = []                                                  # exposing internal functions for testing
 
+
+# thid doesn't check the secret independence of the implementation, but sets the
```
franziskuskiefer:
```suggestion
# This doesn't check the secret independence of the implementation, but sets the
```

libcrux-ml-dsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,218 @@
...
+                    ) -> Result<(), arrayref::SignError> {
+                        crate::ml_dsa_generic::multiplexing::$module::sign_mut(
+                            signing_key.declassify_ref(),
+                            payload,
+                            T::context(),
+                            randomness.declassify(),
+                            signature
+                        )
+                        .map_err(|err| match err {
+                            crate::types::SigningError::RejectionSamplingError=> arrayref::SignError::InvalidRandomness,
```
franziskuskiefer: If we add some includes a lot in here will get a little nicer to read.
And if we put the match into a `From` implementation, we could almost (or maybe really) skip the macro.

ed25519/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,82 @@
...
+    const SIGNATURE_LEN: usize = 64;
+    const RAND_KEYGEN_LEN: usize = SIGNING_KEY_LEN;
+
+    impl Sign<SIGNING_KEY_LEN, VERIFICATION_KEY_LEN, SIGNATURE_LEN, RAND_KEYGEN_LEN> for Signer {
+        /// Sign a payload with a provided signing key.
+        fn sign(
+            payload: &[u8],
+            signing_key: &[U8; SIGNING_KEY_LEN],
+            signature: &mut [u8; SIGNATURE_LEN],
+            _aux: (),
```
franziskuskiefer, line 21: Having a useless parameter in the API isn't great. Maybe the `aux` version should be a different trait? Or maybe traits just make it more complicated than necessary for signatures.

libcrux-ml-dsa/src/impl_signature_trait.rs
```diff
@@ -0,0 +1,218 @@
...
+            pub struct $name;
+            impl Context for $name {
+                type Buf = [u8; $len];
+                fn context() -> &'static [u8; $len] {
+                    $context
+                }
+            }
+        };
+    }
+    pub use impl_context;
```
franziskuskiefer, line 105: Asking the user to implement these types is a little odd. Also, I'm not sure how well this works for more dynamic contexts?
We could make this slightly different to just have array references and the caller just wraps them. I made a quick example here https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=661ba82dce767334432a31da021307e1

franziskuskiefer:
Closing, there'll be a new one.

PR #1081, opened 2025-07-23
[PSQ] Session storage

### LINKED ISSUE #1077: [PSQ] Session storage

jschneider-bensch:
This PR provides session storage for PSQv2.

- The `Transport` struct has been renamed `Session` and now holds the session key (as before), a `pk_binder` which is created on initialization and binds the available longterm public keys associated with the session, as well as a `channel_counter` which gets incremented for every secure channel derived from the session key.
- `Session` no longer implements `Protocol` directly, but allows to create fresh `Channel`s with `make_channel()`, which derives two fresh keys bound to the session, one for sending and one for receiving, i.e. `skChanneli2r = KDF(skCS, "i2r channel key" | pk_binder | channel_counter)` and  `skChannelr2i = KDF(skCS, "r2i channel key" | pk_binder | channel_counter)` . `Channel` implements `Protocol` and can then be used to send and receive messages.
- `Session` can be serialized (consuming it) with `Session::serialize`. Even though `Session` implements `tls_codec::Serialize`, this should not be called directly, since it does not consume the session and would allow to later deserialize an old session state, which could lead to nonce re-use. 
- `Session` can be deserialized with `Session::deserialize`, which recomputes the `pk_binder` given references to public keys provided by the caller. If the recomputed binder does not match the deserialized one, an error is raised.

Fixes #1077

### DIFF
--- a/libcrux-psq/src/protocol/api.rs
+++ b/libcrux-psq/src/protocol/api.rs
@@ -35,6 +35,7 @@ pub enum Error {
     TransportState,
     OutputBufferShort,
     PayloadTooLong,
+    ChannelError,
     Storage,
     OtherError,
     IdentifierMismatch,
@@ -184,7 +185,10 @@ impl Session {
 
     pub fn channel(&mut self) -> Result<Channel, Error> {
         let channel = Channel::new(&self, matches!(self.principal, SessionPrincipal::Initiator))?;
-        self.channel_counter += 1;
+        self.channel_counter = self
+            .channel_counter
+            .checked_add(1)
+            .ok_or(Error::ChannelError)?;
         Ok(channel)
     }
 


libcrux-psq/src/protocol/session.rs
```diff
@@ -7,6 +9,7 @@ pub const SESSION_ID_LENGTH: usize = 32;
 pub const SESSION_KEY_LENGTH: usize = 32;
 
 // XXX: Session storage to be implemented (cf. https://github.com/cryspen/libcrux/issues/1077)
+#[derive(TlsSerialize, TlsDeserialize, TlsSize)]
```
franziskuskiefer, line 11: There are `Secret*` versions of some of the structs. Maybe that's something we should use for the more permanent state? No strong opinion though.

libcrux-psq/src/protocol/api.rs
```diff
@@ -29,25 +35,123 @@ pub enum Error {
...
         Ok(Self {
-            session_key: derive_session_key(k2, tx2)?,
+            session_key,
+            pk_binder,
+            channel_counter: 0,
         })
     }
 
+    // XXX: tls_serialize should not be called directly, since it does
+    // not consume `Session`. This opens the possiblity for nonce
```
franziskuskiefer:
```suggestion
    // not consume `Session`. This opens the possibility for nonce
```

libcrux-psq/src/protocol/api.rs
```diff
@@ -29,25 +35,123 @@ pub enum Error {
...
+        )?;
         Ok(Self {
-            session_key: derive_session_key(k2, tx2)?,
+            session_key,
+            pk_binder,
+            channel_counter: 0,
         })
     }
 
+    // XXX: tls_serialize should not be called directly, since it does
```
franziskuskiefer: That's a good point. But is there something we should change? It looks like you're doing the right thing already. I'd drop the `XXX` then.

libcrux-psq/src/protocol/api.rs
```diff
@@ -65,17 +169,26 @@ struct TransportMessage {
     tag: [u8; 16],
 }
 
-impl Protocol for Transport {
+pub struct Channel {
```
franziskuskiefer, line 222: Doc comments?

libcrux-psq/src/protocol/api.rs
```diff
@@ -104,7 +216,7 @@ impl Protocol for Transport {
 }
 
 pub trait IntoTransport {
-    fn into_transport_mode(self) -> Result<Transport, Error>;
+    fn into_transport_mode(self) -> Result<Session, Error>;
```
franziskuskiefer: The function name here doesn't seem appropriate anymore? Maybe something like `into_session` instead? Same goes for the trait name itself.
The docs on this should then also make clear how it's used and why there's the session in between.

libcrux-psq/src/protocol/api.rs
```diff
@@ -29,25 +35,123 @@ pub enum Error {
...
+            responder_pq_pk,
+        )? == session.pk_binder
+        {
+            Ok(session)
+        } else {
+            Err(Error::Storage)
+        }
+    }
+
+    pub fn make_channel(&mut self) -> Result<(u32, Channel), Error> {
```
franziskuskiefer: Maybe just `channel`?

libcrux-psq/src/protocol/responder.rs
```diff
@@ -226,7 +226,14 @@ impl<'a, Rng: CryptoRng> Responder<'a, Rng> {
         let out_len = out_msg
             .tls_serialize(&mut &mut out[..])
             .map_err(Error::Serialize)?;
-        self.state = ResponderState::ToTransport(ToTransportState { tx2, k2 }.into());
+        self.state = ResponderState::ToTransport(
+            ToTransportState {
+                tx2,
+                k2,
+                initiator_ecdh_pk: Some(state.initiator_longterm_ecdh_pk.clone()),
```
franziskuskiefer: There should be no need to clone here. We're consuming the old state.
```suggestion
                initiator_ecdh_pk: Some(state.initiator_longterm_ecdh_pk),
```

libcrux-psq/src/protocol/api.rs
jschneider-bensch: I was noting this down, in case there is a better idea, or a way to enforce this perhaps.

Thinking a bit more about it, even this does not, I think, entirely prevent this sort of weakness, since the sessions and channels are symmetric at the initiator and the receiver. So one may store their session state, while the other keeps using it. When the storing party then deserializes and starts writing to any channel without first "catching up" with the other party's use in the meantime that's again a nonce re-use.

jschneider-bensch: I think the way around this is to do it like the spec says and have two keys per `Channel` instead of one.

franziskuskiefer: Yes, I agree, the two keys would be much safer. Let's do that.

libcrux-psq/src/protocol/session.rs
jschneider-bensch: Ah, that could be nice. I think the `AEADKey` would be good to zeroize on drop, but I'm not sure to achieve that. 

The type is the following:
```rust
pub struct AEADKey([u8; KEY_LEN], #[tls_codec(skip)] [u8; NONCE_LEN]);
```
(We `skip` the nonce, since including it leads to problems when we use the serialization of the key in another key derivation. The problem is that one party will have incremented the nonce after serialization while the other increments it after deserialization, when decrypting).

We derive `TlsSerialize` on that, and it must use the provided implementation for `[u8; l]` then. To do the `Secret*` version, we'd have to manually implement `Serialize`, right? Then again, that will zeroize on drop the serialization wrapper and not the `[u8; KEY_LEN]`, no?

libcrux-psq/src/protocol/api.rs
jschneider-bensch: I implemented it with two keys now. Every channel message will also include its channel identifier, so if the initiator and receiver go out of sync, that should become apparent on the first `Channel::read`. There is some room for improvement on the ergonomics in this case, since the peer that is "behind" on their channel counter now has to create channels until the channel numbers match again, but that can be addressed separately I think. I believe the way it's done now at least prevents accidental nonce re-use.

libcrux-psq/src/protocol/api.rs
jschneider-bensch: I would like to rename/move and document all of this more. But maybe that can be done in a follow-up, which also moves the legacy PSQ into a separate feature-guarded module.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -134,7 +133,7 @@ fn session_key_id(key: &AEADKey) -> Result<[u8; SESSION_ID_LENGTH], Error> {
     let prk = libcrux_hkdf::extract(
         Algorithm::Sha256,
         [],
```
franziskuskiefer: Should we use some static salt here to ensure domain separation. I don't think we hkdf extract on the key anywhere else. But better safe than sorry.

libcrux-psq/src/protocol/keys.rs
```diff
@@ -146,26 +145,63 @@ fn session_key_id(key: &AEADKey) -> Result<[u8; SESSION_ID_LENGTH], Error> {
...
+
+    AEADKey::new(
+        &session.session_key.key,
+        &ChannelKeyInfo {
+            domain_separator: if IS_INITIATOR {
+                I2R_CHANNEL_KEY_LABEL
+            } else {
+                R2I_CHANNEL_KEY_LABEL
+            },
+            pk_binder: session.pk_binder.clone(),
```
franziskuskiefer: We could avoid the clone by using a slice for `pk_binder`. But not too important, the compiler will handle this probably anyway.

libcrux-psq/tests/registration.rs
```diff
@@ -89,17 +89,32 @@ fn registration(pq: bool) {
     assert!(initiator.is_handshake_finished());
     assert!(responder.is_handshake_finished());
 
-    let mut i_transport = initiator.into_transport_mode().unwrap();
-    let mut r_transport = responder.into_transport_mode().unwrap();
+    let i_transport = initiator.into_session().unwrap();
+    let mut r_transport = responder.into_session().unwrap();
+
+    // test serialization, deserialization
+    i_transport.serialize(&mut msg_channel).unwrap();
```
franziskuskiefer: Maybe let's use a different tmp storage for this? Using `msg_channel` may be a bit confusing.

libcrux-psq/src/protocol/api.rs
```diff
@@ -29,54 +35,224 @@ pub enum Error {
...
         // We match the maximum payload length of Noise.
         if payload.len() > 65535 {
             return Err(Error::PayloadTooLong);
         }
         let mut ciphertext = vec![0u8; payload.len()];
-        let tag = self
-            .session_key
-            .key
-            .encrypt(payload, &[], &mut ciphertext)?;
+        let tag = self.send_key.encrypt(payload, &[], &mut ciphertext)?;
```
franziskuskiefer, line 257: Shall we allow providing the `aad`?

libcrux-psq/src/protocol/api.rs
```diff
@@ -29,54 +35,224 @@ pub enum Error {
...
+        {
+            Ok(session)
+        } else {
+            Err(Error::Storage)
+        }
+    }
+
+    pub fn channel(&mut self) -> Result<Channel, Error> {
+        let channel = Channel::new(&self, matches!(self.principal, SessionPrincipal::Initiator))?;
+        self.channel_counter += 1;
```
franziskuskiefer: This should check that it doesn't overflow before incrementing.

libcrux-psq/src/protocol/session.rs
franziskuskiefer: Yeah, the entire thing would need to drop things. I don't have a strong opinion here. Happy to skip for this PR.

franziskuskiefer (approved):
Thanks, lgtm with a few nits.

libcrux-psq/src/protocol/keys.rs
jschneider-bensch: Sure, it won't hurt :+1:

libcrux-psq/src/protocol/keys.rs
jschneider-bensch: Right, that's an easy fix

libcrux-psq/src/protocol/api.rs
jschneider-bensch: We don't currently have that in the spec. It would be static `aad` for the channel (different for initiator and responder?) to stay within the `read`/`write` API. If you think that would be useful I'll change it and add a comment to the spec.

libcrux-psq/src/protocol/session.rs
jschneider-bensch: Then I'm deferring this for later, along with using `conditionally_deserializable` for the `pk_binder` check in deserialization.

libcrux-psq/src/protocol/api.rs
franziskuskiefer: Let's add a comment to spec but not do anything here. We can add it later if we want to.

PR #1082, opened 2025-07-24
fix(mlkem/hax): edit hardcoded hash instance names

W95Psp:
https://github.com/cryspen/hax/pull/1570 kills metasized and sized traits, thus numbering changes.

For example:
```
fn f<T: Foo>() {}
```
`T` has an implicit bound `T: Sized`.

With https://github.com/cryspen/hax/pull/1570, this bound is dropped.
For local bounds, hax names them with numbers. Sized being the first bound and `Foo` the second, `T: Foo` used to be call `i1` and `T: Sized` was called `i0`.
Now `T: Sized` is dropped and `T: Foo` is called `i0`.

Since we hardcoded this name in Rust directly in ML-KEM, this breaks.

### DIFF
--- a/libcrux-ml-kem/src/ind_cca.rs
+++ b/libcrux-ml-kem/src/ind_cca.rs
@@ -416,7 +416,7 @@ pub(crate) fn decapsulate<
     hax_lib::fstar!(
         "assert_norm (pow2 32 == 0x100000000);
         assert (v (sz 32) < pow2 32);
-        assert (i4.f_PRF_pre (sz 32) $to_hash);
+        assert (i1.f_PRF_pre (sz 32) $to_hash);
         lemma_slice_append $to_hash $implicit_rejection_value ${ciphertext}.f_value"
     );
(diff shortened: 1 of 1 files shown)


W95Psp:
@jschneider-bensch, I pushed the fix I told you about, **should** be good now :)

W95Psp:
Note that CI will fail with hax main.
We will merge this PR when https://github.com/cryspen/hax/pull/1570 is in.

jschneider-bensch (approved):
Since these are in `fstar!` blocks, it wouldn't be possible to derive the names automatically, right? (Regardless whether hax currently supports that from Rust style annotations.)

W95Psp:
We actually could, just as you can write `fstar!("$some_rust_name")`, we would like to write something like `fstar!("${impl for T: Foo}")`: this is issue https://github.com/cryspen/hax/issues/1313.

W95Psp:
I'm re-running jobs, now that the PR is merged on hax side.

PR #1088, opened 2025-08-04
Add no_std check and replace wasm check

wysiwys:


.github/workflows/wasm-and-no-std-checks.yml
```diff
@@ -0,0 +1,36 @@
+name: Libcrux - wasm and no_std check
```
franziskuskiefer: No need for the libcrux prefix. It's the librux repo.

no-std-check.sh
franziskuskiefer, line 1: Maybe update the name of this file as well when it's doing more now?

wysiwys: Would something like `no-std-build.sh` work well for this use case? This script still builds the libcrux crates without std features enabled, but a name like this could reflect that the script is now used for multiple checks, not just for a single one

keks: I think no-std-build makes sense, given that this is the command it runs. But then again that is what it was doing before so I am not sure we need to rename it.

keks (approved):
Looks good!

no-std-check.sh
wysiwys: I changed the name to `no-std-build.sh` to make the name a bit more specific to how it is used.

PR #1091, opened 2025-08-07
[PSQ] Allow derivation of an Unregistered PSK.

georgio:
Hello :-)

I added a simple function which enables a PSQ initiator to extract the PSK before the PSQ Responder sends their message.
This is very helpful for nym.

Thanks for taking a look, I'm happy to answer any questions.

### DIFF
--- a/libcrux-psq/src/psk_registration.rs
+++ b/libcrux-psq/src/psk_registration.rs
@@ -156,8 +156,8 @@ impl Initiator {
 
     /// Returns the unregisterd PSK.
     /// This is computed while generating the inital message.
-    pub fn unregistered_psk(&self) -> Psk {
-        self.unregistered_psk
+    pub fn unregistered_psk(&self) -> &Psk {
+        &self.unregistered_psk
     }
 }
 


libcrux-psq/src/psk_registration.rs
```diff
@@ -147,6 +147,11 @@ impl Initiator {
 
         Ok(RegisteredPsk { psk, psk_handle })
     }
+
+    /// Derives a PSK from the initiator's PQ-PrePSK, if initialized.
+    pub fn derive_unregistered_psk(&self) -> Result<Psk, Error> {
+        derive_psk(&self.k_pq)
```
franziskuskiefer: This way we have to compute the PSK at least twice.
How about we add the `psk` to the state and derive it in `send_initial_message` right away? Then this becomes a getter (e.g. `unregistered_psk`), and we can just return it in `complete_handshake`?

Also, it would be great to add it to a test to make sure it works and that the derived PSK is the same here and when completing the handshake.

franziskuskiefer (commented):
Thanks, generally lgtm but I think we could improve it a little bit.

libcrux-psq/src/psk_registration.rs
georgio: Yes that makes more sense. I will do that.

libcrux-psq/src/psk_registration.rs
```diff
@@ -143,9 +148,16 @@ impl Initiator {
...
+            psk: self.unregistered_psk,
+            psk_handle,
+        })
+    }
 
-        Ok(RegisteredPsk { psk, psk_handle })
+    /// Returns the unregisterd PSK.
+    /// This is computed while generating the inital message.
+    pub fn unregistered_psk(&self) -> Psk {
+        self.unregistered_psk
```
franziskuskiefer: Let's make this a reference in case the caller doesn't need to own it.

```suggestion
    pub fn unregistered_psk(&self) -> &Psk {
        &self.unregistered_psk
```

PR #1092, opened 2025-08-08
Sha3 lax check

parrot7483:
This PR adds lax checking for SHA3. It also includes some changes towards full type checking. The theta, pi and chi functions are admitted.  

It is based on my [develop branch](https://github.com/Parrot7483/libcrux/tree/develop) which changes are requested in [PR 1076](https://github.com/cryspen/libcrux/pull/1076)


fstar-helpers/core-models/src/abstractions/bitvec.rs
```diff
@@ -343,6 +343,7 @@ pub mod int_vec_interp {
                         #[hax_lib::opaque]
                         #[hax_lib::lemma]
                         #[hax_lib::fstar::smt_pat(BitVec::[< to_ $name >](BitVec::[<from_ $name>](x)))]
+                        #[allow(dead_code)]
```
franziskuskiefer: Why is this needed? We shouldn't have dead code in here. Also, given that this is a public function in a library.

libcrux-intrinsics/src/avx2.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unused_unsafe)]
```
franziskuskiefer: We shouldn't do this blanket for the entire crate.
Some intrinsics were made safe recently. Let's instead identify them, remove the unsafe, and document.

libcrux-sha3/src/traits.rs
```diff
@@ -1,10 +1,14 @@
+use hax_lib;
+
 // XXX: These should be default functions on `KeccakItem`, but hax doesn't
 //      support that yet. cryspen/hax#888
+#[cfg_attr(hax, hax_lib::requires(i < 5 && j < 5))]
```
franziskuskiefer: The `cfg_attr` shouldn't be needed anywhere.

libcrux-sha3/src/traits.rs
```diff
@@ -71,7 +96,7 @@ pub(crate) trait Squeeze1<T: KeccakItem<1>> {
 /// Store blocks `N = 2`
 #[cfg(feature = "simd128")]
 pub(crate) trait Squeeze2<T: KeccakItem<2>> {
-    fn squeeze<const RATE: usize>(
+    fn squeeze2<const RATE: usize>(
```
franziskuskiefer, line 103: It's not great that we need to rename these, but it's fine. But we should mention the hax bug that forces us to do this in both cases.

libcrux-sha3/Cargo.toml
```diff
@@ -16,11 +16,11 @@ bench = false # so libtest doesn't eat the arguments to criterion
 [dependencies]
 libcrux-platform = { version = "0.0.2", path = "../sys/platform" }
 libcrux-intrinsics = { version = "0.0.3", path = "../libcrux-intrinsics" }
+hax-lib.workspace = true
 
 # This is only required for verification.
 # The hax config is set by the hax toolchain.
 [target.'cfg(hax)'.dependencies]
```
franziskuskiefer: Looks like this isn't needed anymore.

libcrux-sha3/Cargo.toml
```diff
@@ -32,6 +32,7 @@ harness = false
 
 [dev-dependencies]
 criterion = "0.7.0"
+hax-lib.workspace = true
```
franziskuskiefer: No need here for this here. It's already a library dependency.

libcrux-sha3/hax.py
```diff
@@ -114,7 +116,7 @@ def __call__(self, parser, args, values, option_string=None) -> None:
             "--interfaces",
             interface_include,
         ]
-        hax_env = {}
+        hax_env = {"RUSTFLAGS": "-C debug-assertions=off"}
```
franziskuskiefer: Why is this needed?

libcrux-sha3/src/lib.rs
```diff
@@ -40,14 +40,13 @@ pub enum Algorithm {
     Sha512 = 4,
 }
 
-impl From<u32> for Algorithm {
```
franziskuskiefer, line 49: Why did you remove the `From` implementation? We'll want to keep this.

libcrux-sha3/src/lib.rs
```diff
@@ -40,14 +40,13 @@ pub enum Algorithm {
...
-    fn from(v: u32) -> Algorithm {
-        match v {
-            1 => Algorithm::Sha224,
-            2 => Algorithm::Sha256,
-            3 => Algorithm::Sha384,
-            4 => Algorithm::Sha512,
-            _ => panic!(),
+impl Clone for Algorithm {
```
franziskuskiefer: Deriving clone should work just fine. What issues did you run into with this?

libcrux-sha3/src/generic_keccak.rs
```diff
@@ -26,12 +26,25 @@ pub(crate) mod simd256;
...
-#[cfg_attr(hax, hax_lib::opaque)]
-#[derive(Clone, Copy)]
+// TODO: The Copy and Clone traits should be able to be derived,
+// but the hax extraction does not type check correctly.
+#[derive(Copy)]
 pub(crate) struct KeccakState<const N: usize, T: KeccakItem<N>> {
     pub(crate) st: [T; 25],
 }
 
+impl<const N: usize, T: KeccakItem<N>> Clone for KeccakState<N, T>
```
franziskuskiefer: Same as the other one. Deriving `Clone` should work fine.

libcrux-sha3/src/generic_keccak.rs
```diff
@@ -86,38 +100,38 @@ impl<const N: usize, T: KeccakItem<N>> KeccakState<N, T> {
...
             T::rotate_left1_and_xor(c[(2 + 4) % 5], c[(2 + 1) % 5]),
             T::rotate_left1_and_xor(c[(3 + 4) % 5], c[(3 + 1) % 5]),
             T::rotate_left1_and_xor(c[(4 + 4) % 5], c[(4 + 1) % 5]),
-        ];
+        ]
+    }
 
+    #[inline(always)]
+    #[hax_lib::fstar::replace_body("assert true")]
+    fn rho(&mut self, t: [T; 5]) {
```
franziskuskiefer, line 101: Looks like calling `theta` in here would make more sense than calling it on the outside. Did you check the benchmarks if any of this has an impact on the performance?

libcrux-sha3/src/simd/portable.rs
```diff
@@ -5,7 +5,7 @@ use crate::{generic_keccak::KeccakState, traits::*};
 #[inline(always)]
 fn rotate_left<const LEFT: i32, const RIGHT: i32>(x: u64) -> u64 {
     debug_assert!(LEFT + RIGHT == 64);
-    x.rotate_left(LEFT as u32)
+    u64::rotate_left(x, LEFT as u32)
```
franziskuskiefer: Why is this change needed?

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -131,24 +131,28 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
+            return 0;
+        }
+
+        // Remaining space to complete a full block
+        let need = RATE - self.buf_len;
+        if input_len < need {
+            return 0;
+        }
+
+        let end = RATE; // buf_len + need == RATE
```
franziskuskiefer: Looks like we could just use `RATE` instead of `end`?

parrot7483: I am going to revert changes in this function. This was towards full type check. I will put that in a separate PR

libcrux-sha3/Cargo.toml
parrot7483: correct. Its needed in `[target.'cfg(hax)'.dependencies]` and `[dev-dependencies]`.

libcrux-sha3/hax.py
parrot7483: I added this because every debug assertion produced a type error. Its fine lax checking

libcrux-sha3/Cargo.toml
parrot7483: Without this rust-analyzer won't work.

libcrux-sha3/src/lib.rs
parrot7483: This was a mistake. The panic in here resolves to the wrong F* function ([Issue 1571](https://github.com/cryspen/hax/issues/1571))

libcrux-sha3/src/lib.rs
parrot7483: Correct. I forgot to remove this change.

libcrux-sha3/src/generic_keccak.rs
parrot7483: correct

libcrux-sha3/src/simd/portable.rs
parrot7483: No. Not needed.

libcrux-sha3/src/traits.rs
parrot7483: correct

libcrux-sha3/src/generic_keccak.rs
parrot7483: I just compared the current version on main and this. Some benchmarks are faster and some slower. All within 1.2%.

libcrux-sha3/Cargo.toml
parrot7483: I miss-understand how `hax_lib` is ignored for release target. Removing it from `dependencies`, nothing gets build and cargo complains about missing `hax_lib` dependency.

fstar-helpers/core-models/src/abstractions/bitvec.rs
parrot7483: These function throw a warning: `lemma_cancel_iv`, `lemma_cancel_iv` because of the `interpretations!` macro invocation

franziskuskiefer: If something here's unused, let's comment it out and open an issue instead to remove it.

libcrux-sha3/Cargo.toml
franziskuskiefer: You should add `hax-lib` as a regular dependency. No `cfg` or `dev-dependencies` needed. Everything else happens inside the `hax-lib` crate.

libcrux-sha3/hax.py
franziskuskiefer: That doesn't sound right. I'll let @karthikbhargavan comment on that.

libcrux-intrinsics/src/avx2.rs
parrot7483: addressed in https://github.com/cryspen/libcrux/issues/1100

libcrux-sha3/src/traits.rs
parrot7483: I added a comment. I thing this might be solved with the full type checking I am currently working on. I moved the pre and post conditions around

parrot7483:
One problem is this [issue](https://github.com/cryspen/hax/issues/1571). I always get the wrong `explicit_panic` in F*.

parrot7483:
> One problem is this [issue](https://github.com/cryspen/hax/issues/1571). I always get the wrong `explicit_panic` in F*.

Fixed in HAX https://github.com/cryspen/hax/commit/ad0e608c05c111ce9fe17c6ba38a16fcd65313ad

libcrux-sha3/src/generic_keccak.rs
franziskuskiefer: That sounds ok for now. But let's keep an eye on it.

libcrux-sha3/src/traits.rs
franziskuskiefer: Can you link to the corresponding hax issue?

parrot7483:
> It looks like you force pushed. That breaks the entire review process. Please don't force push during the review.

Yes, sorry. I rebased because I wanted the Makefile changes to be merged separately.

parrot7483:
The new CI does not work with the Makefile. Will investigate. My first time working with Github workflows.

libcrux-sha3/hax.py
parrot7483: `RUSTFLAGS` where removed. Not a problem for LAX checking. Will have to revisit for full type check.

parrot7483:
CI still fails because hax-lib is pulled from 

```
--include /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hax-lib-0.3.2/proof-libs/fstar/core 
--include /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hax-lib-0.3.2/proof-libs/fstar/rust_primitives 
--include /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hax-lib-0.3.2/proofs/fstar/extraction
```

W95Psp:
As we discussed on Google Chat, the problem here probably has nothing to do with the hax installation action.
The Makefile of F* will ask cargo for the path in which it has to find hax-lib.
So here, I believe you should just pin `hax-lib` in `Cargo.toml` to the git commit you want to use.

Does that make sense?

parrot7483:
The `Makefile.base` automatically finds hax-lib/proofs-lib/fstar/code twice. That's why I use the HAX_HOME variable in the SHA3 [Makefile](https://github.com/cryspen/libcrux/blob/139eac03d88f28a10d644cbf7ce0b5698005b223/libcrux-sha3/proofs/fstar/extraction/Makefile)

```
--include /home/runner/.cargo/git/checkouts/hax-580ebeee043cdea1/6b34a68/hax-lib/proof-libs/fstar/core
--include /home/runner/.cargo/git/checkouts/hax-580ebeee043cdea1/6b34a68/hax-lib/proof-libs/fstar/rust_primitives
--include /home/runner/.cargo/git/checkouts/hax-580ebeee043cdea1/6b34a68/hax-lib/proofs/fstar/extraction
--include /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hax-lib-0.3.2/proof-libs/fstar/core
```

```
cargo tree --invert registry+https://github.com/rust-lang/crates.io-index#hax-lib@0.3.2
+ command cargo tree --invert registry+https://github.com/rust-lang/crates.io-index#hax-lib@0.3.2
hax-lib v0.3.2
├── libcrux-intrinsics v0.0.3 (/home/tristan/Repositories/libcrux/libcrux-intrinsics)
│   └── libcrux-sha3 v0.0.3 (/home/tristan/Repositories/libcrux/libcrux-sha3)
└── libcrux-secrets v0.0.3 (/home/tristan/Repositories/libcrux/secrets)
    └── libcrux-traits v0.0.3 (/home/tristan/Repositories/libcrux/traits)
        └── libcrux-sha3 v0.0.3 (/home/tristan/Repositories/libcrux/libcrux-sha3)
```
        
```
cargo tree --invert git+https://github.com/cryspen/hax.git?rev=6b34a68a0b38cd2aff09a4ad05569d712620af27#hax-lib@0.3.2
+ command cargo tree --invert 'git+https://github.com/cryspen/hax.git?rev=6b34a68a0b38cd2aff09a4ad05569d712620af27#hax-lib@0.3.2'
hax-lib v0.3.2 (https://github.com/cryspen/hax.git?rev=6b34a68a0b38cd2aff09a4ad05569d712620af27#6b34a68a)
└── libcrux-sha3 v0.0.3 (/home/tristan/Repositories/libcrux/libcrux-sha3)
```

W95Psp:
Oh, I see, I recall your PR introducing `HAX_HOME` in the README.
I think this is not so great: now, you can have a Rust hax-lib inconsistant with the F* hax-lib.

Looking at the commit change, I don't see where `HAX_HOME` is supposed to be defined, in CI.
The libcrux jobs CI are not using the flake.nix from this repo.

parrot7483:
I changed the version of hax-lib in root [Cargo.toml](https://github.com/cryspen/libcrux/pull/1092/commits/d1b7358ab1905f131cad39ac8329213f696ee0a1#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) to use the latest commit. Only using latest version of hax-lib in libcrux-sha3/Cargo.toml causes the `Makefile.generic` to find different versions of hax-lib and then verification fails. 

What do you think @franziskuskiefer

franziskuskiefer:
> I changed the version of hax-lib in root [Cargo.toml](https://github.com/cryspen/libcrux/pull/1092/commits/d1b7358ab1905f131cad39ac8329213f696ee0a1#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) to use the latest commit. Only using latest version of hax-lib in libcrux-sha3/Cargo.toml causes the `Makefile.generic` to find different versions of hax-lib and then verification fails.
> 
> What do you think @franziskuskiefer

That's ok to get this in. But please file a follow-up issue to get this fixed.

Also, we need to get the hax action fixed to avoid this hack. I filed https://github.com/hacspec/hax-actions/issues/5 for that.

W95Psp:
Thanks for filing the issue, Franziskus, I replied there. I believe this isn’t something the action can address, as it works as intended in this case.
What F* library picked is decided by the Makefile that lives in this repo, the action has no power over it.

IMO, the way to improve things here is to allocate time on the workbench to kill makefiles altogether.

franziskuskiefer:
Ok, then we'll have to get this fixed in here properly before we can merge it.

W95Psp:
You can now update `hax-lib` to `0.3.4`, hax was released.

parrot7483:
I created a separate bump https://github.com/cryspen/libcrux/pull/1127.

I will clean up the last couple commits and rebase.

libcrux-sha3/Cargo.toml
```diff
@@ -16,16 +16,14 @@ bench = false # so libtest doesn't eat the arguments to criterion
...
-# The hax config is set by the hax toolchain.
-[target.'cfg(hax)'.dependencies]
+libcrux-traits = { version = "=0.0.3", path = "../traits/", optional = true }
 hax-lib.workspace = true
 
 [features]
+default = ["digest-traits"]
 simd128 = ["libcrux-intrinsics/simd128"]
 simd256 = ["libcrux-intrinsics/simd256"]
+digest-traits = ["libcrux-traits"]
```
franziskuskiefer: This shouldn't be a feature. If you want to ignore this for hax, do that instead `#[cfg(not(hax))]`.

libcrux-sha3/src/traits.rs
```diff
@@ -59,10 +71,27 @@ pub(crate) trait Absorb<const N: usize> {
...
+}
+"
+)]
 pub(crate) trait Squeeze1<T: KeccakItem<1>> {
     fn squeeze<const RATE: usize>(&self, out: &mut [u8], start: usize, len: usize);
 }
 
+// Renaming the squeeze functions of the Squeeze2 and Squeeze4 Trait is currently
+// necessary because F* will not find the correct function, pre and post conditions otherwise.
+// Check the follwoing issue: https://github.com/cryspen/hax/issues/1595
```
franziskuskiefer:
```suggestion
// Check the following issue: https://github.com/cryspen/hax/issues/1595
```

.github/workflows/sha3-hax.yml
```diff
@@ -63,3 +110,27 @@ jobs:
...
+    if: ${{ always() }}
+    needs: [get-hax-ref, lax-portable]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Successful
+        if: ${{ !(contains(needs.*.result, 'failure')) }} 
+        run: exit 0 
+      - name: Failing
+        if: ${{ (contains(needs.*.result, 'failure')) }} 
+        run: exit 1
```
franziskuskiefer:
```suggestion
        run: exit 1

```

.github/workflows/sha3-hax.yml
franziskuskiefer, line 1: Let's only run this workflow if something in the sha3 crate changed. You can use [paths](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore) for that.

.github/workflows/sha3-hax.yml
```diff
@@ -63,3 +110,27 @@ jobs:
       - name: Failing
         if: ${{ (contains(needs.*.result, 'failure')) }} 
         run: exit 1 
+
+  sha3-extract-portable-hax-status:
```
franziskuskiefer: Do we always need a separate status here? I think all of this is always running and needs to succeed?

franziskuskiefer (changes requested):
Thanks, almost there. Just a few nits, then we can get this in.

.github/workflows/sha3-hax.yml
parrot7483: Why do we need this status steps? I added them because they were in the other workflows.

franziskuskiefer: They are necessary to handle jobs that run on the merge queue. Some jobs, for example benchmarks, are only run on the merge queue, not on PRs. But because they are required, they need a status. The GH merge queue is not great yet. Hence these workarounds.
I'm fine with keeping this as is for now. But we may need to update that later.

libcrux-sha3/src/lib.rs
```diff
@@ -10,9 +10,9 @@ mod simd;
 
 mod generic_keccak;
 
-#[cfg(not(eurydice))]
+#[cfg(not(hax))]
```
franziskuskiefer: You removed the `eurydice` guard on these. We need to keep them.

franziskuskiefer (approved):
Thanks! Let's get this in.

parrot7483:
I don't see an error in the log that failed. What is the problem here?

franziskuskiefer:
Looks like the branch is pretty out of date. Updating may fix it. If not, C extraction must be updated.

Issue #1054, opened 2025-07-09
secrets: Blanket impls should also work for public integers (and just provide identity functions)


keks:
this code https://github.com/cryspen/libcrux/blob/bdf3bbb83d2c97a8c6886bb4d3c992e5f42c28a4/secrets/src/int/classify.rs should also be in scope for public integers, because we need a uniform API for both.

franziskuskiefer:
@keks check if this is fixed in #1094

Issue #299, opened 2024-06-04
Improve secret independence guarantees in compiled code.


xvzcf:
Following up on this: https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/hqbtIGFKIpU, we should re-examine the way we write secret-independent code and we should also check the resulting binary code to see whether the compiler has optimized away from the constant time code or deviated in any other way from our intentions.

We could mark any function that is intended to produce a mask in constant time as "never inline".

We could also run checks on the resulting binary using https://github.com/rozbb/dudect-bencher. Perhaps we could mimic a TIMECOP like checking approach using [Miri](https://github.com/rust-lang/miri) and https://doc.rust-lang.org/nomicon/leaking.html.

jschneider-bensch:
Addressed by addition of valgrind testing via secrets.

We're working towards integrating these across the workspace.

PR #1094, opened 2025-08-12
[secrets] Add constant time swap and select

jschneider-bensch:
This PR integrates the `Swap` and `Select` traits from this branch into the `secrets` crate.
This provides implementations of constant-time swapping and selection of values for `u8, u16, u32, u64`, their secret type counterparts as well as slices of these types:
- for `aarch64` we use the `csel` instruction via inline assembly to guarantee constant time,
- for everything else we provide a best-effort portable implementation

I think this could be simplified a bit once https://github.com/cryspen/libcrux/issues/1054 is addressed, but I didn't want to do that in this PR.


secrets/src/int.rs
```diff
@@ -122,3 +122,604 @@ impl_cast_ops!(I64);
...
+
+#[cfg(not(any(target_arch = "aarch64")))]
+mod portable {
+    use super::*;
+    use super::{Select, Swap};
+
+    // Don't inline this to avoid that the compiler optimizes this out.
+    #[inline(never)]
+    fn is_non_zero(selector: u8) -> u64 {
+        core::hint::black_box(((!(selector as u64)).wrapping_add(1) >> 63) & 1)
```
keks: should this be `selector as u64` also on 32 bit architectures?

secrets/src/int.rs
```diff
@@ -122,3 +122,604 @@ impl_cast_ops!(I64);
...
+    }
+
+    #[cfg(feature = "check-secret-independence")]
+    impl Select for [U8] {
+        fn select(&mut self, other: &Self, selector: u8) {
+            let lhs = self.declassify_ref_mut();
+            let rhs = other.declassify_ref();
+            lhs.select(rhs, selector);
+        }
+    }
```
keks: could we put this into a macro so we don't have to duplicate it for all the ints?

secrets/src/int.rs
```diff
@@ -122,3 +122,604 @@ impl_cast_ops!(I64);
...
+
+    macro_rules! swap {
+        ($t:ty, $lhs:expr, $rhs:expr, $selector:expr) => {
+            let mask = core::hint::black_box(
+                ((((!($selector as u64)).wrapping_add(1) >> 63) & 1) as $t).wrapping_sub(1),
```
keks, line 200: same question re what about 32bit architectures here

secrets/src/int.rs
```diff
@@ -122,3 +122,604 @@ impl_cast_ops!(I64);
...
+    #[cfg(feature = "check-secret-independence")]
+    impl Swap for [U8] {
+        #[inline]
+        fn cswap(&mut self, other: &mut Self, selector: u8) {
+            debug_assert_eq!(self.len(), other.len());
+            let lhs = self.declassify_ref_mut();
+            let rhs = other.declassify_ref_mut();
+            swap!(u8, lhs, rhs, selector);
+        }
+    }
```
keks: same comment regarding pulling all of this into a macro

secrets/src/int.rs
```diff
@@ -122,3 +122,604 @@ impl_cast_ops!(I64);
...
+
+    impl<T: Select> Select for [T] {
+        #[inline]
+        fn select(&mut self, other: &Self, selector: u8) {
+            debug_assert_eq!(self.len(), other.len());
+            for i in 0..self.len() {
+                (&mut self[i]).select(&other[i], selector);
+            }
+        }
+    }
```
keks: should this live in the architecture-specific code or can we move it to next to the trait definition?

secrets/src/int.rs
```diff
@@ -122,3 +122,604 @@ impl_cast_ops!(I64);
...
+
+    impl<T: Swap> Swap for [T] {
+        #[inline]
+        fn cswap(&mut self, other: &mut Self, selector: u8) {
+            for i in 0..self.len() {
+                (&mut self[i]).cswap(&mut other[i], selector);
+            }
+        }
+    }
```
keks: same question here

secrets/src/int.rs
```diff
@@ -122,3 +122,604 @@ impl_cast_ops!(I64);
...
+
+    use super::*;
+    use rand::{rng, Fill, Rng, RngCore};
+
+    fn test<T: Classify + Default + Copy + PartialEq + Eq + Debug>(rng: &mut impl RngCore)
+    where
+        [T]: Fill + Select,
+    {
+        let selector: bool = rng.random();
+        let selector = if selector { 0 } else { 1 };
```
keks: this is a massive nitpick, but how about `let selector: u8 = rng.random()&1`?

secrets/src/int.rs
jschneider-bensch: Good question!
@franziskuskiefer: Could we use `usize` to make this independent of the architecture?
```rust
    fn is_non_zero(selector: u8) -> usize {
        core::hint::black_box(((!(selector as usize)).wrapping_add(1) >> (usize::BITS - 1)) & 1)
    }
```

secrets/src/int.rs
jschneider-bensch: Good call, will do!

secrets/src/int.rs
jschneider-bensch: In the portable version, we don't actually implement `Select` and `Swap` on the base types (maybe we should, though), but only on slices, since we can then compute the mask value once for the entire slice. Not sure if the compiler could re-use the mask if we re-computed it for every slice index instead, since we also wrap it in a `core::hint::black_box`.

secrets/src/int.rs
jschneider-bensch: Sure, that works!

secrets/src/int.rs
franziskuskiefer: It depends on the usage. I don't think we want the portable arithmetic to be platform dependent. This wouldn't be usable with `usize`. We could do a `u32` and a `u64` version though.

jschneider-bensch: How would you decide which version to use?

franziskuskiefer: Well, this is a purely internal function. So it depends on the caller here what it needs. I guess right now it needs u64. The current versions in mlkem and mldsa use u8 I think.

jschneider-bensch: Okay, so would it be safe to use `selector as u32` for `u8, u16` and `u32` and `selector as u64` for `u64`? We cast the result to the type that needs masking anyway, so if we are on a 32-bit machine that means for anything within word-size we don't do any unnecessary computation.

franziskuskiefer: Yeah I think that makes sense.

jschneider-bensch: I changed it, as described above.

secrets/src/int.rs
jschneider-bensch: See discussion above.

secrets/src/int.rs
jschneider-bensch: See comment above.

secrets/src/int.rs
```diff
@@ -125,88 +125,52 @@ impl_cast_ops!(I128);
...
-    impl Select for [u16] {
-        fn select(&mut self, other: &Self, selector: u8) {
-            let mask = (is_non_zero(selector) as u16).wrapping_sub(1);
-            for i in 0..self.len() {
-                self[i] = (self[i] & mask) | (other[i] & !mask);
-            }
-        }
+    fn is_non_zero_32(selector: u8) -> u32 {
+        core::hint::black_box(((!(selector as u32)).wrapping_add(1) >> 31) & 1)
     }
```
keks: I don't think this is enough. The compiler sees what arguments the function is called with and will specialize it if possible: https://godbolt.org/z/YjnWPEMjG

I think the issue is that it did the full computation with optimizations, but the calling code is not forbidden to optimize depending on the result. What we want is the opposite I think, so we put `selector` in a black box instead of the whole term. Unfortunately, in that case it optimizes to use a `cmp` instruction, which defeats the purpose: https://godbolt.org/z/Gb4hT47oP

Adding one more black box in the middle seems to do the trick, at the cost of a few pointless loads and stores: https://godbolt.org/z/K48EePT3c

This can even be inlined: https://godbolt.org/z/raYzbzodc

I don't like the additional instructions, but maybe that is fine for now? I think the only way around them would be to use assembly, which might be a good idea at some point.

secrets/src/traits.rs
```diff
@@ -138,3 +138,21 @@ impl Scalar for core::arch::aarch64::uint64x1_t {}
 
 #[cfg(target_arch = "aarch64")]
 impl Scalar for core::arch::aarch64::uint64x2_t {}
+
+/// Best effort constant time swapping of values.
+pub trait Swap {
+    /// Depending on `selector`, keep everything as is, or swap `self` and `other`.
+    ///
+    /// If `selector == 0`, the values are unchanged, otherwise swap.
+    fn cswap(&mut self, other: &mut Self, selector: u8);
```
keks: `selector` should be U8, no?

secrets/src/traits.rs
```diff
@@ -138,3 +138,21 @@ impl Scalar for core::arch::aarch64::uint64x1_t {}
...
+}
+
+/// Best effort constant time selection of values.
+pub trait Select {
+    /// Select `self` or `other`, depending on `selector`.
+    /// The selected value will be in `self`.
+    ///
+    /// If `selector != 0`, select `other`, otherwise
+    /// `self` is unchanged.
+    fn select(&mut self, other: &Self, selector: u8);
```
keks: same here?

secrets/src/int.rs
keks: Oh and finally one more question: I noticed the compiler always optimizes away the `&1`, and I wonder if it's really needed. We right-shift a 32 bit number by 31, isn't that guaranteed to be 0 or 1?

keks (dismissed):
Thanks! I missed some thing last round, but that should have been it. Let's get these fixed and then merge

secrets/src/int.rs
jschneider-bensch: I added the additional black box here, to make the compilation more robust. I think we can afford the extra instructions since these operations should probably not be run in a loop.
If it turns out to be a bottleneck somewhere, we may revisit, but to me it seems we'd best be as conservative as possible.

secrets/src/traits.rs
jschneider-bensch: You're right! I've addressed that now.

secrets/src/int.rs
```diff
@@ -123,31 +123,55 @@ impl_cast_ops!(I64);
...
     #[inline(never)]
     fn is_non_zero_32(selector: u8) -> u32 {
-        core::hint::black_box(((!(selector as u32)).wrapping_add(1) >> 31) & 1)
+        core::hint::black_box(
+            ((!(core::hint::black_box(selector) as u32)).wrapping_add(1) >> 31) & 1,
+        )
     }
 
     // Don't inline this to avoid that the compiler optimizes this out.
     #[inline(never)]
```
keks, line 161: I think we can drop the inline[never] now

keks (approved):
I think this looks good! Can you add an issue for figuring out how to check on CI that the generated code doesn't get optimized away? Not sure how exactly that would work, but maybe something like "grep for cmp and fail if its there" and/or "grep for inc/not/shr and require they are there" would be a good start.

I left one comment, not sure we want to do that now or after we have the check on CI. We could open another issue for that.

secrets/src/int.rs
jschneider-bensch: We might! But indeed, I'd feel more comfortable dropping this only after we have somewhat automated checks that this doesn't break the assembly.

jschneider-bensch:
Right, I think it could be a part of https://github.com/cryspen/libcrux/issues/299.
I think we'd need a check that can be feasibly automated like you describe, and we'd want to run that on the actual uses of the code that should be constant time. I'm not sure what we gain from checking it on, say example code, since perhaps it would be treated differently by the compiler in another context of a real application. I find it tricky to have confidence here; I guess one way would be to always build a `cdylib` and check that?

jschneider-bensch:
> Right, I think it could be a part of #299. 

Filed a sub-issue for this: https://github.com/cryspen/libcrux/issues/1111

keks:
Right. I think it will be difficult to perform such checks if the function is inlined because it will be hard to isolate the code, but you are right that this would be closer to what we actually want to check. I would expect that the compiler would optimize small examples more aggressively because it can infer more stuff, that's why I thought we could do it that way. Maybe a mix would be good:

- have a cfg_attr that enables inline(never). In that assembly of real code, isolate the function and make the checks.
- have some small examples compiled with inlining and check in the assembly that this is as expected.

That way we are *maybe* convinced that we are constant time with inline(always) for most code.

But before we go all that length we should probably measure what the perf impact of not inlining is :)

franziskuskiefer:
Looks like this breaks C extraction.

PR #1096, opened 2025-08-14
Add old C extraction configuration

keks:


protz:
Thanks -- @keks how does this differ from branch protz/ci? Is it a straight copy of what's over there?

@franziskuskiefer and I talked this morning and another solution (which you might prefer) would be to put this over in the Eurydice tree as part of Eurydice's CI. There are a couple advantages to this solution:
1. the code can be checked in without risk of confusion, meaning that I can track very closely the impact of my changes in Eurydice to libcrux (good for libcrux)
2. no need to maintain copies of the headers -- the eurydice build finds these automatically via fetching dependencies
3. no need for me to submit PRs (and for you to review them) -- should anything change in the way charon is invoked, or the way eurydice is used, you know that there is a reference version that works upstream in eurydice and you can use that whenever you want to do a toolchain update

Thoughts? Thanks for this PR!

franziskuskiefer:
I think esp. if we need to duplicate the build and test files anyway, I indeed don't see many benefits of maintaining it here in libcrux. Moving it closer to the tool also makes sense.

Unless @keks thinks differently, I'd be happy to help transferring these bits over to Eurydice.

protz:
Sounds great. Thanks

franziskuskiefer:
As discussed, closing this in favour of moving this to eurydice.

PR #1101, opened 2025-08-19
Makefile improvement

Parrot7483:
This PR changes two things in the default Makefile. 
- `FINDLIBS_OUTPUT ?=` makes it possible to disable automatic import of modules in a makefile using the default one.
- Prevent the `clean` to run the build target and remove more auxiliary files 

PR https://github.com/cryspen/libcrux/pull/1092 is based on these two commits.

### DIFF
--- a/fstar-helpers/Makefile.generic
+++ b/fstar-helpers/Makefile.generic
@@ -100,7 +100,7 @@ endef
 export FINDLIBS
 
 FSTAR_INCLUDE_DIRS_EXTRA ?=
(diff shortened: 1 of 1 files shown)


W95Psp:
My approval is worthless on libcrux I think, you should ask review from i.e. @jschneider-bensch

parrot7483:
> My approval is worthless on libcrux I think, you should ask review from i.e. @jschneider-bensch

I just realized that as well.

franziskuskiefer (approved):

Issue #1065, opened 2025-07-17
[PSQ] Full PQ mode


jschneider-bensch:
Add ML-DSA 65 signature options.

PR #1108, opened 2025-08-20
[PSQ] Allow using Classic McEliece for PQ KEM

jschneider-bensch:
This PR does four things:
1. Clean up, some renaming and more documentation for  more of the new PSQ protocol. In particular, the doc test at the crate level showcases how a responder can be queried for its ciphersuite and how to use the response to perform a registration with the responder-supported ciphersuite. 
2. Move the old PSQ code to a feature-guarded `v1` module, with the exception of some of the Classic McEliece code.
3. Introduce an abstraction for PQ KEM keys, which allows to use either ML-KEM 768 or Classic McEliece 460896f as the PQ KEM.
4. Introduce names for all supported ciphersuites, and those that will be supported in the future (AES-GCM ones)

In terms of possible "configurations" for 3. I went the simplest option, i.e. an initiator can have either an ML-KEM or CMC longterm public key of the responder, and the responder can have either an ML-KEM or CMC key pair. If there is a mismatch, i.e. the initiator sends a registration message for which the responder does not have a matching decapsulation key, then the responder will ignore that message and return to its initial state.

**After this PR public-facing APIs should be stable**, I hope, with possible exception of changes from #1065 that I can't yet anticipate.

---
**TODO**
- [x] Ciphersuite builder
- [x] Test configuration mismatch


libcrux-psq/Cargo.toml
```diff
@@ -24,9 +24,9 @@ libcrux-sha2 = { version = "=0.0.3", path = "../sha2" }
 classic-mceliece-rust = { version = "3.1.0", features = [
     "mceliece460896f",
     "zeroize",
-], optional = true }
+]}
```
franziskuskiefer: Shouldn't this still be optional?

libcrux-psq/Cargo.toml
```diff
@@ -46,17 +46,22 @@ log = "0.4.26"
 pretty_env_logger = "0.5.0"
 
 [features]
-classic-mceliece = ["dep:classic-mceliece-rust", "rand_old"]
+legacy = []
```
franziskuskiefer: Since this will still be used, we should maybe just call it `v1`? `legacy` sounds a little negative.

libcrux-psq/Cargo.toml
```diff
@@ -46,17 +46,22 @@ log = "0.4.26"
...
+# classic-mceliece = ["dep:classic-mceliece-rust", "rand_old"]
```
franziskuskiefer: Should this go back in?

libcrux-psq/src/handshake/responder.rs
```diff
@@ -393,3 +414,37 @@ impl<'a, Rng: CryptoRng> IntoSession for Responder<'a, Rng> {
...
+    responder_longterm_ecdh_sk: &DHPrivateKey,
+    tx2: &Transcript,
+) -> Result<AEADKey, Error> {
+    let responder_ikm = K2IkmQuery {
+        k0,
+        g_xs: &DHSharedSecret::derive(responder_longterm_ecdh_sk, initiator_ephemeral_ecdh_pk)?,
+        g_xy: &DHSharedSecret::derive(responder_ephemeral_ecdh_sk, initiator_ephemeral_ecdh_pk)?,
+    };
+
+    AEADKey::new(&responder_ikm, tx2).map_err(|e| e.into())
```
franziskuskiefer: `?` call `into` for you. So something like this would work as well.

```suggestion
    Ok(AEADKey::new(&responder_ikm, tx2)?)
```

libcrux-psq/src/handshake/pqkem.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+impl<'a> From<SharedSecret<'a>> for PQSharedSecret<'a> {
+    fn from(value: SharedSecret<'a>) -> Self {
+        Self::ClassicMcEliece(value)
+    }
+}
+
+impl<'a> PQPublicKey<'a> {
+    /// Encapsulate a PQ-shared secret towards the given PQ-KEM public key
+    pub(crate) fn encapsulate(&self, rng: &mut impl CryptoRng) -> (PQCiphertext, PQSharedSecret) {
```
franziskuskiefer, line 124: You should add lifetimes here to avoid warnings.

libcrux-psq/src/handshake/pqkem.rs
```diff
@@ -0,0 +1,176 @@
...
+            sk: &value.sk,
+        }
+    }
+}
+
+#[derive(TlsSerialize, TlsDeserialize, TlsSize)]
+#[repr(u8)]
+/// A PQ-KEM key encapsulation
+pub enum PQCiphertext {
+    MlKem768(MlKem768Ciphertext),
```
franziskuskiefer, line 88: This is pretty bad for the CMCE ciphertext (156 vs 1088). But maybe that's fine? Or would it make sense to use references here as well?

franziskuskiefer (commented):
Just some early thoughts

libcrux-psq/benches/psq_v2.rs
```diff
@@ -27,7 +30,7 @@ fn query<const PQ: bool>(c: &mut Criterion) {
 
     // External setup
     let responder_ecdh_keys = DHKeyPair::new(&mut rng);
-    let responder_pq_keys = PQKeyPair::new(&mut rng);
+    let responder_pq_keys = libcrux_ml_kem::mlkem768::rand::generate_key_pair(&mut rng);
```
franziskuskiefer: Should the `into` once happen up here already?

libcrux-psq/src/handshake/builder.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /// Set the long-term ECDH key pair.
+    pub fn longterm_ecdh_keys(mut self, longterm_ecdh_keys: &'a DHKeyPair) -> Self {
+        self.longterm_ecdh_keys = Some(longterm_ecdh_keys);
+        self
+    }
+
+    /// Set the long-term PQ key pair.
+    pub fn longterm_pq_keys(mut self, longterm_pq_keys: impl Into<PQKeyPair<'a>>) -> Self {
```
franziskuskiefer: We should make the API consistent. The DH above gets just a reference. We should pick a style and stick to it for both types of keys.

libcrux-psq/tests/query.rs
```diff
@@ -14,18 +17,18 @@ fn query() {
     // External setup
     let responder_ecdh_keys = DHKeyPair::new(&mut rng);
 
-    let responder_pq_keys = PQKeyPair::new(&mut rng);
+    let responder_pq_keys = libcrux_ml_kem::mlkem768::rand::generate_key_pair(&mut rng);
```
franziskuskiefer: Same comment as for the builder below. We should pick a style for both. Here in the call site we just see the difference as well.
I think either could be fine. Either a generic `key_gen` with an algorithm identifier, or explicitly generating the keys. But I think I like one function for all a little better.

libcrux-psq/src/handshake/builder.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // properties
+
+    /// Set the context.
+    pub fn context(mut self, context: &'a [u8]) -> Self {
+        self.context = context;
+        self
+    }
+
+    /// Set the inner AAD.
```
franziskuskiefer, line 49: Somewhat unrelated to this PR but I think the aad setters need more documentation to tell the user what the outer and inner aads are.

libcrux-psq/src/handshake/builder.rs
```diff
@@ -0,0 +1,151 @@
...
+        QueryInitiator::new(responder_ecdh_pk, self.context, self.outer_aad, self.rng)
+    }
+
+    /// Build a new [`RegistrationInitiator`].
+    ///
+    /// This requires that a `longterm_ecdh_keys` and a `peer_longterm_ecdh_pk`
+    /// is set.
+    /// It also uses the `context`, `inner_aad`, `outer_aad`, and
+    /// `peer_longterm_pq_pk`.
+    pub fn build_registration_initiator(self) -> Result<RegistrationInitiator<'a, Rng>, Error> {
```
franziskuskiefer: I think there's no way to ensure that we set a pq key if we want a pq ciphersuite, right?
If the caller forgets to set the pq key, it'll just do dh only. That's not great.
That also goes back to the question about how to generate keys and ciphersuites.

I suggest that we add a ciphersuite in the `new` of the builder and that way enforce that the correct set of keys are available.
This allows then generating keys based on it. That's not so important though. It's long term keys and it's not clear whether that would actually be used.

libcrux-psq/Cargo.toml
jschneider-bensch: Good question! I would like it to be, but I built the PQ KEM selection with an enum now, and that would then have to be `non_exhaustive`, no?
I recall now we had a similar situation in the original PSQ and then resolved to do this with a commonly implemented trait, instead of an enum. Ideally that would be one of the KEM traits from `libcrux-traits`... I'll see if I can implement that for the CMC wrapper here.

franziskuskiefer (commented):
Some thoughts

libcrux-psq/Cargo.toml
jschneider-bensch: Sure!
I just wanted to indicate that we're probably not going to prioritize further development of that version. It's a funny thing that "legacy" has negative connotations in software, since in other contexts it's what people hope to be remembered by. :D

libcrux-psq/Cargo.toml
jschneider-bensch: Yes, I temporarily disabled that too work on the new version as if CMC was unconditionally included, but in any case the old protocol needs the feature.

libcrux-psq/src/handshake/builder.rs
jschneider-bensch: I agree, it should be an explicit choice to use DH only!

jschneider-bensch:
I've changed the proposal here to use a trait mechanism for implementing ciphersuites. This makes it easier to pull out Classic McEliece into a non-default feature.

The downside is that in the way I've implemented it here, we need to pre-serialize the PQ encapsulation, to make the `HandshakeMessage` types oblivious to the used ciphersuite and we need to do the same with the PQ encapsulation key in another place, since we don't yet have a blanket impl of `Serialize` for `Option<&'a T>` from `T: Serialize`. I tried getting around this by introducing a separate associated type for the reference to the encapsulation key, that would also require a `Serialize` bound, but could not get it to work. At some point I needed to go from `Option<C::EncapsulationKey>` to `Option<C::EncapsulationKeyRef>` and I couldn't figure out which (if any) bounds would allow me to do that.

One more thing that is missing here is a builder for the ciphersuites, so now they are statically constructed in the tests.

libcrux-psq/src/handshake/ciphersuite.rs
```diff
@@ -0,0 +1,747 @@
...
+    fn from(value: &'a PublicKey) -> Self {
+        Self::CMC(value)
+    }
+}
+
+#[derive(TlsSize, TlsSerializeBytes)]
+#[repr(u8)]
+pub enum DynamicSharedSecret<'a> {
+    None,
+    MlKem(MlKemSharedSecret),
```
franziskuskiefer: There are a couple clippy warnings, including the boxing for the enums.

libcrux-psq/src/handshake/ciphersuite.rs
```diff
@@ -0,0 +1,747 @@
...
+            #[cfg(not(feature = "classic-mceliece"))]
+            InitiatorCiphersuite::X25519_ClassicMcEliece_ChaChaPoly_HkdfSha256(_) => {
+                panic!("unsupported ciphersuite")
+            }
+        }
+    }
+}
+
+impl<'a> CiphersuiteBuilder<'a> {
+    pub fn new() -> Self {
```
franziskuskiefer: For builders you usually add a `builder()` function to the resulting object as well to make it more seamless. Maybe that's difficult here because of the different results. Also not too important. Just a note.

libcrux-psq/src/handshake/ciphersuite.rs
```diff
@@ -0,0 +1,747 @@
...
+            #[cfg(not(feature = "classic-mceliece"))]
+            CiphersuiteName::X25519ClassicMcElieceChachaPolyHkdfSha256 => {
+                panic!("unsupported cipersuite")
+            }
+        }
+    }
+
+    pub fn finish_responder(
+        self,
+        name: CiphersuiteName,
```
franziskuskiefer: I'd expect the name in `new`. This is fine too. I was just wondering where the ciphersuite is.

libcrux-psq/src/handshake/ciphersuite.rs
```diff
@@ -0,0 +1,747 @@
...
+    ),
+    #[cfg(not(feature = "classic-mceliece"))]
+    X25519_ClassicMcEliece_ChaChaPoly_HkdfSha256(DummyCiphersuite),
+}
+
+#[allow(non_camel_case_types)]
+pub enum ResponderCiphersuite<'a> {
+    X25519_ChaChaPoly_HkdfSha256(ResponderX25519ChaChaPolyHkdfSha256<'a>),
+    X25519_MlKem768_ChaChaPoly_HkdfSha256(ResponderX25519MlKem768ChaChaPolyHkdfSha256<'a>),
+    #[cfg(feature = "classic-mceliece")]
```
franziskuskiefer: Not the nicest, but I'm not sure what kind of trouble this may still be able to do. So I think this is fine. When everything is boxed it may also be equivalent in memory.

libcrux-psq/src/handshake/ciphersuite.rs
```diff
@@ -0,0 +1,747 @@
...
+            )),
+            #[cfg(feature = "classic-mceliece")]
+            ResponderCiphersuite::X25519_ClassicMcEliece_ChaChaPoly_HkdfSha256(
+                responder_x25519_cmc_cha_cha_poly_hkdf_sha256,
+            ) => Some(DynamicEncapsulationKeyRef::CMC(
+                responder_x25519_cmc_cha_cha_poly_hkdf_sha256.longterm_pq_encapsulation_key,
+            )),
+            #[cfg(not(feature = "classic-mceliece"))]
+            ResponderCiphersuite::X25519_ClassicMcEliece_ChaChaPoly_HkdfSha256(_) => {
+                panic!("unsupported ciphersuite")
```
franziskuskiefer: Maybe let's return `None` in these cases and print an error + `debug_assert`. Just to avoid panics at runtime in deployments.

jschneider-bensch:
While cleaning this up I hit upon a weird issue where an enum
```rust
#[derive(TlsSize, TlsDeserialize, TlsSerialize)]
#[repr(u8)]
pub enum DynamicCiphertext {
    MlKem(MlKem768Ciphertext),
    #[cfg(feature = "classic-mceliece")]
    CMC(Ciphertext),
    #[cfg(not(feature = "classic-mceliece"))]
    CMC,
}
```
is serialized or deserialized incorrectly. I had it print out bytes at initiator and responder, and the bytes agree, but when a `DynamicCiphertext::MlKem(_)` is serialized, it is deserialized as a `DynamicCiphertext::CMC`. (This is for the case where the feature is off, did not test the other case yet.) Perhaps the derive macro does some unexpected thing with the conditional compilation going on here. I tried disambiguation, giving the two CMC variant different tags in any case, but the issue persisted.
Next thing I'll try is implementing the `tls_codec` traits manually to see if that helps, or if the issue lies elsewhere.

jschneider-bensch:
I'm hopeful this PR is finally in good shape for a proper review:
- I would like the APIs in here to be what the stable API going forward should be.
- I named all ciphersuites (even ones that are not yet supported).
- I expanded the tested scenarios to include various ciphersuite mismatches (compatible and incompatible)
- The top-level docs now show an example of an initiator dynamically receiving ciphersuite information from the responder via query mode and subsequently performing a registration with the responder.

libcrux-psq/src/handshake/ciphersuite/types.rs
```diff
@@ -0,0 +1,59 @@
...
+#[cfg(feature = "classic-mceliece")]
+impl<'a> From<&'a PublicKey> for PQEncapsulationKey<'a> {
+    fn from(value: &'a PublicKey) -> Self {
+        Self::CMC(value)
+    }
+}
+
+#[derive(TlsSize, TlsSerializeBytes)]
+#[repr(u8)]
+pub enum PQSharedSecret<'a> {
```
franziskuskiefer: The lifetime here is only for the cmc case, which should always be `'static` right? So could we skip the lifetime here?

libcrux-psq/tests/registration.rs
```diff
@@ -89,60 +180,175 @@ fn registration(pq: bool) {
...
+            CiphersuiteName::X25519_CLASSICMCELIECE_CHACHA20POLY1305_HKDFSHA256,
+            CiphersuiteName::X25519_CLASSICMCELIECE_CHACHA20POLY1305_HKDFSHA256,
+        ),
+        Err(TestError::Builder(BuilderError::UnsupportedCiphersuite))
+    );
+}
+
+// Building any of the AES suites should fail until they are supported.
+#[test]
+fn unsupported_aes() {
```
franziskuskiefer, line 337: This test takes forever to run. Can we skip the cmc variants to speed this up or have some other way to check this?

libcrux-psq/src/session/session_key.rs
```diff
@@ -0,0 +1,74 @@
...
+        &SerializeBytes::tls_serialize(&key).map_err(Error::Serialize)?,
+        SESSION_KEY_INFO,
+    )
+    .map_err(|_| Error::CryptoError)?;
+
+    Ok(
+        session_id.try_into().map_err(|_| Error::CryptoError)?, // We don't expect this to fail, unless HDKF gave us the wrong output length
+    )
+
+    // let prk = libcrux_hkdf::extract(
```
franziskuskiefer: What's up with his code down here?

libcrux-psq/src/session/session_key.rs
```diff
@@ -0,0 +1,74 @@
...
+    libcrux_hkdf::sha2_256::hkdf(
+        &mut session_id,
+        SESSION_KEY_SALT,
+        &SerializeBytes::tls_serialize(&key).map_err(Error::Serialize)?,
+        SESSION_KEY_INFO,
+    )
+    .map_err(|_| Error::CryptoError)?;
+
+    Ok(
+        session_id.try_into().map_err(|_| Error::CryptoError)?, // We don't expect this to fail, unless HDKF gave us the wrong output length
```
franziskuskiefer: This `try_into` is useless.

libcrux-psq/src/handshake.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[derive(TlsSerializeBytes, TlsSize)]
+    struct K1Ikm<'a> {
+        k0: &'a AEADKey,
+        ecdh_shared_secret: &'a DHSharedSecret,
+        pq_shared_secret: Option<PQSharedSecret<'a>>,
+    }
+
+    let ecdh_shared_secret = DHSharedSecret::derive(own_longterm_key, peer_longterm_pk)?;
+
+    // XXX: This makes clippy unhappy, but is a lifetime error for feature `classic-mceliece` if we return directlyr
```
franziskuskiefer, line 151: Hm, seems to work fine for me (to fix the clippy warning).

libcrux-psq/src/aead.rs
```diff
@@ -0,0 +1,141 @@
...
+        libcrux_hkdf::sha2_256::hkdf(
+            &mut key,
+            &[],
+            &ikm.tls_serialize().map_err(AEADError::Serialize)?,
+            &info.tls_serialize().map_err(AEADError::Serialize)?,
+        )
+        .map_err(|_| AEADError::CryptoError)?;
+
+        Ok(AEADKey(
+            key.try_into().map_err(|_| AEADError::CryptoError)?,
```
franziskuskiefer: Unnecessary `try_into`.

libcrux-psq/tests/registration.rs
jschneider-bensch: Indeed, it takes so long because we repeatedly do the CMC key generation...

libcrux-psq/src/handshake/ciphersuite/types.rs
jschneider-bensch: Yep, you're right :+1:

libcrux-psq/tests/registration.rs
jschneider-bensch: I changed the tests to re-use one `CommonSetup` across all tests, which should reduce the cost for CMC to the minimum for these tests.

libcrux-psq/src/session/session_key.rs
jschneider-bensch: Sorry, merge leftover (the HKDF API had changed on main).

libcrux-psq/src/handshake.rs
jschneider-bensch: Right, I can't reproduce it anymore either... weird, but fine with me.

libcrux-psq/src/handshake/builder.rs
jschneider-bensch: Right, I made an issue to improve docs overall #1203, which should be addressed before release.

libcrux-psq/tests/query.rs
jschneider-bensch: At the moment we generate the keys explicitly, but we may indeed want to offer a crate-local way of generating the keys with a generic `key_gen` in the future, so users don't have to pull in another crate explicitly.

jschneider-bensch: #1204

libcrux-psq/src/handshake/ciphersuite.rs
jschneider-bensch: I made #1205 to keep track of this.

libcrux-psq/src/handshake/ciphersuite.rs
jschneider-bensch: Indeed, I've changed it so that attempting to build an unsupported ciphersuite will return an Error instead.
So it's not the earliest possible error (which would be in `new`), but it means that any of the other methods on unsupported ciphersuites must be unreachable, since you can't construct unsupported ciphersuites.

PR #1109, opened 2025-08-20
Remove unsafe expression on safe AVX2 instructions

parrot7483:
The changed AVX2 simd instructions in this PR are no longer unsafe. The Compiler gives a warning when using them in an unsafe block.

### DIFF
--- a/libcrux-intrinsics/src/avx2.rs
+++ b/libcrux-intrinsics/src/avx2.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(hax, allow(unused_unsafe))]
+
 #[cfg(all(target_arch = "x86", not(hax)))]
 pub use core::arch::x86::*;
 #[cfg(all(target_arch = "x86_64", not(hax)))]


franziskuskiefer (changes requested):
CI is failing

parrot7483:
I miss some understanding on how to use compilation targets and configuration flags here. 

For example the [_mm256_sign_epi32](https://doc.rust-lang.org/stable/core/arch/x86/fn._mm256_sign_epi32.html) function is safe and is for the `x86` or `x86-64` target and feature `avx2`. Because the wrapping function uses `#[inline(always)]`, one can not use `#[target_feature(enable = "avx2")] `

```rust
#[inline(always)]
#[hax_lib::fstar::before(r#"[@@ "opaque_to_smt"]"#)]
pub fn mm256_sign_epi32(a: Vec256, b: Vec256) -> Vec256 {
    _mm256_sign_epi32(a, b)
}
```

Function [_mm256_setzero_si256](https://doc.rust-lang.org/stable/core/arch/x86/fn._mm256_setzero_si256.html) is for the same target but feature `avx`. It is also safe. Some other functions in `avx2.rs` might be for the older SSE2 instruction set extension. Some are also unsafe.

Is the correct way to use these SIMD functions for different features to put the correct `#[cfg(target_feature = "...")]` on each of them? This way one can keep using `#[inline(always)]`. The user of the library would then set the needed features. 

The motivation for this being that safe functions are not wrapped in an unsafe block. 

Maybe its also a good idea to separate them into different modules? A function for SSE2 being in `avx2.rs` might be confusing (at least for me).

parrot7483:
CPU intrinsics [require an unsafe block](https://doc.rust-lang.org/stable/core/arch/index.html#overview) because they are hardware-dependent. Running them on an unsupported CPU will crash the program. Only when using the HAX target we get a warning.

franziskuskiefer (approved):
CI is failing

PR #1112, opened 2025-08-21
Refactor benchmarks to support comments too

keks:


keks:
Unfortunately simply disabling the other actions didn't work, because then Github won't put up the benchmarks job. I'll see if there is something better I can do on Monday.

Issue #1107, opened 2025-08-20
API overhaul


franziskuskiefer:
Meta issue to track updating APIs for v0.1.

List of crates and APIs
- [ ] ecdh
  - [x] p-curves
  - [x] x25519
- [x] kem
  - [ ] ecdh
  - [x] ml-kem
- [ ] aead
  - [x] chacha-poly
  - [x] xchacha-poly
  - [ ] aes-gcm
- [ ] signatures
  - [ ] ed25519
  - [ ] ecdsa
  - [ ] rsa-pss?
  - [ ] ml-dsa
- [x] Hash
  - [x] Sha2
  - [x] Sha3
  - [x] Blake2
- [ ] KDF
  - [ ] HKDF - https://github.com/cryspen/libcrux/pull/1177
- [ ] MAC
  - [ ] keyed blake2
  - [ ] HMAC

franziskuskiefer:
Investigate
- [ ] kem
- [ ] ecdh

franziskuskiefer:
#1241  should close this.

PR #1115, opened 2025-08-25
AEAD: Key-Centric API with Multiplexing Support

keks:
This PR explores how we can have a generic key-centric API that also allows us to multiplex algorithms efficiently. This is not done, and we only experiment with AEAD for now.

- The code in traits is for all the generic stuff, including traits and keys.
- The code in chachapoly implements the traits so we can test that this is actually instantiable and runnable.
- The code in src/aead.rs is for testing multiplexing.

Multiplexing is only implemented for keys with borrowed types to avoid differences in key size to make short keys long.

Action items from call on 2025-09-03:
- [x] Add AeadConsts trait
- [x] Rename referece types (Key, Tag, None) to ...Ref
- [x] ref style encrypt: don't return tag
- [x] type aliases for Key, Tag, KeyRef, .. types in algo modules
- [x] use named structs for the ref types
- [x] use secret types for Tag, Nonce
- [x] Drop AeadExt and add methods to Aead as default impls
- [x] make a crate libcrux-aead instead of putting the multiplexed stuff into the libcrux crate

Works towards #1107

### DIFF
--- a/chacha20poly1305/src/impl_aead_trait.rs
+++ b/chacha20poly1305/src/impl_aead_trait.rs
@@ -142,7 +142,7 @@ mod impl_xchachapoly {
             let mut new_nonce = [0u8; super::NONCE_LEN];
 
             crate::xchacha20_poly1305::derive(key, nonce, &mut subkey, &mut new_nonce);
-            ChaCha20Poly1305::decrypt(plaintext, &subkey, &new_nonce, &aad, ciphertext, tag)
+            ChaCha20Poly1305::decrypt(plaintext, &subkey, &new_nonce, aad, ciphertext, tag)
         }
     }
 
--- a/chacha20poly1305/src/xchacha20_poly1305.rs
+++ b/chacha20poly1305/src/xchacha20_poly1305.rs
@@ -76,8 +76,8 @@ pub(crate) fn derive(
     let (nonce_16, nonce_8) = nonce_in.split_at(16);
     let nonce_16: &[u8; 16] = nonce_16.try_into().unwrap();
 
-    hchacha20(key_in, &nonce_16, key_out);
-    nonce_out[4..].copy_from_slice(&nonce_8);
+    hchacha20(key_in, nonce_16, key_out);
+    nonce_out[4..].copy_from_slice(nonce_8);
 }
 
 /// Convert the `key` and `nonce` into the subkey.


chacha20poly1305/src/impl_aead_trait.rs
```diff
@@ -112,4 +120,55 @@ mod impl_xchachapoly {
...
+        XNONCE_LEN
+    );
+}
+
+#[cfg(test)]
+mod tests {
+    use libcrux_traits::aead::typed_owned;
+    use libcrux_traits::aead::typed_refs;
+
+    type Key = typed_owned::Key<super::ChaCha20Poly1305>;
```
franziskuskiefer, line 163: Should the crate expose these types?

traits/src/aead/typed_owned.rs
```diff
@@ -0,0 +1,207 @@
...
+
+// These are the types for the arguments to AEAD. I wonder if it makes sense to do stuff like
+// implementing a random constructor for Nonce for all LEN in 24..64 or so
+//
+// There also is the question whether we want these to be byte-oriented or whether we want to just
+// make these associated types. That would mean that we'd have to define them separately for all
+// implementations.
+
+#[repr(transparent)]
+pub struct Key<Algo: Aead>(Algo::Key);
```
franziskuskiefer, line 16: Related to the multiplexing question. Should there be a function to generate a new key based on an algorithm identifier? Also see comment on `aead.rs`. But maybe I misunderstand how this is supposed to be used.

traits/src/aead/typed_refs.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// Performs AEAD encryption. If algo is multiplexed, it checks that the algorithms specified
+    /// in the key, nonce and tag are consistent.
+    pub fn encrypt(
+        &self,
+        ciphertext: &mut [u8],
+        tag: TagMut<'a, Algo>,
+        nonce: Nonce<'a, Algo>,
+        aad: &[u8],
+        plaintext: &[U8],
+    ) -> Result<Tag<'a, Algo>, EncryptError> {
```
franziskuskiefer: I don't think I understand why there's a `tag` argument and a `Tag` return value. That seems a little odd.

traits/src/aead/typed_refs.rs
```diff
@@ -0,0 +1,505 @@
...
+// implementing a random constructor for Nonce for all LEN in 24..64 or so
+//
+// There also is the question whether we want these to be byte-oriented or whether we want to just
+// make these associated types. That would mean that we'd have to define them separately for all
+// implementations.
+
+/// A Key with the given algorithm. The bytes are borrowed. Contains a marker for which
+/// algorith the key is to be used with.
+#[derive(Clone, Copy)]
+pub struct Key<'a, Algo>(Algo, &'a [U8]);
```
franziskuskiefer: Wouldn't associated types work here as well? The lifetimes are a little annoying. But then we wouldn't need to carry around the `Algo`. Not sure about the ergonomics in that case.

src/aead.rs
```diff
@@ -11,11 +11,126 @@
 //! 128-bit SIMD implementation is used.
```
franziskuskiefer: If we put this in here, we should update the docs and clean up the rest of the code.

chacha20poly1305/src/impl_aead_trait.rs
```diff
@@ -6,12 +6,14 @@ use crate::{KEY_LEN, NONCE_LEN, TAG_LEN};
 ///
 /// Note: Plaintext, ciphertext and aad need to be at most [`u32::MAX`] long, due to limitations of
 /// the implementation.
+#[derive(Clone, Copy, PartialEq)]
 pub struct ChaCha20Poly1305;
```
franziskuskiefer, line 10: Should there be a static item in here for this struct? Then the callers wouldn't need to instantiate the struct first.

src/aead.rs
```diff
@@ -11,11 +11,126 @@
...
+        match self {
+            Aead::ChaCha20Poly1305 => 12,
+            Aead::XChaCha20Poly1305 => 24,
+        }
+    }
+
+    fn encrypt<'a>(
+        &self,
+        ciphertext: &mut [u8],
+        tag: libcrux_traits::aead::typed_refs::TagMut<'a, Self>,
```
franziskuskiefer: I'm not sure I understand how this should be used.
How does the caller create these arguments?

franziskuskiefer (commented):
Generally looks pretty good already. But I'm not sure about the multiplexing story.

There are a bunch of lifetimes missing that generate warnings.

traits/src/aead/typed_owned.rs
keks: We have that in the Multiplexes trait. But we only support multiplexing for reference-style keys, to avoid the issue with different lengths.

traits/src/aead/typed_refs.rs
keks: We still need the Algo for multiplexing. I think we should be able to move to associated types though.

traits/src/aead/typed_refs.rs
keks: Right, I should have put a comment here.

That's what I initially had, but the caller didn't have access to the TagMut anymore because it was moved into the encrypt function, so they couldn't get a Tag out again. I experimented with making the tag argument `&mut TagMut<_>`, but that lead to more problems and was a bit weird because it's a double reference. That's why we move ownership of the &mut (in the form of TagMut) inside encrypt, and then return a Tag, which is immutable.

Does that make sense?

I suppose an alternative design would be to encourage defining an owned Tag and then allowing to get TagMut an TagRef out of it. We could them name all the others KeyRef, NonceRef and so on as well. Maybe prettier? Not sure.

src/aead.rs
keks: This is just to check/show that multiplexing works as intended. But yeah wherever we put it, it should have docs.

chacha20poly1305/src/impl_aead_trait.rs
keks: The type name also serves as a zero-cost constructor, so you can just do `ChaCha20Poly1305.encrypt(...)`. No static item needed.

chacha20poly1305/src/impl_aead_trait.rs
keks: I am undecided. I think it would be neat if the caller didn't really need to name these types, but we have to see if we can make that ergonomic.

src/aead.rs
keks: Using the methods of the Multiplexes trait, see the test at the end of the file.

keks:
I didn't see any lifetime-related warnings. I only saw one warning, which was due to an unneeded import and got rid of that.

franziskuskiefer:
> I didn't see any lifetime-related warnings. I only saw one warning, which was due to an unneeded import and got rid of that.

You can see it in the [CI log](https://github.com/cryspen/libcrux/actions/runs/17293882531/job/49087338656?pr=1115#step:12:104) as well.

keks:
Got it, these warnings were added in 1.89 and I was on an older version. Should we make CI fail if there are compilation warnings?

keks:
@franziskuskiefer  Should we meet to discuss this?

traits/src/aead/typed_refs.rs
```diff
@@ -0,0 +1,505 @@
...
+        tag: Tag<'a, Self>,
+    ) -> Result<(), DecryptError>;
+}
+
+/// Provides convenience functions. Has a blanket impl for all Aead.
+pub trait AeadExt: Aead {
+    /// Creates a new key given the algorithm.
+    fn new_key<'a>(self, key: &'a [U8]) -> Result<Key<'a, Self>, WrongLengthError>;
+    /// Creates a new tag given the algorithm.
+    fn new_tag<'a>(self, tag: &'a [u8]) -> Result<Tag<'a, Self>, WrongLengthError>;
```
franziskuskiefer: Is there a reason not all of these are secret bytes `U8`?

keks:
We discussed this on a call today. I updated the PR above with the action items and will mark the review conversations as resolved, because I think we talked about all of them.

.github/workflows/aead-build_test.yml
```diff
@@ -0,0 +1,128 @@
+name: AEAD - Build & Test
+
+on:
+  merge_group:
+  push:
+    branches: ["main", "dev"]
```
franziskuskiefer, line 6: Let's start adding paths and not running on main here already? cc [#1129](https://github.com/cryspen/libcrux/issues/1129)

.github/workflows/aead-build_test.yml
```diff
@@ -0,0 +1,128 @@
...
+
+jobs:
+  build:
+    if: ${{ github.event_name != 'merge_group' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        bits: [32, 64]
+        os:
+          - macos-13 # Intel mac
```
franziskuskiefer: Only build here for intel mac as well?

traits/src/aead/consts.rs
```diff
@@ -0,0 +1,5 @@
+pub trait AeadConsts {
```
franziskuskiefer, line 2: Let's add some doc comments.

chacha20poly1305/src/impl_aead_trait.rs
```diff
@@ -72,13 +84,29 @@ mod impl_chachapoly {
     }
 
     libcrux_traits::aead::slice::impl_aead_slice_trait!(ChaCha20Poly1305 => KEY_LEN, TAG_LEN, NONCE_LEN);
```
franziskuskiefer, line 89: Let's add some comments here what this is implementing.

chacha20poly1305/src/lib.rs
```diff
@@ -11,6 +26,22 @@ pub const TAG_LEN: usize = 16;
...
 pub const NONCE_LEN: usize = 12;
 
+/// An owned ChaCha20Poly1305 key.
+pub type Key = typed_owned::Key<ChaCha20Poly1305>;
+/// An owned ChaCha20Poly1305 tag.
+pub type Tag = typed_owned::Tag<ChaCha20Poly1305>;
+/// An owned ChaCha20Poly1305 nonce.
+pub type Nonce = typed_owned::Nonce<ChaCha20Poly1305>;
+
+/// A referece to a ChaCha20Poly1305 key.
```
franziskuskiefer:
```suggestion
/// A reference to a ChaCha20Poly1305 key.
```

chacha20poly1305/src/lib.rs
```diff
@@ -11,6 +26,22 @@ pub const TAG_LEN: usize = 16;
...
+pub type KeyRef<'a> = typed_refs::KeyRef<'a, ChaCha20Poly1305>;
+/// A referece to a ChaCha20Poly1305 tag.
```
franziskuskiefer:
```suggestion
/// A reference to a ChaCha20Poly1305 tag.
```

chacha20poly1305/src/lib.rs
```diff
@@ -11,6 +26,22 @@ pub const TAG_LEN: usize = 16;
...
+pub type TagRef<'a> = typed_refs::TagRef<'a, ChaCha20Poly1305>;
+/// A mutable referece to a ChaCha20Poly1305 tag.
```
franziskuskiefer:
```suggestion
/// A mutable reference to a ChaCha20Poly1305 tag.
```

chacha20poly1305/src/lib.rs
```diff
@@ -11,6 +26,22 @@ pub const TAG_LEN: usize = 16;
...
+pub type TagMut<'a> = typed_refs::TagMut<'a, ChaCha20Poly1305>;
+/// A referece to a ChaCha20Poly1305 nonce.
```
franziskuskiefer:
```suggestion
/// A reference to a ChaCha20Poly1305 nonce.
```

chacha20poly1305/src/xchacha20_poly1305.rs
```diff
@@ -7,6 +7,25 @@ use crate::{
...
+pub use crate::impl_aead_trait::XChaCha20Poly1305;
+
+/// An owned XChaCha20Poly1305 key.
+pub type Key = typed_owned::Key<XChaCha20Poly1305>;
+/// An owned XChaCha20Poly1305 tag.
+pub type Tag = typed_owned::Tag<XChaCha20Poly1305>;
+/// An owned XChaCha20Poly1305 nonce.
+pub type Nonce = typed_owned::Nonce<XChaCha20Poly1305>;
+
+/// A referece to a XChaCha20Poly1305 key.
```
franziskuskiefer: Same typos here

crates/primitives/aead/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+mod multiplexed;
```
franziskuskiefer, line 154: This crate could use some doc comments. Including some usage examples.

crates/primitives/aead/src/multiplexed.rs
```diff
@@ -0,0 +1,190 @@
...
+/// A multiplexed AEAD.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Aead {
+    #[cfg(feature = "chacha20poly1305")]
+    ChaCha20Poly1305,
+    #[cfg(feature = "xchacha20poly1305")]
+    XChaCha20Poly1305,
+}
+
+/// A referece to a multiplexed AEAD key.
```
franziskuskiefer: Same typos

crates/primitives/aead/src/lib.rs
```diff
@@ -0,0 +1,9 @@
...
+
```
franziskuskiefer, line 155: Should this be `no_std` as well?

chacha20poly1305/src/lib.rs
```diff
@@ -1,7 +1,22 @@
 #![no_std]
```
franziskuskiefer, line 64: Let's add some usage example and doc comments here too.

franziskuskiefer (approved):
Thanks, lgtm with a few nits.

- I see some clippy warnings in the chacha crate (unrelated to this PR). We should fix them and make sure we run clippy on ci.

PR #1116, opened 2025-08-26
C: minor update

franziskuskiefer:
A little cleanup for https://github.com/cryspen/home/issues/320.

### DIFF
--- a/libcrux-ml-kem/cg/boring/eurydice_glue.h
+++ b/libcrux-ml-kem/cg/boring/eurydice_glue.h
@@ -12,7 +12,8 @@
 // TODO(crbug.com/404286922): Use designated initializers instead of /*name=*/
 // comments when switching to C++20.
 
-#pragma once
+#ifndef EURYDICE_HEADER_EURYDICE_GLUE_H
+#define EURYDICE_HEADER_EURYDICE_GLUE_H
 
 #include <inttypes.h>
 #include <stdbool.h>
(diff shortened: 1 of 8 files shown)


libcrux-ml-kem/cg/boring/eurydice_glue.h
```diff
@@ -61,38 +78,19 @@ using std::type_identity_t;
 //   arithmetic based on the type of the address)
 // - if you need to use `len` for a C style function (e.g. memcpy, memcmp), you
 //   need to multiply it by sizeof t, where t is the type of the elements.
-//
-// EURYDICE_SLICE computes `NULL + start`. If your compiler is ok with that,
-// using a `NULL` in `ptr` is ok. If you need to support older compilers that
-// do not support arithemtic operations on `NULL`, ensure that `NULL` is never
-// used here.
```
keks, line 68: After reading the comments on Gerrit and the documentation around how slices work in Rust I am not sure if this is allowed to be NULL. In the code generated by Rust it isn't, but I am not sure how that translates to the C world. If NULL is forbidden, I think this should be noted here.

franziskuskiefer: It is not forbidden. It used to in old compilers because of arithmetic on `NULL` used to be undefined. But with all the C++ compilers we care about here, it's allowed.

keks: I understand the issues with C compilers. I was worried we break something by not honoring Rust's assumptions. But I think that probably only matters when we use Rust codegen.

PR #1120, opened 2025-08-28
SHA3 SIMD tests

parrot7483:
Tests for https://github.com/cryspen/libcrux/issues/966

### DIFF
--- a/libcrux-sha3/src/generic_keccak.rs
+++ b/libcrux-sha3/src/generic_keccak.rs
@@ -191,6 +191,7 @@ impl<const N: usize, T: KeccakItem<N>> KeccakState<N, T> {
     where
         Self: Absorb<N>,
     {
+        #[cfg(not(eurydice))]
         debug_assert!(blocks.iter().all(|buf| buf.len() == blocks[0].len()));
 
         self.load_block::<RATE>(blocks, start);
@@ -207,6 +208,8 @@ impl<const N: usize, T: KeccakItem<N>> KeccakState<N, T> {
         Self: Absorb<N>,
     {
         debug_assert!(N > 0 && len < RATE);
+
+        #[cfg(not(eurydice))]
         debug_assert!(last.iter().all(|buf| buf.len() == last[0].len()));
 
         self.load_last::<RATE, DELIM>(last, start, len);


libcrux-sha3/tests/sha3.rs
```diff
@@ -335,6 +429,131 @@ mod portable {
...
+        );
+    }
+
+    #[test]
+    fn sha3_shake128_squeeze_first_five() {
+        let mut state = incremental::init();
+
+        incremental::shake128_absorb_final(
+            &mut state,
+            test_vectors::EMPTY,
```
franziskuskiefer: All functions that operate in parallel require that all inputs have the same length. Looks like there are some debug asserts missing that would tell you (and some documentation). Can you add that?

parrot7483: Will add that. Should I then use 4 different inputs with same length for tests?

franziskuskiefer: Yes, different inputs is good, but they need the same length.

libcrux-sha3/src/generic_keccak.rs
```diff
@@ -205,6 +205,7 @@ impl<const N: usize, T: KeccakItem<N>> KeccakState<N, T> {
         Self: Absorb<N>,
     {
         debug_assert!(N > 0 && len < RATE);
+        debug_assert!(last.iter().all(|buf| buf.len() == last[0].len()));
```
franziskuskiefer, line 213: Thanks for adding. Do we have a nassert like this on all the parallel functions now? Adding a doc comment to the function would be nice as well.

libcrux-sha3/tests/sha3.rs
```diff
@@ -100,25 +277,113 @@ mod test_vectors {
...
+                  53c77776c5818bd08f5632cd03236595a0c1f9f2b8df52a8f7058e1d6dfa\
+                  80cc3d1bd4e59968f33dceb9e0d286cf0c83ede2a7e52d33a3b891e73b4a\
+                  b558ed01956b689d0e53b9c19f9d0a7f6d13c291cd9f34f21b9cb193c242\
+                  51224ee93a452e5728f44735af4293f43a62720771ab127f026ec08e74a0\
+                  7351335e5fff8e14e54c4cb027282bed13e";
     }
 }
 
 // Portable implementation tests
+// #[cfg(not(any(feature = "simd128", feature = "simd256")))]
```
franziskuskiefer: Remove?
```suggestion
```

libcrux-sha3/tests/sha3.rs
```diff
@@ -335,6 +617,207 @@ mod portable {
     }
 }
 
+#[cfg(feature = "simd256")]
+mod simd256 {
```
franziskuskiefer: This file is getting pretty long. Maybe move the platform modules into separate files.

libcrux-sha3/src/generic_keccak.rs
parrot7483: Yes. They all use `absorb_block` or `absorb_final` from generic_keccak.rs. I'll add the check to `absorb_block` as well.

franziskuskiefer (approved):
Thanks!

franziskuskiefer:
Looks like the tests are failing on CI 😬

parrot7483:
Makes sense. I can't run the arm tests.

karthikbhargavan:
Interesting, there appears to be a performance regression.

franziskuskiefer:
About the failing C CI. There's a real change in the extraction. But it shouldn't be there. I opened https://github.com/AeneasVerif/eurydice/issues/273 to track the issue. What we can do in the meantime is adding `#[cfg(not(eurydice))]` in front of the debug assert.

> Interesting, there appears to be a performance regression.

hugh, that may be fluke? The debug assert really shouldn't be compiled in release mode.

PR #1121, opened 2025-08-29
Update C Extraction Toolchain

### LINKED ISSUE #1117: Update Eurydice
The extraction in https://github.com/cryspen/libcrux/pull/1094 currently fails. The code making it fail in Eurydice has since been replaced. We need to update our C extraction toolchain to make this go away.

keks:
This PR updates Eurydice, Charon an Karamel.

We may take another look at the glue, and somewhere is an undefined constant. But it extracts an builds (on my machine, let's see what CI has to say about that...)

Closes #1117


libcrux-ml-kem/cg/eurydice_glue.h
```diff
@@ -112,6 +112,11 @@ typedef struct {
 #define Eurydice_slice_subslice2(s, start, end, t) \
   EURYDICE_SLICE((t *)s.ptr, (start), (end))
 
+// Previous version above does not work when t is an array type (as usual). Will
+// be deprecated soon.
+#define Eurydice_slice_subslice3(s, start, end, t_ptr)                         \
+  EURYDICE_SLICE((t_ptr)s.ptr, start, end)
```
franziskuskiefer: This is the bug I mentioned.

```suggestion
  EURYDICE_SLICE((t_ptr)s.ptr, (start), (end))
```

I opened https://github.com/AeneasVerif/eurydice/pull/268.

libcrux-ml-kem/cg/eurydice_glue.h
```diff
@@ -128,6 +133,9 @@ typedef struct {
 #define Eurydice_array_to_subslice2(x, start, end, t) \
   EURYDICE_SLICE((t *)x, (start), (end))
 
+// Same as above, variant for when start and end are statically known
+#define Eurydice_array_to_subslice3(x, start, end, t_ptr)                      \
+  EURYDICE_SLICE((t_ptr)x, start, end)
```
franziskuskiefer: This as well.

```suggestion
  EURYDICE_SLICE((t_ptr)x, (start), (end))
```

libcrux-ml-dsa/cg/libcrux_core.h
```diff
@@ -209,19 +208,200 @@ typedef struct Result_53_s {
...
 - SIZE= 3309
 */
 static inline libcrux_ml_dsa_types_MLDSASignature_8f
-libcrux_ml_dsa_types_zero_8f_fa(void) {
-  libcrux_ml_dsa_types_MLDSASignature_8f lit;
-  uint8_t repeat_expression[3309U] = {0U};
-  memcpy(lit.value, repeat_expression, (size_t)3309U * sizeof(uint8_t));
-  return lit;
+libcrux_ml_dsa_types_zero_c5_fa(void) {
+  return (libcrux_ml_dsa_types_MLDSASignature_8f{
```
franziskuskiefer: You should use karamel main. Then we don't have this ugly change.

libcrux-ml-kem/cg/libcrux_mlkem768_avx2.h
```diff
@@ -5150,12 +5151,11 @@ static KRML_MUSTINLINE void libcrux_ml_kem_ind_cpa_generate_keypair_unpacked_22(
...
     void *lvalue = (void *)0U;
     error_as_ntt[i] =
-        libcrux_ml_kem_ind_cpa_generate_keypair_unpacked_call_mut_4c_22(&lvalue,
+        libcrux_ml_kem_ind_cpa_generate_keypair_unpacked_call_mut_73_22(&lvalue,
                                                                         i);
   }
-  LowStar_Ignore_ignore(libcrux_ml_kem_ind_cpa_sample_vector_cbd_then_ntt_b4(
-                            error_as_ntt, prf_input, domain_separator),
-                        uint8_t, void *);
+  libcrux_ml_kem_ind_cpa_sample_vector_cbd_then_ntt_b4(error_as_ntt, prf_input,
```
franziskuskiefer, line 5157: The return value here should be ignored. I'm not sure why that's not being generated anymore. I opened https://github.com/AeneasVerif/eurydice/issues/269.

franziskuskiefer (commented):
A couple small things. But otherwise this looks good.

The warning (https://github.com/cryspen/hax/issues/1622) doesn't appear to actually impact the extraction.

libcrux-ml-dsa/cg/libcrux_core.h
keks: Quick note: If I use Karamel master (80f5435f2fc505973c469a4afcc8d875cddd0d8b), I get this error on `make test` in Eurydice:

```
sed -i 's/  KaRaMeL version: .*//' out/test-dst/**/*.{c,h} # This changes on every commit
sed -i 's/  KaRaMeL invocation: .*//' out/test-dst/**/*.{c,h} # This changes between local and CI
cd out/test-dst && cc -Wall -Werror -Wno-unused-variable  -I/home/keks/code/github.com/AeneasVerif/eurydice//../karamel/include -I. -I../../include  dst.c main.c && ./a.out
In file included from dst.h:11,
                 from dst.c:8:
dst.c: In function ‘dst_alloc3’:
../../include/eurydice_glue.h:81:29: error: missing braces around initializer [-Werror=missing-braces]
   81 | #define KRML_CLITERAL(type) (type)
      |                             ^
../../include/eurydice_glue.h:556:49: note: in definition of macro ‘Eurydice_box_new’
  556 |   ((t_dst)(malloc_and_init(sizeof(t), (char *)(&init))))
      |                                                 ^~~~
dst.c:28:44: note: in expansion of macro ‘KRML_CLITERAL’
   28 |         .ptr = (dst_T3 *)Eurydice_box_new((KRML_CLITERAL(dst_S_a4){ .foo = 0U, .my_data = { 0U } }),
      |                                            ^~~~~~~~~~~~~
../../include/eurydice_glue.h:81:29: error: missing braces around initializer [-Werror=missing-braces]
   81 | #define KRML_CLITERAL(type) (type)
      |                             ^
dst.c:27:7: note: in expansion of macro ‘KRML_CLITERAL’
   27 |       KRML_CLITERAL(Eurydice_dst_f1){
      |       ^~~~~~~~~~~~~
cc1: all warnings being treated as errors
make: *** [Makefile:85: test-dst] Error 1
rm out/test-closure_fn_cast/main.c out/test-const_generics/main.c out/test-array2d/main.c out/test-array/main.c out/test-dst/main.c
```

@franziskuskiefer suggested to ignore that for now and see whether extraction works anyway

franziskuskiefer: Looks like this is in `Eurydice_box_new`, which we don't use. So that should be fine. But probably worth opening an issue for in Eurydice.

libcrux-ml-kem/cg/libcrux_mlkem768_avx2.h
keks: I just noticed that the extraction prints "LowStar.Ignore.ignore doesn't go anywhere"

libcrux-ml-dsa/cg/libcrux_core.h
keks: https://github.com/AeneasVerif/eurydice/issues/271

keks:
Looks like the errors occur in the C extraction, but so far I only did the boring extraction :grimacing: Looking into it

libcrux-ml-kem/c.sh
keks, line 1: Turns out we don't really need this one for this particular PR, but it still seems like the correct thing to do

libcrux-ml-kem/c/headeronly/eurydice_glue.h
keks, line 1: I moved this to a subdirectory because we are calling the script with --clean, which removes the glue file and then either leaves us with eurydice's default glue file or none at all (if called with --no-glue). To keep the file around I moved it to a subdir. It is then copied into c/ (similar to cg/boring).

keks (commented):
Some notes

libcrux-ml-kem/c/headeronly/eurydice_glue.h
franziskuskiefer: The `--clean` is a little overloaded now in `c.sh`. We could also skip the hand written files there. But copying is fine for now as well.

libcrux-ml-kem/c.sh
franziskuskiefer: But now we can't copy the glue from upstream anymore?

.github/workflows/docker-c-build-test.yml
```diff
@@ -55,8 +55,8 @@ jobs:
       - name: Extract C
         run: |
           export HOME=/home/user
-          ./c.sh --no-glue
-
+          ./c.sh --config cg.yaml --mlkem768 --no-unrolling --out c --clean --no-glue
+          cp c/headeronly/eurydice_glue.h c/
```
franziskuskiefer: It would be nice if `c.sh` would do this as well such that we can locally run what's run on ci.

libcrux-ml-dsa/cg/libcrux_core.h
franziskuskiefer: Resolving this, the discussion continues on the issue.

libcrux-ml-kem/c/headeronly/eurydice_glue.h
franziskuskiefer: Oh, and now we have the file twice?

libcrux-ml-kem/c.sh
keks: Oh it can. If there is no glue argument, the default value of $glue is to copy from $EURYDICE_HOME.

.github/workflows/docker-c-build-test.yml
keks: Ah yeah, the idea of the changes to c.sh actually was to allow this, but I didn't actually use it. I changed the script to use the --glue option now

libcrux-ml-kem/c/headeronly/eurydice_glue.h
keks: We have it once as the source of truth and once as part of the generate stuff. We have the same with cg/eurydice_glue.h and cg/boring/eurydice_glue.h, but now I realized I didn't change the latter but it still works :thinking: not sure what's going on there.

libcrux-ml-kem/c.sh
franziskuskiefer: Ah, sorry yes. That was hidden in the diff view.

libcrux-ml-kem/c/headeronly/eurydice_glue.h
franziskuskiefer: Maybe we should just change the `--clean` to not clobber everything. Then we don't need these multiple copies and confusion if/what's working. So we can either add a `--clean-c`, or only delete `libcrux_*.{c,h}`.

franziskuskiefer (commented):
Ok, let's do this.

There's the one question about the glue. Otherwise this is good to go.

But we'll have to get #1123 in first I think?

libcrux-ml-kem/c/headeronly/eurydice_glue.h
keks: Ah right. Over there it's called with --no-glue and it doesn't do --clean, so the glue file in cg is neither deleted nor overwritten.

https://github.com/cryspen/libcrux/blob/b185a26684a0d164b7733432c5c0d88aefefce3e/libcrux-ml-kem/boring.sh#L28-L29

franziskuskiefer (approved):
Thanks, let's get this in when CI is happy.
That'll need the new docker image from #1123.

PR #1127, opened 2025-09-03
bump HAX version

Parrot7483:


karthikbhargavan (approved):

parrot7483:
Merge queue is failing because of eurydice conversion on Windows. What should be do here @karthikbhargavan ?

karthikbhargavan:
@franziskuskiefer @keks any advice on the windows issue. i think this was solved do maybe there is an easy fix?

franziskuskiefer:
@keks is looking into the issue. I think something in the upgrade yesterday wasn't quite right.

Issue #1129, opened 2025-09-04
[Meta] Speedup CI


franziskuskiefer:
CI is doing too much unnecessary work right now and slowing down merges unnecessarily. This issue tracks a couple immediate steps to improve this.

- [ ] Drop macos-13, or only run a simple build

There are some other issues that we may not want to tackle as part of this quick fix

- [x] https://github.com/hacspec/hax-actions/issues/6

franziskuskiefer:
hax CI setup is down to ~2 minutes now which speeds up hax jobs quite a bit.

jschneider-bensch:
We should revisit the merge queue setup and whether it makes sense to run some things nightly only, especially if #1133 is addressed.

jschneider-bensch:
Still relevant.

PR #1130, opened 2025-09-05
speedup ci a little

franziskuskiefer:
Towards #1129

### DIFF
--- a/.github/workflows/mlkem-build-test.yml
+++ b/.github/workflows/mlkem-build-test.yml
@@ -29,8 +29,6 @@ jobs:
         exclude:
           - bits: 32
             os: "macos-latest"
-          - bits: 32
-            os: "macos-13"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/mlkem-build-test.yml
```diff
@@ -23,7 +23,6 @@ jobs:
       matrix:
         bits: [32, 64]
         os:
-          - macos-13 # Intel mac
```
keks, line 26: There still is a macos-13 one in the exclude block below

keks (approved):
One nit, otherwise looks ready to merge!

.github/workflows/mlkem-build-test.yml
franziskuskiefer: Indeed. Removed

PR #1142, opened 2025-09-10
mldsa: use correct readme in Cargo.toml

keks:


franziskuskiefer:
@keks what's the state here?

Issue #1147, opened 2025-09-15
[C] mlkem: add 768 and 1024 C


franziskuskiefer:
Add `extraction/c` to mlkem with C extraction for 768 and 1024.

franziskuskiefer:
Fixed in #1168

PR #1150, opened 2025-09-15
Restructure C Code

### LINKED ISSUE #1049: [C] mlkem: move C extractions
Move `mlkem/cg` to `mlkem/extracts/cpp` (or maybe `cpp768`).

Should we open an issue to rename the cg folder to cpp, or would that just be unnecessary friction?

_Originally posted by @keks in https://github.com/cryspen/libcrux/pull/1046#pullrequestreview-2981985018_

### LINKED ISSUE #1148: Update `C-CODE.md` to reflect the current state
After moving the C extraction this file should be moved into `extracts` and update with the current structure.

keks:
- [x] move around extraction scripts and config
- [x] update CI
- [ ] check everything still works
- [ ] update C-CODE.md

Closes #1148.
Closes #1049.


libcrux-ml-kem/extracts/c-nss/c.sh
franziskuskiefer, line 1: The NSS version is outdated and I don't expect to still work. Let's drop them. #1147 would essentially replace that.

libcrux-ml-kem/extracts/c/c.sh
franziskuskiefer, line 1: I'm not sure if it's good to duplicate the `c.sh` script. So far this had all the generic things needed for extraction that we then specialise for example for `cpp` with `boring.sh`. We could structure this differently. But having multiple copies of `c.sh` doesn't sound great.

keks: Done.

keks:
> Some thoughts
> 
>     * Looking at this I'm not sure if `out` is the best name for the actual code. Maybe something like `generated`?

Darn, hadn't seen this earlier. I'll try to come up with something, might use generated.

>     * `c` and `cpp` are header only versions, whereas the new version from #1147 will generate C files. We could use the targets like `boring` and `nss` as names. No strong opinion in any direction. Just wondering what the best naming scheme is.

I have no real opinions on the names, but what you suggest sounds reasonable. Do you mean we should rename c -> nss, cpp -> boring?

I will make these changes tomorrow and squash them into this commit, so we have a clean base for the reviews afterwards.

franziskuskiefer:
yeah target names would be nss (new C from #1147), openssh (now c) and boring (now cg). The nss (or new C) may be more generally interesting. Others too. That's why specific names may not be that great either. We could also use more specific names cpp_header_only (cg), c_header_only (c), c (new C) would be something more generic, but also longer.

libcrux-ml-kem/extracts/cpp_header_only/generated/README.md
```diff
@@ -1,17 +1,17 @@
...
 The C code is generated from Rust using [Charon], [Eurydice] and [Karamel].
-The [c.sh](../c.sh) bash script drives the extraction, using the [cg.yaml](../cg.yaml)
+The [extract.sh](../c.sh) bash script drives the extraction, using the [cg.yaml](../cg.yaml)
 configuration file.
 While running the commands separately is possible, it is not recommended because
 the script sets all necessary configuration flags.
 
 ```bash
-./c.sh --config cg.yaml --out cg --mlkem768 --no-glue --no-unrolling
+./extract.sh --no-clean
```
jschneider-bensch, line 14: Why require users to pass `--no-clean` here, and not make it part of `extract.sh`?

keks: Yeah that is something that has been bugging me... The reason is that I just renamed the `boring.sh` script to `extract.sh`, and that often tool that as an argument. I had already planned to move this back to `boring.sh` and then have an `extract.sh` that just calls `boring.sh --no-clean` in the next PR in this area.

franziskuskiefer (commented):
The `generated` directory also contains the hand written code from eurydice and karamel. That may be a little misleading.

jschneider-bensch:
Looks like some paths in the CI workflows need to be updated still:
- ML-DSA extraction
- ML-KEM s390x build

keks:
> The `generated` directory also contains the hand written code from eurydice and karamel. That may be a little misleading.

Yes, I mostly just put whatever is in the current `c`/`cg` directories into the `generate` ones. We recently discussed how to handle the glue in a different PR and I had it somewhere else and then copied in, which we moved away from to avoid duplication. I suppose the alternative would be to properly split it. Not sure how much work that would be getting that to build. Should I look into it or should we leave it for now?



> Looks like some paths in the CI workflows need to be updated still:
> 
>     * ML-DSA extraction
> 
>     * ML-KEM s390x build

Ah, thanks for looking into it!

franziskuskiefer:
> Should I look into it or should we leave it for now?

no, let's get this in and do follow ups. I just noticed that we should probably update the structure a little there.

PR #1152, opened 2025-09-17
Fix warning. Unused Import.

Parrot7483:
Fixes warning when extracting libcrux-sha3

```rust
warning: unused import: `crate::traits::*`
 --> secrets/src/int/classify_public.rs:5:5
  |
5 | use crate::traits::*;
  |     ^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```


jschneider-bensch (approved):
Thanks!

Issue #1007, opened 2025-05-28
ML-DSA/ML-KEM - Update kats


franziskuskiefer:
Update kats

- [x] [ML-DSA wycheproof](https://github.com/C2SP/wycheproof/pull/146)
  - [x] `sign_seed`
  - [x] `sign_noseed`
  - [x] `verify`
- [x] [ML-KEM wycheproof](https://github.com/C2SP/wycheproof/blob/main/testvectors_v1/mlkem_test.json)
- [x] ML-DSA ACVP [keygen](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-keyGen-FIPS204), [sign](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigGen-FIPS204), [verify](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigVer-FIPS204)
- [ ] ML-KEM ACVP [keygen](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-KEM-keyGen-FIPS203), [encap/decap](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-KEM-encapDecap-FIPS203)

- [x] Move KATs to separate crate

franziskuskiefer:
Let's do this as part of the iot work.

PR #1153, opened 2025-09-17
`libcrux-kats` crate

wysiwys:
This pull request introduces a `libcrux-kats` crate for KAT test vectors that can be used in tests for other crates (e.g. `libcrux-ml-dsa`). Currently, this includes the wycheproof KAT test vectors for ML-DSA (sign `noseed` and verify).

These structs in `libcrux-kats` represent the test sets:
- `libcrux_kats::wycheproof::mldsa::MlDsaSignTests`
- `libcrux_kats::wycheproof::mldsa::MlDsaVerifyTests` 

Example usage:
```rust
let signing_tests = MlDsaSignTests::load(ParameterSet::MlDsa44);
for test_group in signing_tests.test_groups {
    for test in test_group.tests {
       // ...
   }
}
```

Also, the wycheproof test vectors used for ML-DSA are updated to the latest version, as of https://github.com/C2SP/wycheproof/tree/b51abcfb8dafa5316791e57cf48512a2147d9671. To go along with these changes, the `serde` schemas used for parsing are adjusted to match the updated schemas.

On the `libcrux-kats` crate, the `"mldsa"` feature flag can be used to enable the ML-DSA sign (`noseed`) and verify tests.

This pull request also updates the wycheproof tests in `libcrux-ml-dsa` to utilize the test vectors provided by the `libcrux-kats` crate. 

Addresses some tasks in #1007

### DIFF
--- a/crates/testing/kats/src/wycheproof/mldsa/sign_schema.rs
+++ b/crates/testing/kats/src/wycheproof/mldsa/sign_schema.rs
@@ -4,12 +4,15 @@ use serde::{Deserialize, Serialize};
 #[serde(rename_all = "camelCase")]
 /// Sign tests for ML-DSA (`noseed`)
 pub struct MlDsaSignTests {
+    /// the primitive tested in the test file
     pub algorithm: String,
 
+    /// additional documentation
     pub header: Vec<String>,
 
     pub notes: Notes,
 
+    /// the number of test vectors in this test
     pub number_of_tests: i64,
 
     pub schema: String,
@@ -52,6 +55,7 @@ pub struct TestGroup {
     #[serde(rename = "type")]
     pub test_group_type: Type,
 
+    /// Encoded ML-DSA private key
     #[serde(with = "hex::serde")]
     pub private_key: Vec<u8>,
 
@@ -69,22 +73,29 @@ pub enum Type {
 #[derive(PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Test {
+    /// Identifier of the test case
     pub tc_id: i64,
 
+    /// A brief description of the test case
     pub comment: String,
 
+    /// The message to sign
     #[serde(with = "hex::serde")]
     pub msg: Vec<u8>,
 
+    /// [optional] The additional context string (empty if not provided)
     #[serde(default)]
     #[serde(with = "hex::serde")]
     pub ctx: Vec<u8>,
 
+    /// The encoded signature (empty in case of expected failure)
     #[serde(with = "hex::serde")]
     pub sig: Vec<u8>,
 
+    /// Test result
     pub result: SignResult,
 
+    /// A list of flags
     pub flags: Vec<Flag>,
 }
 
--- a/crates/testing/kats/src/wycheproof/mldsa/verify_schema.rs
+++ b/crates/testing/kats/src/wycheproof/mldsa/verify_schema.rs
@@ -4,14 +4,17 @@ use serde::{Deserialize, Serialize};
 #[serde(rename_all = "camelCase")]
 /// Verify tests for ML-DSA
 pub struct MlDsaVerifyTests {
+    /// the primitive tested in the test file
     pub algorithm: String,
 
     pub generator_version: String,
 
+    /// additional documentation
     pub header: Vec<String>,
 
     pub notes: Notes,
 
+    /// the number of test vectors in this test
     pub number_of_tests: i64,
 
     pub schema: String,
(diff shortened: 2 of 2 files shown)


crates/testing/kats/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+repository.workspace = true
+
+[dependencies]
+hex = { version = "0.4.3", features = ["serde"] }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = { version = "1.0" }
+
+[features]
+mldsa = []
+wycheproof = []
```
franziskuskiefer: The algorithm features make sense but I'm not sure the wycheproof one does. When would you want to disable them?

crates/testing/kats/Readme.md
```diff
@@ -0,0 +1,11 @@
+# KATs
+
+This crate provides KAT test vectors for:
```
franziskuskiefer, line 3: This should mention somewhere that it's internal to libcrux and people should only use at their own risk.

Cargo.toml
```diff
@@ -32,7 +32,7 @@ members = [
     "secrets",
     "fstar-helpers/core-models",
     "test-utils",
-    "crates/primitives/aead",
+    "crates/primitives/aead", "crates/testing/kats",
```
franziskuskiefer: Format

crates/testing/kats/Cargo.toml
wysiwys: I think it depends on how the other libcrux crates will use the `libcrux-kats` crate. If each crate will always use all of the `libcrux-kats` KATs that are relevant to it (e.g. wycheproof, ACVP,.. for ML-DSA), then separate feature flags like `wycheproof` are not needed here.

wysiwys: Removed this feature flag, since there is not a need to disable it, (there isn't a scenario where only the `wycheproof` KATs would be needed and not the other KATs).

crates/testing/kats/Cargo.toml
jschneider-bensch, line 1: This crate should probably be excluded from publishing, right?

libcrux-ml-dsa/tests/wycheproof_sign.rs
```diff
@@ -1,43 +1,30 @@
 // This module tests the implementation against the Wycheproof signing
 // test vectors for the final version of the ML-DSA standard, added in
 // commit
+// TODO: add current commit
```
jschneider-bensch: This is still open, but I wonder if it's not enough to refer to the KATs crate for details on the exact origin of the KATs. That way it's only one place that has to be kept up to date.

Cargo.toml
```diff
@@ -33,6 +33,7 @@ members = [
     "fstar-helpers/core-models",
     "test-utils",
     "crates/primitives/aead",
+    "crates/testing/kats",
```
jschneider-bensch, line 36: If we want to stick to directory structure outlined in `CONTRIBUTING.md`, this crate should live in `crates/utils/kats`, I would say.

crates/testing/kats/src/wycheproof/mldsa/sign_schema.rs
jschneider-bensch, line 1: Are these generated from the JSON Schema still, or do we manually keep them up to date?
It doesn't have to be in this PR, but I thought it would be nice to have the `description` fields from the JSON Schema as doc comments here, if available.

jschneider-bensch (changes requested):
I think this looks good! Just have some small comments/nitpicks.

crates/testing/kats/src/wycheproof/mldsa/sign_schema.rs
wysiwys: These versions were actually manually updated, to make the smaller usability changes (like parsing hex values to `Vec<u8>` and renaming `Result` -> `SignResult`/`VerifyResult`). Since the schemas were manually updated, it's convenient to also manually set the doc comments to the `description` field, where available. I updated the doc comments for the Sign and Verify schemas - is this what you had in mind? There may be a way to generate the schemas automatically with these adjustments, but it doesn't look like it's possible with `quicktype`

wysiwys: It would be really nice to be able to automatically generate these in this format though - maybe there's a tool that allows that

jschneider-bensch: Thanks for adding the additional documentation! :+1:

Cargo.toml
wysiwys: Thanks for catching that! To address this, and based on an earlier discussion offline, PR #1164 added a `/testing` directory to the directory structure in `CONTRIBUTING.md`. I'll rebase before merging the PR

jschneider-bensch: That makes sense to me :+1:

wysiwys:
Thanks! I rebased this branch to `main`.

PR #1154, opened 2025-09-18
Blake2: implement `libcrux_traits::digest` traits for supported digest sizes only

wysiwys:
This PR implements the incremental digest traits (introduced in #1078) for only valid digest sizes, for each of the two algorithm structs `Blake2bHash<N>` and `Blake2sHash<N>`, to only allow valid hashers to be constructed.

One issue currently is that, when initializing the `Hasher` types that are based on these algorithms, the digest size needs to be checked at runtime for validity. The changes in this PR mean that, for these hasher types, code that initializes hashers with invalid digest sizes will not compile.

```rust
// This will not compile, because the provided digest sizes are not valid
// let hasher: Blake2bHasher<0> = todo!();
// let hasher: Blake2bHasher<65> = todo!();

// This will compile, because the digest size 32 is valid
let hasher: Blake2bHasher<32> = todo!();
```

These changes will also be helpful later on when implementing a `Hasher::new()` API that does not require a runtime check for digest size validity:
```rust
let hasher = Blake2bHasher::<32>::new();

// This will not compile, because the digest size 0 is invalid
let hasher = Blake2bHasher::<0>::new();
```

### DIFF
--- a/blake2/src/impl_digest_trait.rs
+++ b/blake2/src/impl_digest_trait.rs
@@ -1,9 +1,134 @@
 use crate::impl_hacl::*;
 use libcrux_traits::digest::{arrayref, slice, DigestIncrementalBase, Hasher, UpdateError};
 
+// Implements sets of allowed digest sizes
+mod digest_size_support {
+    pub(super) struct Blake2bOutSupport;
+    pub(super) struct Blake2sOutSupport;
+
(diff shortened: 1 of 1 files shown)


keks (approved):
Looks good! 👍

PR #1155, opened 2025-09-18
Add API for initializing digest state

wysiwys:
This PR adds more convenient APIs for initializing the digest state associated with an implementor `D` of `DigestIncrementalBase`. A `D::IncrementalState` can now be initialized by calling the `new()` method.

A new API is also added to `libcrux_traits::digest::Hasher`, that builds on this API:
```rust
/// Initialize a hasher.
pub fn new() -> Self {
    Self {
        state: D::IncrementalState::new(),
    }
}
``` 
Using `Hasher::new()` to initialize hashers:
```rust
// type aliases for `libcrux_traits::digest::Hasher<_,_>`
let mut hasher = libcrux_sha2::Sha256Hasher::new();
let mut hasher = libcrux_blake2::Blake2sHasher::<32>::new();
```
Additionally, the tests in the `blake2` and `sha2` crates are updated to include tests for this API.

This PR is based on the branch in:
- https://github.com/cryspen/libcrux/pull/1154

This PR addresses #1156

### DIFF
--- a/blake2/src/impl_digest_trait.rs
+++ b/blake2/src/impl_digest_trait.rs
@@ -1,5 +1,7 @@
 use crate::impl_hacl::*;
-use libcrux_traits::digest::{arrayref, slice, DigestIncrementalBase, Hasher, UpdateError};
+use libcrux_traits::digest::{
+    arrayref, slice, DigestIncrementalBase, Hasher, InitializeDigestState, UpdateError,
+};
 
 // Implements sets of allowed digest sizes
 mod digest_size_support {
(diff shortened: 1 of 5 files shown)


blake2/src/impl_digest_trait.rs
```diff
@@ -122,7 +124,19 @@ mod digest_size_support {
...
+    ($out_size:ident, $type:ty, $blake2:ty, $hasher:ty, $set:ident, $builder:ty) => {
+        impl<const $out_size: usize> InitializeDigestState for $blake2
+        where
+            // implement for supported digest lengths only
+            $set: SupportsLen<$out_size>,
+        {
+            fn new() -> Self {
+                <$builder>::new_unkeyed()
+                    .build_const_digest_len()
+                    .expect("trait was implemented for invalid digest size")
```
jschneider-bensch: Can this ever happen, if the implementation is bounded with `SupportsLen<OUT_SIZE>`?
Maybe this is a place where `unwrap_unchecked()` (with a safety comment) would be justified.

jschneider-bensch (approved):
This looks good I think!

blake2/src/impl_digest_trait.rs
wysiwys: Since the implementation is bounded with `SupportsLen<OUT_SIZE>`, it's not something that can happen. 

One other way to avoid needing to `expect()` here, while not introducing unsafe code, could be to introduce an internal function like `build_const_digest_len_internal()`, which skips the length check (and maybe `build_const_digest_len` could call it internally)

wysiwys: I added an example implementation of the approach described above to this branch. Do you think this approach works well here, or does it introduce too much complexity? I can see advantages to either solution, though
```rust
fn new() -> Self {
    <$builder>::new_unkeyed()
        // NOTE: this version of `build_const_digest_len()` does not check
        // that the digest length is valid. This check is not required
        // since this trait is only implemented for supported lengths.
        .build_const_digest_len_internal()
        .into()
}
```

wysiwys: Actually, I restored this to the way it was for now. I think the question of whether to `unwrap()`/`unwrap_unchecked()`/`build_internal()` here could be something to take another look at later, though. For example, the `blake2` builders themselves could be adapted to only compile if valid lengths are provided, to avoid returning a `Result` from the builder.

jschneider-bensch: Fully agree! Like you said, we can come back to this later, if we see a need for it :+1:

Issue #395, opened 2024-07-14
Verify libcrux-sha3


karthikbhargavan:
Verify the libcrux-sha3 crate and show that libcrux-ml-kem and libcrux-ml-dsa use it correctly.

franziskuskiefer:
@Parrot7483 let's use this as the meta issue for the overall SHA-3 verification project.

franziskuskiefer:
First steps: #1092

PR #1157, opened 2025-09-18
Add full type check for SHA3 portable

Parrot7483:
This PR adds full type checking for the SHA3 portable version for https://github.com/cryspen/libcrux/issues/395

Extract using `hax.py extract --portable`, prove using `hax.py prove`.

Besides prove code the following things where changed:
- reorder functions `libcrux-sha3/src/generic_keccak/xof.rs`
- The `absorb_full` function in `libcrux-sha3/src/generic_keccak/xof.rs` previously did not always consume the internal buffer when it was full. This state could only be reached by misusing the API. The only bug I found.
- split up the rho and pi function for verification purpose

This version contains a few things I would consider "hacks"
- The extraction for the squeeze does not work. We add the expected F* code using `hax_lib::fstar::replace`
- `impl From<u32> for Algorithm` does not work because of the panic.
-  In `libcrux-sha3/src/generic_keccak/xof.rs` the `#[hax_lib::fstar::options("...")]` has no effect. I use an `hax_lib::fstar::before `instead


crates/algorithms/sha3/src/generic_keccak.rs
parrot7483, line 1: It might be a good idea to switch back to a `theta_rho` function (how [it was before](https://github.com/cryspen/libcrux/blob/33fd4f1562f93a7d3a89c7fb3d7177f1442bc0db/libcrux-sha3/src/generic_keccak.rs#L50) lax check), splitting it up the same way.

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -22,7 +63,19 @@ pub(crate) struct KeccakXofState<
     sponge: bool,
 }
 
+#[cfg(hax)]
+pub(crate) fn keccak_xof_state_inv<
```
parrot7483: PARALLEL_LANES is not needed here anymore

.github/workflows/sha3-hax.yml
```diff
@@ -103,6 +103,28 @@ jobs:
...
+        with:
+          hax_reference: ${{ github.event.inputs.hax_ref || needs.get-hax-ref.outputs.hax_ref }}
+          fstar: v2025.02.17
+
+      - uses: actions/download-artifact@v5
+        with:
+          name: fstar-extractions-portable
+          path: .
+
+      - name: 🏃 Full type check SHA3 crate (portable)
```
franziskuskiefer, line 100: What's the performance here. THe proofs are usually slower and we only run them on nightly. But if it's fast we can keep it here for now.

libcrux-sha3/src/generic_keccak/portable.rs
```diff
@@ -41,11 +70,34 @@ impl KeccakState<1, u64> {
...
+  if n = 0 then ()
+  else if k = n - 1 then ()
+  else mul_succ_le k (n - 1) d
+"#
+)]
+#[hax_lib::requires(
+    RATE != 0 &&
+    RATE <= 200 &&
+    RATE % 8 == 0 /* &&
+    data.len() + RATE <= usize::MAX */
```
franziskuskiefer: Remove the commented code?

libcrux-sha3/src/generic_keccak/portable.rs
```diff
@@ -41,11 +70,34 @@ impl KeccakState<1, u64> {
...
+#[hax_lib::ensures(|_|
+    future(out).len() == out.len())
+]
+#[hax_lib::fstar::options("--split_queries always --query_stats --z3rlimit 300")]
 #[inline]
 pub(crate) fn keccak1<const RATE: usize, const DELIM: u8>(data: &[u8], out: &mut [u8]) {
     let mut s = KeccakState::<1, u64>::new();
     let data_len = data.len();
     for i in 0..data_len / RATE {
+        hax_lib::fstar!("mul_succ_le (v i) (v (data_len /! v_RATE)) (v v_RATE)");
```
franziskuskiefer: Let's add some space between code and proof to make it easier to read.

```suggestion
        hax_lib::fstar!("mul_succ_le (v i) (v (data_len /! v_RATE)) (v v_RATE)");

```

libcrux-sha3/src/generic_keccak/portable.rs
```diff
@@ -60,6 +112,8 @@ pub(crate) fn keccak1<const RATE: usize, const DELIM: u8>(data: &[u8], out: &mut
     } else {
         s.squeeze::<RATE>(out, 0, RATE);
         for i in 1..blocks {
+            hax_lib::loop_invariant!(|_: usize| out.len() == outlen);
+            hax_lib::fstar!("mul_succ_le (v i) (v blocks) (v v_RATE)");
```
franziskuskiefer:
```suggestion
            hax_lib::fstar!("mul_succ_le (v i) (v blocks) (v v_RATE)");

```

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -1,10 +1,51 @@
+use hax_lib::int::*;
+
 use crate::{
     generic_keccak::KeccakState,
-    traits::{Absorb, KeccakItem, Squeeze1},
+    traits::{Absorb, KeccakItem, Squeeze},
 };
 
+// Helper lemma for F* verification in the `absorb_full` loop.
```
franziskuskiefer: Let's move the proof bits into a separate module.

crates/algorithms/sha3/src/generic_keccak/xof.rs
franziskuskiefer: So it should be removed?

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -107,67 +208,100 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
+    pub(crate) fn absorb(&mut self, inputs: &[&[u8]; PARALLEL_LANES])
+    where
+        KeccakState<PARALLEL_LANES, STATE>: Absorb<PARALLEL_LANES>,
+    {
+        let input_remainder_len = self.absorb_full(inputs);
+
+        // ... buffer the rest if there's not enough input (left).
+        if input_remainder_len > 0 {
+            debug_assert!(
+                // self.buf_len == 0 || // We consumed everything (or it was empty all along).
```
franziskuskiefer: What about this commented bit? Is this not true anymore?

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -201,16 +341,33 @@ impl<const RATE: usize, STATE: KeccakItem<1>> KeccakXofState<1, RATE, STATE> {
...
+                // For a and b with b < a
+                // (a / b) * b + a % b = a
+
                 let remaining = out_len % RATE;
                 if remaining > 0 {
                     // Squeeze out the last partial block
                     self.inner.keccakf1600();
+
+                    hax_lib::fstar!("lemma_div_mul_mod out_len v_RATE");
+                    hax_lib::fstar!("assert (v blocks * v v_RATE + v remaining == v out_len)");
```
franziskuskiefer:
```suggestion
                    hax_lib::fstar!("assert (v blocks * v v_RATE + v remaining == v out_len)");

```

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -107,67 +208,100 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
+        let input_remainder_len = self.absorb_full(inputs);
+
+        // ... buffer the rest if there's not enough input (left).
+        if input_remainder_len > 0 {
+            debug_assert!(
+                // self.buf_len == 0 || // We consumed everything (or it was empty all along).
+                self.buf_len + input_remainder_len <= RATE
+            );
+
+            hax_lib::assert!(input_remainder_len.to_int() + self.buf_len.to_int() <= RATE.to_int());
```
franziskuskiefer: This requires the unconditional hax_lib dependency for debug builds. Is there a better way to do this?

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -107,67 +208,100 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
         for i in 0..num_blocks {
-            // We only get in here if `input_len / RATE > 0`.
-            self.inner
-                .load_block::<RATE>(inputs, input_consumed + i * RATE);
+            hax_lib::loop_invariant!(|_: usize| self.buf_len == self_buf_len);
+
+            hax_lib::fstar!("lemma_offset_plus_rate_le_total (v input_consumed) (v i) (v num_blocks) (v v_RATE)");
+            let start = i * RATE + input_consumed;
+
+            // TODO: The cfg should not be needed here
```
franziskuskiefer: `end` is only defined in `cfg(hax)`. Hence this is needed because it gets made into a regular Rust debug.

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -107,67 +208,100 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
             self.buf_len = 0;
         }
 
-        // We only need to consume the rest of the input.
-        let input_to_consume = inputs[0].len() - input_consumed;
+        #[cfg(hax)]
```
franziskuskiefer, line 192: That's a lot of `cfg(hax)` here. You could throw the code into a block `{...}` to make it a little cleaner.

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -41,47 +103,81 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
 
-            let input_len = inputs[0].len();
+            #[cfg(hax)]
+            let self_buf_len = self.buf_len; // ghost variable for F* proof
 
             #[allow(clippy::needless_range_loop)]
             for i in 0..PARALLEL_LANES {
-                self.buf[i][self.buf_len..self.buf_len + input_remainder_len]
-                    .copy_from_slice(&inputs[i][input_len - input_remainder_len..]);
+                hax_lib::loop_invariant!(|_: usize| { self.buf_len == self_buf_len });
```
franziskuskiefer, line 136:
```suggestion
                hax_lib::loop_invariant!(|_: usize| { self.buf_len == self_buf_len });

```

franziskuskiefer (commented):
Sorry for being a little slow on this.
Here a couple first comments. Can you address them and update? Then we can do another round.

crates/algorithms/sha3/src/generic_keccak/xof.rs
parrot7483: The variables are used below and would not be in scope if in a block

libcrux-sha3/src/generic_keccak/xof.rs
parrot7483: I don't think I understand this correctly. Would putting a `cfg(hax)` be enough? It is only needed for proof.

crates/algorithms/sha3/src/generic_keccak/xof.rs
parrot7483: It is needed. I was wrong.

libcrux-sha3/src/generic_keccak/xof.rs
parrot7483: still true

.github/workflows/sha3-hax.yml
parrot7483: Do you mean not running the check on merge? On my modern AMD laptop the check takes 90seconds.

parrot7483:
@franziskuskiefer I applied your feedback and also cherry-picked one commit from my AVX/NEON branch adding the correct precondition on the traits and doing some light refactoring (mainly having consistent names for variables).

.github/workflows/sha3-hax.yml
franziskuskiefer: Yes. But sounds like it's not doing much yet, we can keep it here.

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -32,6 +62,13 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
     }
 
     /// Generate a new keccak xof state.
+    #[hax_lib::requires(
+        PARALLEL_LANES == 1 && // TODO: Generalize for the parallel case
```
franziskuskiefer: What's up with this todo?

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -41,47 +78,83 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
+    /// content to consume, and `0` otherwise.
+    /// If `consumed > 0` is returned, `self.buf` contains a full block to be
+    /// loaded.
+    // Note: consciously not inlining this function to avoid using too much stack
+    // #[hax_lib::fstar::options("--fuel 5")]
+    #[hax_lib::requires(
+        PARALLEL_LANES == 1 &&
+        keccak_xof_state_inv(self)
+    )]
+    #[hax_lib::ensures(|consumed|
```
franziskuskiefer, line 110: Can you move larger things like these out into functions as well?

crates/algorithms/sha3/src/lib.rs
```diff
@@ -15,8 +15,15 @@ mod impl_digest_trait;
 #[cfg(not(any(hax, eurydice)))]
 pub use impl_digest_trait::*;
```
franziskuskiefer, line 17: This file has merge conflicts.

libcrux-sha3/src/simd/portable.rs
```diff
@@ -77,25 +100,41 @@ pub(crate) fn load_last<const RATE: usize, const DELIMITER: u8>(
...
+    #[cfg(hax)]
+    let out_len = out.len(); // ghost variable
+
     for i in 0..octets {
-        out[start + 8 * i..start + 8 * i + 8]
-            .copy_from_slice(&get_ij(s, i / 5, i % 5).to_le_bytes());
+        hax_lib::loop_invariant!(|i: usize| out.len() == out_len);
+
+        let value = get_ij(s, i / 5, i % 5);
+        let bytes = value.to_le_bytes();
```
franziskuskiefer: Are these changes needed for the proof? If yes, they'd need some explanation to why. If not, we should keep it as is.
Same for the similar change below.

crates/algorithms/sha3/src/generic_keccak.rs
```diff
@@ -97,32 +100,43 @@ impl<const N: usize, T: KeccakItem<N>> KeccakState<N, T> {
     }
 
     #[inline(always)]
-    #[hax_lib::fstar::replace_body("assert true")]
-    fn rho(&mut self, t: [T; 5]) {
+    fn rho_0(&mut self, t: [T; 5]) {
```
franziskuskiefer, line 109: If this setup is required for the proofs to work, please add some comments for that.

libcrux-sha3/src/lib.rs
```diff
@@ -46,16 +53,17 @@ pub enum Algorithm {
     Sha512 = 4,
 }
 
-#[hax_lib::attributes]
+// TODO: Verification fails because of the panic
```
franziskuskiefer: Todo

libcrux-sha3/src/lib.rs
```diff
@@ -46,16 +53,17 @@ pub enum Algorithm {
...
+#[cfg(not(any(hax, eurydice)))]
 impl From<u32> for Algorithm {
-    #[hax_lib::requires(v <= 4)]
+    #[hax_lib::requires(v == 1 || v == 2 || v == 3 || v == 4)]
```
franziskuskiefer: If it's ignored with the config, we shouldn't need this?

crates/algorithms/sha3/src/traits.rs
```diff
@@ -52,11 +56,45 @@ pub(crate) trait KeccakItem<const N: usize>: Clone + Copy {
 }
 
 /// Trait to load new bytes into the state.
+#[hax_lib::attributes]
 pub(crate) trait Absorb<const N: usize> {
     /// Absorb a block
+    #[hax_lib::requires(
+      from_bool(
```
franziskuskiefer, line 67: Please move large pre and post conditions out into separate functions.

crates/algorithms/sha3/src/traits.rs
```diff
@@ -72,19 +110,52 @@ pub(crate) trait Absorb<const N: usize> {
...
+        { pred ==>
+          (Core.Slice.impl__len #u8 out_future <: usize) =. (Core.Slice.impl__len #u8 out <: usize)
+        };
   f_squeeze:v_RATE: usize -> x0: v_Self -> x1: t_Slice u8 -> x2: usize -> x3: usize
     -> Prims.Pure (t_Slice u8)
         (f_squeeze_pre v_RATE x0 x1 x2 x3)
         (fun result -> f_squeeze_post v_RATE x0 x1 x2 x3 result)
 }
+
+// TODO: See above
```
franziskuskiefer, line 143: Todo

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -1,10 +1,24 @@
...
     generic_keccak::KeccakState,
-    traits::{Absorb, KeccakItem, Squeeze1},
+    traits::{Absorb, KeccakItem, Squeeze},
 };
 
+#[cfg(hax)]
+#[hax_lib::fstar::replace("open Libcrux_sha3.Lemmas")]
+const _: () = ();
+
+// TODO: Should not be needed. Use hax_lib::fstar::options("")
```
franziskuskiefer: Todo

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -22,7 +36,23 @@ pub(crate) struct KeccakXofState<
     sponge: bool,
 }
 
+#[cfg(hax)]
```
franziskuskiefer: Please move proof code into separate modules.

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -22,7 +36,23 @@ pub(crate) struct KeccakXofState<
...
+) -> bool {
+    RATE != 0
+        && RATE <= 200
+        && RATE % 8 == 0
+        && (RATE % 32 == 8 || RATE % 32 == 16)
+        && xof.buf_len <= RATE
+}
+
 #[hax_lib::attributes]
+#[hax_lib::fstar::options("--split_queries always --z3rlimit 300")] // TODO: Has no effect
```
franziskuskiefer: Todo

libcrux-sha3/src/generic_keccak/xof.rs
franziskuskiefer: Todo

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -41,47 +78,83 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
-    where
-        KeccakState<PARALLEL_LANES, STATE>: Absorb<PARALLEL_LANES>,
-    {
-        let input_remainder_len = self.absorb_full(inputs);
+    /// Returns the `consumed` bytes from `inputs` if there's enough buffered
+    /// content to consume, and `0` otherwise.
+    /// If `consumed > 0` is returned, `self.buf` contains a full block to be
+    /// loaded.
+    // Note: consciously not inlining this function to avoid using too much stack
+    // #[hax_lib::fstar::options("--fuel 5")]
```
franziskuskiefer: snip

crates/algorithms/sha3/src/generic_keccak/xof.rs
franziskuskiefer: Sure, you just need to get them in scope. Something like this would do the trick
```rust
let (self_buf_len, end) = {
    let end = consumed + num_blocks * RATE;
    hax_lib::assert!(end <= inputs[0].len());
     (self.buf_len, end)
};
```

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -108,66 +179,101 @@ impl<const PARALLEL_LANES: usize, const RATE: usize, STATE: KeccakItem<PARALLEL_
...
         for i in 0..num_blocks {
-            // We only get in here if `input_len / RATE > 0`.
-            self.inner
-                .load_block::<RATE>(inputs, input_consumed + i * RATE);
+            hax_lib::loop_invariant!(|_: usize| self.buf_len == self_buf_len);
+            hax_lib::fstar!("mul_succ_le (v i) (v num_blocks) (v v_RATE)");
+
+            let start = i * RATE + consumed;
+
+            #[cfg(hax)]
```
franziskuskiefer, line 207: Is this `cfg` needed?

crates/algorithms/sha3/src/traits.rs
```diff
@@ -72,19 +110,52 @@ pub(crate) trait Absorb<const N: usize> {
...
+// [@@ FStar.Tactics.Typeclasses.tcinstance]
+// let _ = fun (v_Self:Type0) (v_T:Type0) {|i: t_Squeeze v_Self v_T|} -> i._super_18390613159176269294
 "
 )]
-pub(crate) trait Squeeze1<T: KeccakItem<1>> {
+#[hax_lib::attributes]
+pub(crate) trait Squeeze<T: KeccakItem<1>> {
```
franziskuskiefer, line 149: Why did this lose the 1?

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -201,16 +313,34 @@ impl<const RATE: usize, STATE: KeccakItem<1>> KeccakXofState<1, RATE, STATE> {
...
+                #[cfg(hax)]
+                let self_buf_len = self.buf_len;
+
                 for i in 0..blocks {
+                    hax_lib::loop_invariant!(
+                        |_: usize| out.len() == out_len && self_buf_len == self.buf_len
+                    );
+                    hax_lib::fstar!("assert (v i * v v_RATE <= v out_len)");
+
+                    #[cfg(hax)]
```
franziskuskiefer: Is this `cfg` needed?

libcrux-sha3/src/generic_keccak/xof.rs
```diff
@@ -201,16 +313,34 @@ impl<const RATE: usize, STATE: KeccakItem<1>> KeccakXofState<1, RATE, STATE> {
                 // How many blocks do we need to squeeze out?
```
franziskuskiefer: This is pretty deep now. Using an early return for `out_len <= 0` would make things look a little nicer.

franziskuskiefer (changes requested):
Thanks, looks pretty good. A couple things and some cleanup, then this can go in.

libcrux-sha3/src/simd/portable.rs
parrot7483: Yes, if I inline everything F* can't figure out that the bounds are correct every time. Sometimes it works sometimes it doesn't. 

I think inlining `let bytes = get_ij(s, octets / 5, octets % 5).to_le_bytes()` should be possible.

crates/algorithms/sha3/src/traits.rs
parrot7483: I split this up into a function for RATE and same length. RATE precondition is used all over the place.

crates/algorithms/sha3/src/traits.rs
parrot7483: I will link to the issue here. I failed to reproduce this bug in the HAX playground several time tho.

crates/algorithms/sha3/src/traits.rs
parrot7483: I want to emphasize that having the Squeeze trait multiple times is because of HAX not supporting &[&mut [u8]; N] or other types with mutables inside. If that would work that would simplify a lot of things in this crate. I will write a comment explaining it.

libcrux-sha3/src/generic_keccak/xof.rs
parrot7483: Known bug: https://github.com/cryspen/hax/issues/1698
Added comment.

libcrux-sha3/src/generic_keccak/xof.rs
parrot7483: https://github.com/cryspen/libcrux/pull/1157/files/3413619f172e8a9cf24d35ca81f49127dd2d9749#r2523397042

libcrux-sha3/src/generic_keccak/xof.rs
parrot7483: This todo is done in the next SIMD PR. Removing the comment.

crates/algorithms/sha3/src/generic_keccak/xof.rs
parrot7483: I would keep it like this here because it is not used anywhere else and specific to the behavior of this function. I would not even know how to name it.

crates/algorithms/sha3/src/generic_keccak/xof.rs
parrot7483: Yes because `end` is also a ghost variable.

libcrux-sha3/src/generic_keccak/xof.rs
parrot7483: no

crates/algorithms/sha3/src/generic_keccak/xof.rs
parrot7483: The invariant must be in the same module otherwise it can not read the private fields.

crates/algorithms/sha3/src/simd/portable.rs
```diff
@@ -35,9 +49,13 @@ fn _veorq_n_u64(a: u64, c: u64) -> u64 {
 }
 
 #[inline(always)]
+#[hax_lib::requires(
+    valid_rate(RATE) &&
+    start.to_int() + RATE.to_int() <= blocks.len().to_int()
+)]
 pub(crate) fn load_block<const RATE: usize>(state: &mut [u64; 25], blocks: &[u8], start: usize) {
-    #[cfg(not(eurydice))]
```
parrot7483, line 39: Do not remove eurydice config

parrot7483:
@franziskuskiefer I applied your review. cherry-picked https://github.com/cryspen/libcrux/pull/1157/commits/1064d0d2423630c0482d5d76039b38fe99214415 from SIMD branch. Merged main.

parrot7483:
I have a tendency to pressing these big green buttons on github messing everything up. sorry.

parrot7483:
What do you want to do about these warnings?

```rust
warning: unused attribute
   --> crates/algorithms/sha3/src/lib.rs:181:1
    |
181 | #[inline(always)]
    | ^^^^^^^^^^^^^^^^^ help: remove this attribute
    |
note: attribute also specified here
   --> crates/algorithms/sha3/src/lib.rs:176:1
    |
176 | #[cfg_attr(not(eurydice), inline(always))]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!`
```

parrot7483:
F* proofs failing with "Namespace 'Core' cannot be found" after core-models directory moved

Reproducer:
- cd libcrux-ml-kem && ./hax.py extract && ./hax.py prove
- cd crates/algorithms/sha3 && ./hax.py extract && ./hax.py prove

Both fail at prove step with same error. ML-KEM and SHA3 both have this issue.

Using the old name "Core_models" still works for single models.

parrot7483:
The auto-discovery in `Makefile.generic` doesn't find the Core F* library from the hax proof-libs, so I had to set `HAX_HOME` explicitly, which doesn't work in CI.

crates/algorithms/sha3/src/traits.rs
```diff
@@ -65,26 +93,66 @@ pub(crate) trait Absorb<const N: usize> {
...
 #[hax_lib::fstar::replace(
-    interface,
-    "
-class t_Squeeze1 (v_Self: Type0) (v_T: Type0) = {
-  f_squeeze_pre:v_RATE: usize -> v_Self -> t_Slice u8 -> usize -> usize -> Type0;
-  f_squeeze_post:v_RATE: usize -> v_Self -> t_Slice u8 -> usize -> usize -> t_Slice u8 -> Type0;
+    interface, "
+class t_Squeeze (v_Self: Type0) (v_T: Type0) = {
+  // TODO: This super variable is problematic and makes typecheck fail
+  // https://github.com/cryspen/hax/issues/1554
```
franziskuskiefer, line 111: Looks like that issue is closed and if it works without the commented code, let's drop it.

crates/algorithms/sha3/src/traits.rs
```diff
@@ -65,26 +93,66 @@ pub(crate) trait Absorb<const N: usize> {
...
+          (Core.Slice.impl__len #u8 out_future <: usize) =. (Core.Slice.impl__len #u8 out <: usize)
+        };
   f_squeeze:v_RATE: usize -> x0: v_Self -> x1: t_Slice u8 -> x2: usize -> x3: usize
     -> Prims.Pure (t_Slice u8)
         (f_squeeze_pre v_RATE x0 x1 x2 x3)
         (fun result -> f_squeeze_post v_RATE x0 x1 x2 x3 result)
 }
+
+// TODO: See above
+// [@@ FStar.Tactics.Typeclasses.tcinstance]
```
franziskuskiefer, line 144: Same here, let's remove when it's not used.

franziskuskiefer:
> What do you want to do about these warnings?

that should be fixed in a separate PR

crates/algorithms/sha3/src/traits.rs
parrot7483: still a problem. I reopened the issue

crates/algorithms/sha3/src/traits.rs
parrot7483: same

crates/algorithms/sha3/hax.py
```diff
@@ -94,10 +94,24 @@ def __call__(self, parser, args, values, option_string=None) -> None:
             env=hax_env,
         )
 
+        # Extract Core models
```
jschneider-bensch: From the output of `./hax.py extract --portable` it looks like nothing is extracted here actually, and you also removed the `core-models` proof directory from `FSTAR_INCLUDE_DIRS_EXTRA` in the Makefile. So is this extraction step needed here, after all?

crates/algorithms/sha3/src/generic_keccak/portable.rs
```diff
@@ -1,18 +1,41 @@
 use super::*;
 
+#[cfg(hax)]
+use crate::proof_utils::valid_rate;
+
+#[cfg(hax)]
+#[hax_lib::fstar::replace("open Libcrux_sha3.Proof_utils.Lemmas")]
```
jschneider-bensch: Is it possible to avoid this manual F* dependency by making the lemmas in `proof_utils::lemmas` into Rust functions (nops which are replaced as they are now) and calling these from Rust instead of invoking them in inline `fstar!` blocks?

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -1,8 +1,25 @@
+use hax_lib::int::*;
```
jschneider-bensch, line 2: This should be behind a `#[cfg(hax)]`, no?

crates/algorithms/sha3/src/generic_keccak/xof.rs
```diff
@@ -1,8 +1,25 @@
...
-    traits::{Absorb, KeccakItem, Squeeze1},
+    traits::{Absorb, KeccakItem, Squeeze},
 };
 
+#[cfg(hax)]
+use crate::proof_utils::valid_rate;
+
+#[cfg(hax)]
+#[hax_lib::fstar::replace("open Libcrux_sha3.Proof_utils.Lemmas")]
+const _: () = ();
```
jschneider-bensch: Same question about avoiding the manual F* `open` as in `generic_keccak::portable`.

crates/algorithms/sha3/src/generic_keccak/xof.rs
jschneider-bensch: Could you move it to `proof_utils` and pass in `xof.buf_len` from here instead of passing in `xof` and accessing it in the invariant body?

parrot7483:
@jschneider-bensch Does `buf_len` need to remain private, or is pub(crate) acceptable? I made it `pub(crate)` so `portable.rs` can access it when calling `keccak_xof_state_inv` in the trait implementations.

crates/algorithms/sha3/hax.py
parrot7483: I am not sure. I am confused about the difference between Core and Core_models

parrot7483:
@jschneider-bensch friendly ping

jschneider-bensch:
> @jschneider-bensch Does `buf_len` need to remain private, or is pub(crate) acceptable? I made it `pub(crate)` so `portable.rs` can access it when calling `keccak_xof_state_inv` in the trait implementations.

Sorry for the late reply!
I think having it `pub(crate)` is okay.

parrot7483:
Can we merge then?

jschneider-bensch:
> Can we merge then?

There was an issue with typechecking on hax 0.3.6 since Core models produced a name clash. We had this same issue in ML-DSA and solved it by patching the clashing names by hand (cf. https://github.com/cryspen/libcrux/pull/1293).
I've replaced `hax.py` here by `hax.sh` lifted from the ML-DSA F* star extraction and adapted to SHA3, which contains the patching logic. With this I can extract and typecheck the portable code here locally, now.
Please let me know if the changes I made are okay for you.

parrot7483:
Extraction and typecheck work on my machine too.

PR #1164, opened 2025-09-22
`CONTRIBUTING.md`: Add `/crates/testing` to directory structure

wysiwys:
This PR adds an entry for `/crates/testing` to the directory structure outlined in `CONTRIBUTING.md`, for containing testing-related crates. It also moves the entry for the `libcrux-test-utils` into `/crates/testing`.

Another option would be to just include all the testing crates in the `/crates/utils` folder.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,18 +13,18 @@ While many old crates do not yet follow the new structure, please use it for new
 crates:
 
 ```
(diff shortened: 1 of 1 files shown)


keks:
Huh, I think the changes look good, but CI failed in strange ways. I made them rerun, and at least for the ECDH one it seems to have worked. Maybe the lax one will also work.

PR #1165, opened 2025-09-22
PQCP APIs

jschneider-bensch:
This PR implements the PQCP common API for ML-KEM, making them available under the `pqcp` feature.

- The parts of the API that work on byte arrays (i.e. those that don't have the `_struct` suffix) are implemented in `mlkem$variant::pqcp`, using the `kem::arrayref::Kem` trait implementation.
- The `_struct` parts of the API are implemented in `mlkem$variant::$platform::unpacked::pqcp`.
- I have tried to stick as close as possible to the C-API given in [mlkem-native](https://github.com/pq-code-package/mlkem-native/blob/7875c832cac79a3c736a7d65d8ccaec86297ec49/mlkem/kem.c), meaning I took the same names for functions and arguments and kept the same argument order.

Some things may need discussion:
- The PR is smaller in scope, compared to #1012: This introduces only the PQCP API, without making further changes. In particular, I think we should provide an idiomatic (key-centric?) API in a follow-up, instead of doing it all at the same time.
- Implications of sticking to the `mlkem-native` API:
  -  The packed API reads and writes byte arrays, and only internally uses our own packed types (e.g. `MlKemPublicKey`, etc). This means from the perspective of the PQCP API we have two levels of key representation: raw byte arrays, and the unpacked key types. Another option would be to change from byte arrays to our own packed types.
  - The API takes `&mut` parameters instead of returning values, only signalling errors using the return type.
- We do not, at the moment, expose `MlKemPrivateKeyUnpacked` types, so I opted to let the `MlKemKeyPairUnpacked` structs stand in for the private keys in `_struct` functions. This means we have some structural differences in key generation where we write to a mutable `MlKemKeyPairUnpacked` instead of separate `MlKem(Private|Public)KeyUnpacked`s. It also means `crypto_kem_sk_from_seed` is just an alias to `crypto_kem_keypair_derand_struct`.
- Public/private key validation: `mlkem-native` performs key validation in `crypto_kem_enc_derand` and `crypto_kem_dec`. We do not, at the moment. We only perform validation in `crypto_kem_parse_(pk|sk)`, when deserializing an unpacked key.


libcrux-ml-kem/src/lib.rs
```diff
@@ -175,6 +191,257 @@ cfg_kyber! {
     }
 }
 
+macro_rules! pqcp_api {
```
franziskuskiefer: This is enough code that it would benefit from its own module.

libcrux-ml-kem/src/lib.rs
```diff
@@ -175,6 +191,257 @@ cfg_kyber! {
...
+                <$trait_implementer>::encaps(ct, ss, pk, &coins)
+                    .map_err(|_| PQCPError::Encapsulation)
+            }
+
+            #[cfg(feature = "rand")]
+            #[doc = "Encapsulate ML-KEM "]
+            #[doc = $variant]
+            #[doc = " shared secret (randomness generated internally)
+
+Does not perform public key validation."]
```
franziskuskiefer: Maybe put this in a regular `///` to make the code here for all the empty line/linebreaks nicer to read?

libcrux-ml-kem/src/lib.rs
franziskuskiefer: That module could also get a doc test on top with an example to show how to use this one.

jschneider-bensch: I've pulled the error and macros out into their own module and documented the macros better.
I've also managed to generate a doc test for the "packed" API that can serve as a usage example. For the unpacked API, I would like to also include a usage example/doc test, but I'm still having difficulty making the docs look nice, since I run into [this issue](https://stackoverflow.com/questions/60905060/prevent-line-break-in-doc-test).

franziskuskiefer:
Sorry, I didn't comment on the overall thing.
I think this is the right call. We can add the nice APIs then separately.

> * Public/private key validation: `mlkem-native` performs key validation in `crypto_kem_enc_derand` and `crypto_kem_dec`. We do not, at the moment. We only perform validation in `crypto_kem_parse_(pk|sk)`, when deserializing an unpacked key.

FIPS 203 is pretty clear that validation shouldn't happen with every invocation of enc/dec. So I think we shouldn't do it for the regular ones. We could do it on other APIs. But I think doing it when parsing is the right call.

libcrux-ml-kem/src/lib.rs
franziskuskiefer: Ah I see. Let's get it in as is

Issue #1166, opened 2025-09-22
[AES-GCM] Cleanup to prepare crate for release


wysiwys:
Cleanup to prepare the crate for a release:
- [ ] Move crate to new directory in `/crates/algorithms/`
- [ ] Add rustdocs
- [ ] Add Readme
- [ ] Remove `rand` from default features

PR #1167, opened 2025-09-23
AES-GCM Initial Review

jschneider-bensch:
Doing an initial review here, before we get started on cleanup (cf. #1159, #1166).

### DIFF
--- a/aesgcm/src/lib.rs
+++ b/aesgcm/src/lib.rs
@@ -13,6 +13,9 @@ mod aes_gcm;
 mod aes_gcm_128;
 mod aes_gcm_256;
 
+use libcrux_traits::aead::{arrayref, consts, slice, typed_owned};
+
+// TODO: should this trait be re-exported here?
 pub use libcrux_traits::aead::arrayref::Aead;
 
 /// Trait for an AES State.
@@ -35,40 +38,48 @@ pub(crate) trait State {
 pub struct DecryptError();
 
 /// AES-GCM 128.
-pub struct AesGcm128 {}
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct AesGcm128;
 
 /// Portable AES-GCM 128.
-pub struct PortableAesGcm128 {}
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct PortableAesGcm128;
 
 /// Neon AES-GCM 128.
 #[cfg(feature = "simd128")]
-pub struct NeonAesGcm128 {}
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct NeonAesGcm128;
 #[cfg(not(feature = "simd128"))]
 pub type NeonAesGcm128 = PortableAesGcm128;
 
 /// AES-NI AES-GCM 128.
 #[cfg(feature = "simd256")]
-pub struct X64AesGcm128 {}
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct X64AesGcm128;
 #[cfg(not(feature = "simd256"))]
 pub type X64AesGcm128 = PortableAesGcm128;
 
 /// AES-GCM 256.
-pub struct AesGcm256 {}
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct AesGcm256;
 
 /// Portable AES-GCM 256.
-pub struct PortableAesGcm256 {}
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct PortableAesGcm256;
 
 /// Neon AES-GCM 256.
 #[cfg(feature = "simd128")]
-pub struct NeonAesGcm256 {}
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct NeonAesGcm256;
 
 /// Neon AES-GCM 256.
 #[cfg(not(feature = "simd128"))]
 pub type NeonAesGcm256 = PortableAesGcm256;
 
 /// AES-NI AES-GCM 256.
+#[derive(Clone, Copy, PartialEq, Eq)]
 #[cfg(feature = "simd256")]
-pub struct X64AesGcm256 {}
+pub struct X64AesGcm256;
 
 /// AES-NI AES-GCM 256.
 #[cfg(not(feature = "simd256"))]
@@ -250,6 +261,23 @@ pub mod x64 {
     x64_pub_mod!(r"AES-GCM 256 ", aes_gcm_256, crate::aes_gcm_256::State<platform::x64::State, platform::x64::FieldElement>);
 }
 
+/// Macro to implement the libcrux_traits public API traits
+///
+/// For the blanket impl of `typed_refs::Aead` to take place,
+/// the `$type` must implement `Copy` and `PartialEq`.
+macro_rules! impl_traits_public_api {
+    ($type:ty, $keylen:expr, $taglen:expr, $noncelen:expr ) => {
+        // prerequisite for typed_owned::Aead
+        impl consts::AeadConsts for $type {
+            const KEY_LEN: usize = KEY_LEN;
+            const TAG_LEN: usize = TAG_LEN;
+            const NONCE_LEN: usize = NONCE_LEN;
+        }
+        // implement typed_owned::Aead
+        typed_owned::impl_aead_typed_owned!($type, KEY_LEN, TAG_LEN, NONCE_LEN);
+    };
+}
+
 /// Macro to implement the different structs and multiplexing.
 macro_rules! api {
     ($mod_name:ident, $variant:ident, $multiplexing:ty, $portable:ident, $neon:ident, $x64:ident) => {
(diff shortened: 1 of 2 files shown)


aesgcm/build.rs
```diff
@@ -0,0 +1,34 @@
...
+    // Force a simd build. Make sure you know what you're doing.
+    let enable_simd128 = read_env("LIBCRUX_ENABLE_SIMD128");
+    let enable_simd256 = read_env("LIBCRUX_ENABLE_SIMD256");
+
+    let simd128_possible = target_arch == "aarch64";
+    if (simd128_possible || enable_simd128) && !disable_simd128 {
+        // We enable simd128 on all aarch64 builds.
+        println!("cargo:rustc-cfg=feature=\"simd128\"");
+    }
+    let simd126_possible = target_arch == "x86_64";
```
jschneider-bensch:
```suggestion
    let simd256_possible = target_arch == "x86_64";
    if (simd256_possible || enable_simd256) && !disable_simd256 {
```
This is a typo that we have in several `build.rs` files in the workspace, but we don't need to copy it over here.

aesgcm/Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "libcrux_aesgcm"
+version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+edition.workspace = true
+repository.workspace = true
+readme = "README.md"
```
jschneider-bensch, line 9: For cleanup: There's no `README.md` yet.

Cargo.toml
```diff
@@ -81,23 +96,27 @@ bench = false                               # so libtest doesn't eat the argumen
 libcrux-platform = { version = "=0.0.2", path = "sys/platform" }
 
 [dependencies]
-libcrux-traits = { version = "=0.0.3", path = "traits" }
+libcrux-hacl-rs.workspace = true
 libcrux-chacha20poly1305 = { version = "=0.0.3", path = "chacha20poly1305" }
```
jschneider-bensch: Bit unrelated, but why not make this a workspace dependency as well?

aesgcm/src/platform/intel_ni/aes_core.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    #[inline]
+    fn xor_key(&mut self, key: &Self) {
+        xor_key1_state(self, key);
+    }
+
+    #[inline]
+    fn aes_enc(&mut self, key: &Self) {
+        aes_enc(self, key);
+        (self, key);
```
jschneider-bensch:
```suggestion
```
Unnecessary line here?

aesgcm/src/gf128.rs
```diff
@@ -0,0 +1,75 @@
...
+#[cfg(test)]
+mod test;
+
+/// Generic Gf128 state.
+pub(crate) struct GF128State<T: GF128FieldElement> {
+    accumulator: T,
+    r: T,
+}
+
+const KEY_LEN: usize = 16;
```
jschneider-bensch: Why is a separate constant necessary here? This will always be `AES_BLOCK_LEN` right?

aesgcm/fuzz/fuzz_targets/encrypt128.rs
jschneider-bensch, line 1: Perhaps it would be good to also fuzz with different input data that exceeds the block size.

aesgcm/src/platform/x64/gf128_core.rs
```diff
@@ -0,0 +1,184 @@
...
+    mm_clmulepi64_si128, mm_slli_si128, mm_srli_si128, mm_unpackhi_epi64, mm_unpacklo_epi64,
+    mm_xor_si128,
+};
+
+// XXX: A lot of the code below is shared with NEON. Refactor!
+
+/// An avx2 gf128 field element.
+#[derive(Clone, Copy)]
+#[repr(transparent)]
+pub(crate) struct FieldElement(pub(super) u128);
```
jschneider-bensch, line 13: Is the potential refactoring the reason that this is not just a `__m128i` like in `intel_ni::aes_core`?

aesgcm/src/platform/neon/aes_core.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn store_block(&self, out: &mut [u8]) {
+        debug_assert!(out.len() == 16);
+        _vst1q_u8(out, *self);
+    }
+
+    #[inline]
+    #[allow(unsafe_code)]
+    fn xor_block(&self, input: &[u8], out: &mut [u8]) {
+        debug_assert!(input.len() == out.len() && input.len() <= 16);
+        // XXX: hot-fix to have enough input and output here.
```
jschneider-bensch, line 88: NB: I did not investigate this further.

aesgcm/src/aes.rs
```diff
@@ -0,0 +1,29 @@
+//! The AES block cipher function.
+
+use crate::platform::*;
+
+pub(crate) type ExtendedKey<T, const NUM_KEYS: usize> = [T; NUM_KEYS];
+
+// This is inlined into the key expansion below.
```
jschneider-bensch: This comment seems superflous here.

aesgcm/src/ctr.rs
```diff
@@ -0,0 +1,103 @@
...
+
+        block_cipher(&mut st, &self.extended_key);
+
+        st.xor_block(input, out);
+    }
+
+    #[inline]
+    fn aes_ctr_xor_blocks(&self, ctr: u32, input: &[u8], out: &mut [u8]) {
+        debug_assert!(input.len() == out.len() && input.len().is_multiple_of(AES_BLOCK_LEN));
+        debug_assert!(input.len() / AES_BLOCK_LEN < u32::MAX as usize);
```
jschneider-bensch: Should this bound be one less? If `input.len() / AES_BLOCK_LEN == u32::MAX - 1` and we start with `ctr == 2` then we'll wrap to 0 below and we'll repeat the initial key block, or did I get something wrong?

aesgcm/src/platform/neon/aes_core.rs
jschneider-bensch, line 1: Don't the AES instructions require `[target_feature(enable = "aes")]` as well as the `neon` target feature?

I think it works because in `lib.rs` we also guard this implementation as a whole with `#[cfg(all(target_arch = "aarch64", target_feature = "aes"))]`.

aesgcm/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+}
+
+#[cfg(feature = "simd128")]
+pub mod neon {
+    pub_mod!(r"AES-GCM 128 ", aes_gcm_128, crate::aes_gcm_128::State<platform::neon::State, platform::neon::FieldElement>);
+    pub_mod!(r"AES-GCM 256 ", aes_gcm_256, crate::aes_gcm_256::State<platform::neon::State, platform::neon::FieldElement>);
+}
+
+#[cfg(feature = "simd256")]
+pub mod x64 {
```
jschneider-bensch, line 181: FWIW, I like this style more than the NEON version, where every individual function in `platform::neon` has to be marked unsafe if it uses an intrinsic.

aesgcm/src/platform/x64.rs
jschneider-bensch, line 1: As discussed, this should probably be named `x64`.

libcrux-intrinsics/src/arm64.rs
```diff
@@ -431,3 +463,36 @@ pub fn _vbcaxq_u64(a: uint64x2_t, b: uint64x2_t, c: uint64x2_t) -> uint64x2_t {
...
+
+#[inline]
+#[target_feature(enable = "neon")]
+pub fn _veorq_u8(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t {
+    veorq_u8(a, b)
+}
+
+#[inline]
+pub fn _vaesmcq_u8(data: uint8x16_t) -> uint8x16_t {
+    unsafe { vaesmcq_u8(data) }
```
jschneider-bensch, line 475: Is the `[target_feature(enable = "aes")]` style not supported?

aesgcm/Cargo.toml
franziskuskiefer: I created a dummy one just to make this valid.

aesgcm/fuzz/fuzz_targets/encrypt128.rs
franziskuskiefer: You mean for the values here that are expected to have a certain length? Like key, nonce, tag, ctxt? Or do you mean fuzzing the internal functions that process blocks?

aesgcm/src/ctr.rs
franziskuskiefer: No that sounds right. I add this as a comment and update the bound. But really we should never call this function this way as this assert won't be there in release builds. So we should check that we check the length correctly before.

aesgcm/src/gf128.rs
franziskuskiefer: Yeah, I guess I added it because it's called key here, not block. But we should at least map it to the block size.

aesgcm/src/lib.rs
franziskuskiefer: We don't really need the feature on Arm, there are different reasons here. But I'll revert the other style anyway.

On x64/x86 we have to tell the compiler about the CPU features like AES. That's why we _need_ the annotations here. In the case of arm we can add hardware features to functions to avoid making certain intrinsic calls unsafe. But they are not needed for the compiler.

However, hax is still on an old rustc version that can't do the feature flags to avoid unsafe. And this would also bump our general MSRV quite a bit. And I didn't really like the ergonomics yet. I think we need to play with that a little more.

But because of all of that I'll change back to making all the arm intrinsics calls unsafe in the intrinsics crate and drop the feature flags.

libcrux-intrinsics/src/arm64.rs
franziskuskiefer: No, not all intrinsics support that yet. But see the other comment. I'm reverting to this style for now again anyway.

aesgcm/src/platform/neon/aes_core.rs
franziskuskiefer: See other comment as well. We don't need to enable any features on arm.
I forgot to update the include here to the feature style that we usually use. Both is fine, it's just a little more consistent this way and allows cross compilation.

aesgcm/src/platform/x64/gf128_core.rs
franziskuskiefer: I don't think there's a good reason for this to be a `u128` right now. We transmute this anyway. But yeah, if we want to unify this with the other version, we could use the same type here.

aesgcm/fuzz/fuzz_targets/encrypt128.rs
jschneider-bensch: I think, yes, what I meant was to have AAD and plaintext lengths that exercise the block splitting internally.

jschneider-bensch:
Closing this in favor of #1185

PR #1173, opened 2025-09-24
Include ML-KEM wycheproof tests in `libcrux-kats` crate and use in `libcrux-ml-kem` tests

wysiwys:
This pull request adds the ML-KEM wycheproof KATs to the `libcrux-kats` crate. It also adds tests using these KATs to the `libcrux-ml-kem` crate.

For better organization in the `libcrux-kats` crate, this PR also moves the shared structs (described in the original `common.json` [schema](https://github.com/C2SP/wycheproof/blob/b319f03a388837d9678501a9fadaf7a395c9519f/schemas/common.json)) into a separate module. For structs representing flags and notes for test vectors and test groups,  the enums generated automatically by `quicktype` are also replaced with structs that more closely correspond to the string types defined in the schema.

Addresses part of #1007


crates/testing/kats/src/wycheproof/schema_common.rs
```diff
@@ -0,0 +1,33 @@
+//! Structs based on schemas/common.json
```
franziskuskiefer: Let's add a link here to that file.

libcrux-ml-dsa/tests/wycheproof_verify.rs
```diff
@@ -25,7 +25,7 @@ macro_rules! wycheproof_verify_test {
                     assert_eq!(test_group.tests.len(), 1);
                     assert!(test_group.tests[0]
                         .flags
-                        .contains(&Flag::IncorrectPublicKeyLength));
+                        .contains(&"IncorrectPublicKeyLength".into()));
```
franziskuskiefer: Hm, what we had here before looked nicer. What's the reason for this change? Now we have to know the exact string here.

libcrux-ml-kem/tests/wycheproof.rs
```diff
@@ -0,0 +1,109 @@
...
+                        // compare shared secret
+                        assert_eq!(shared_secret_from_decapsulate, test.shared_secret.as_ref());
+
+                        // assert result is valid
+                        assert_eq!(test.result, MlKemResult::Valid);
+                    }
+                }
+            }
+            #[test]
+            fn wycheproof_encaps() {
```
franziskuskiefer:
```suggestion
            fn encaps() {
```

libcrux-ml-kem/tests/wycheproof.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use libcrux_ml_kem::{MlKemCiphertext, MlKemPublicKey};
+
+macro_rules! wycheproof_test {
+    ($name:ident, $parameter_set:expr, $module:path) => {
+        mod $name {
+            use super::*;
+
+            #[test]
+            fn wycheproof_keygen_and_decaps() {
```
franziskuskiefer: Since the test is called `wycheproof` already the prefixes seem unnecessary.
```suggestion
            fn keygen_and_decaps() {
```

libcrux-ml-kem/tests/wycheproof.rs
```diff
@@ -0,0 +1,109 @@
...
+                }
+            }
+            #[test]
+            fn wycheproof_encaps() {
+                use $module::*;
+                let katfile = MlKemTests::load($parameter_set);
+
+                for test_group in katfile.encaps_tests() {
+                    for test in &test_group.tests {
+                        // all tests have an Invalid results and a `ModulusOverflow` flag
```
franziskuskiefer, line 51: Oh only the negative tests are in here. Interesting. Then we should definitely keep the others around.

franziskuskiefer (changes requested):
lgtm with a few small things

libcrux-ml-dsa/tests/wycheproof_verify.rs
wysiwys: The reason for replacing the flags with strings is to improve maintainability (the flags are represented by strings rather than enums in the [schema](https://github.com/C2SP/wycheproof/blob/cd136e97040de0842c3a198670b1c5e4f423c940/schemas/mlkem_test_schema.json#L114), and the `Flag` enum variants generated by `quicktype` only contain the variants that were found in this version of the ML-KEM test data. If new test vectors are added later, they might introduce new flags that are not included in the enum).

However, the enum representation is definitely cleaner and more user-friendly. Would it be fine to keep `Flag` as an enum, and then update the enum if new flag types are added?

franziskuskiefer: Yeah I think keeping the enum and just updating when we get new variants sounds better to me.

PR #1175, opened 2025-09-24
AES-GCM: Implement remaining `libcrux_traits::aead` traits

### LINKED ISSUE #1159: AES-GCM API

wysiwys:
This PR implements the remaining `libcrux_traits::aead` traits for the AES-GCM algorithm structs in this crate:
- `slice::Aead` (internal)
- `typed_owned::Aead` (public, key-centric)
- `typed_refs::Aead` (public, key-centric, implemented by a blanket impl based on `typed_owned::Aead`)

It also adds basic tests for the key-centric functionality added by the `typed_owned::Aead` and `typed_refs::Aead` traits.

Resolves #1159


aesgcm/src/lib.rs
```diff
@@ -13,6 +13,9 @@ mod aes_gcm;
 mod aes_gcm_128;
 mod aes_gcm_256;
 
+use libcrux_traits::aead::{arrayref, consts, slice, typed_owned};
+
+// TODO: should this trait be re-exported here?
```
franziskuskiefer, line 18: Yes, we should always re-export the relevant crates so that users don't have to find the right crate with the trait.

aesgcm/src/lib.rs
```diff
@@ -35,40 +38,48 @@ pub(crate) trait State {
 pub struct DecryptError();
 
 /// AES-GCM 128.
-pub struct AesGcm128 {}
+#[derive(Clone, Copy, PartialEq)]
```
franziskuskiefer: If `PartialEq` is there, it should also have `Eq`.

aesgcm/src/lib.rs
```diff
@@ -406,3 +462,57 @@ api!(
     NeonAesGcm256,
     X64AesGcm256
 );
+
+#[cfg(test)]
+mod tests {
```
franziskuskiefer: This should be an integration or doc test I guess?

franziskuskiefer (changes requested):
Generally lgtm with a few nits.

Do you want to drop the other public APIs in a separate PR is that supposed to be part of this one?

jschneider-bensch:
Ah, looks like there's a problem when the slice implementation macro defines a test of the same name for the 128-bit and 256-bit variants.

aesgcm/src/lib.rs
wysiwys: Ah, definitely, these would make more sense as integration tests (and also doctests) :+1: And this probably applies to [these tests in the chachapoly crate](https://github.com/cryspen/libcrux/blob/bfdf7f9550b58c61dd6cd0b9649c4f689b328008/chacha20poly1305/src/impl_aead_trait.rs#L158-L202), too

wysiwys:
@jschneider-bensch  it looks like this happened because the slice implementation macro was invoked multiple times in the same namespace. I moved the trait implementations for each algorithm struct to a separate module That change appears to have fixed the issue!

aesgcm/tests/key_centric.rs
```diff
@@ -0,0 +1,48 @@
...
+    let mut ct = [0; 43];
+    let mut pt_out = [0; 43];
+
+    k.encrypt(&mut ct, &mut tag, &nonce, b"", pt).unwrap();
+    k.decrypt(&mut pt_out, &nonce, b"", &ct, &tag).unwrap();
+    assert_eq!(pt, &pt_out);
+}
+
+#[test]
+fn test_key_centric_refs() {
```
jschneider-bensch, line 29: I think this test could be a good usage example for the crate, noting also that `AesGcm128` is the one to use if you want runtime multiplexing of the fastest implementation for a given platform.

jschneider-bensch (approved):
To me this looks good for getting the traits implemented.

I think what can be done on this crate still, before release, is:
- dropping the initial public API
- Improve documentation:
  - Module-level docs, README.md
  - Item-level docs giving information on when you should use which `*AesGcm128` struct

But those could also be done in separate PRs.

franziskuskiefer (commented):
Thanks

franziskuskiefer (approved):

aesgcm/tests/key_centric.rs
wysiwys: Sounds good! I added this and the points above to the cleanup issue #1166

PR #1177, opened 2025-09-24
Improve HKDF API

keks:
Instead of creating a kdf crate, I am iterating on the HKDF API.

- [x] docs & examples
- [x] use secret types
- [x] move hkdf into crates/

### DIFF
--- a/crates/algorithms/hkdf/src/hkdf.rs
+++ b/crates/algorithms/hkdf/src/hkdf.rs
@@ -189,10 +189,10 @@ impl Algorithm {
 
 /// Generates HKDF implementation modules for specific hash algorithms.
 ///
-/// This macro creates a complete HKDF implementation module for a specific SHA algorithm,
+/// This macro creates a complete HKDF implementation module for a specific SHA2 algorithm,
 /// including both typed API methods on the [`Hkdf`] struct and standalone module functions.
 /// It generates implementations for extract, expand, and full HKDF operations with both
-/// fixed-size and variable-size array variants.
+/// fixed-size array references and variable-size slice variants.
 ///
 /// Parameters:
 ///
@@ -214,7 +214,7 @@ macro_rules! impl_hkdf {
         /// specifically for the underlying hash algorithm. It includes both standalone
         /// functions and methods on the typed [`Hkdf`] struct.
         ///
-        /// The `_fixed` variants work with compile-time known PRK sizes for better type safety,
+        /// The `_arrayref` variants work with compile-time known PRK sizes for better type safety,
         /// while the regular variants accept slices and perform runtime validation.
         pub mod $name {
             use libcrux_secrets::U8;
@@ -229,12 +229,12 @@ macro_rules! impl_hkdf {
                 /// Returns [`ExtractError::ArgumentTooLong`] if one of `ikm` or `salt` is longer than
                 /// [`u32::MAX`] bytes.
                 #[inline(always)]
-                pub fn extract_fixed(
+                pub fn extract_arrayref(
                     prk: &mut [U8; $hash_len],
                     salt: &[U8],
                     ikm: &[U8],
-                ) -> Result<(), FixedExtractError> {
-                    extract_fixed(prk, salt, ikm)
+                ) -> Result<(), ArrayReferenceExtractError> {
+                    extract_arrayref(prk, salt, ikm)
                 }
 
                 /// HKDF extract using the `salt` and the input key material `ikm`.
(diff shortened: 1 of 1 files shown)


crates/primitives/kdf/Readme.md
```diff
@@ -0,0 +1,8 @@
+# KDF
+
+This crate provides an interface to the key derivation functions (KDFs) implemented in libcrux.
+
+We currently support:
+
+* Extract-then-Expand: HKDF
+* PRF: HMAC, Blake2
```
franziskuskiefer: Not sure if people should really use hmac as a prf. But there are of course some historical protocols that do.

franziskuskiefer (commented):
You didn't want to put the APIs just into the respective crates? What APIs should consumers use when they only want to use blake2 or hkdf in this case?

keks:
I dropped the attempts to standardise KDF and PRF and instead tried making the hkdf API a bit neater.

crates/algorithms/hkdf/src/hkdf.rs
```diff
@@ -0,0 +1,458 @@
...
+/// Returns [`ExpandError::PrkTooShort`] if `prk` is shorter than hash length.
+pub fn expand(algo: Algorithm, okm: &mut [U8], prk: &[U8], info: &[u8]) -> Result<(), ExpandError> {
+    match algo {
+        Algorithm::Sha256 => sha2_256::expand(okm, prk, info),
+        Algorithm::Sha384 => sha2_384::expand(okm, prk, info),
+        Algorithm::Sha512 => sha2_512::expand(okm, prk, info),
+    }
+}
+
+pub fn hkdf(
```
jschneider-bensch, line 117: This could get a copy of the doc comment from below.

crates/algorithms/hkdf/src/hkdf.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+pub struct Sha2_256;
+pub struct Sha2_384;
+pub struct Sha2_512;
+
+pub struct Hkdf<Algo>(PhantomData<Algo>);
+
+impl Algorithm {
+    /// Returns the length of the underlying hash function.
```
jschneider-bensch:
```suggestion
    /// Returns the digest length of the underlying hash function.
```

crates/algorithms/hkdf/src/hkdf.rs
```diff
@@ -0,0 +1,458 @@
...
+    pub const fn hash_len(self) -> usize {
+        match self {
+            Algorithm::Sha256 => 32,
+            Algorithm::Sha384 => 48,
+            Algorithm::Sha512 => 64,
+        }
+    }
+}
+
+macro_rules! impl_hkdf {
```
jschneider-bensch, line 209: Can you add documentation on the expected values for the macro parameters, or what they're used for?

crates/algorithms/hkdf/src/hkdf.rs
```diff
@@ -0,0 +1,458 @@
...
+
+            impl Hkdf<$struct_name> {
+                /// HKDF extract using the `salt` and the input key material `ikm`.
+                /// The result is written to `prk`.
+                ///
+                /// Returns nothing on success.
+                /// Returns [`ExtractError::ArgumentTooLong`] if one of `ikm` or `salt` is longer than
+                /// [`u32::MAX`] bytes.
+                #[inline(always)]
+                pub fn extract_fixed(
```
jschneider-bensch: Would it make sense to call this `_arrayref` instead of `_fixed`? That would be more in line with the naming of other traits.

jschneider-bensch (dismissed):
I think it looks good, beside the issue with the doc test.
Thanks for simplifying the PSQ code while you were updating it to the new API!

crates/algorithms/hkdf/src/hkdf.rs
```diff
@@ -0,0 +1,458 @@
...
+//!
+//! ### Using the dynamic API
+//!
+//! ```
+//! use libcrux_hkdf::{extract, expand, Algorithm};
+//! use libcrux_secrets::{U8, Classify, DeclassifyRef};
+//!
+//! // Input key material and salt
+//! let ikm = &[U8::classify(0x0b); 22];
+//! let salt = b"salt".declassify_ref();
```
franziskuskiefer: Should we show all the (de)classifies here? If we do, it may be nicer when doing it always on the array, rather than the individual `u8`s. Just for readability.

crates/algorithms/hkdf/src/hkdf.rs
franziskuskiefer: We should add a lint to force public functions to have comments.

crates/algorithms/hkdf/src/hkdf.rs
keks: I was thinking of doing that initially but didn't like the name :sweat_smile: But yeah maybe consistency is more important.

crates/algorithms/hkdf/src/hkdf.rs
keks: We have to, otherwise the doc tests fail if we have the check-secret-independence feature set. And I agree that putting the classify at the end looks much nice - I thought I had run into issues with that earlier but it seems to work here

crates/algorithms/hkdf/src/hkdf.rs
keks: I'll open a PR that adds the check to CI, but it will fail. The secrets trait is missing a lot of docs, not sure about other stuff.

keks: https://github.com/cryspen/libcrux/pull/1193

jschneider-bensch (approved):
Looks good!

Issue #1103, opened 2025-08-19
Make all traits use secret types


keks:
If the check-secret-independence feature is not enabled, this behaves exactly the same as the regular types, so the caller can just throw in normal buffers. But it's also possible for callers to do the checks on their code.

PR #1184, opened 2025-09-29
Make aead use secrets and add keygen functions

### LINKED ISSUE #1102: Add keygen function to Aead traits
While all the AEADs we have now just use random bytes, this is not true for all AEADs. Also, we are going to need it later for the key-centric APIs.

keks:
A step in the direction of #1103  and addresses #1102.

Closes #1102

### DIFF
--- a/chacha20poly1305/src/lib.rs
+++ b/chacha20poly1305/src/lib.rs
@@ -5,7 +5,7 @@
 //!
 //! ```rust
 //! # fn main(){
-//! # use libcrux_secrets::{Classify, ClassifyRef, Declassify, U8};
+//! # use libcrux_secrets::{Classify, ClassifyRef, Declassify, DeclassifyRef, U8};
 //! # let key_bytes = [0u8.classify(); 32];
 //! # const MSG_LEN: usize = 19;
 //! #
--- a/.github/workflows/aead-build_test.yml
+++ b/.github/workflows/aead-build_test.yml
@@ -87,6 +87,12 @@ jobs:
 
       # Build ...
 
+      - name: 🧐 Check Secret Independene
+        run: |
+          rustc --print=cfg
+          cargo check --verbose -F check-secret-independence -p libcrux-aead
+          cargo check --verbose -F check-secret-independence -p libcrux-chacha20poly1305
+
       - name: 🔨 Build
         run: |
           rustc --print=cfg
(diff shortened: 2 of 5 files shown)


chacha20poly1305/src/impl_aead_trait.rs
```diff
@@ -32,11 +35,11 @@ mod impl_chachapoly {
     impl Aead<KEY_LEN, TAG_LEN, NONCE_LEN> for ChaCha20Poly1305 {
         fn encrypt(
             ciphertext: &mut [u8],
-            tag: &mut [u8; TAG_LEN],
-            key: &[u8; KEY_LEN],
-            nonce: &[u8; NONCE_LEN],
+            tag: &mut [U8; TAG_LEN],
```
jschneider-bensch, line 38: Does the tag need to be secret for encryption? I could understand the concern in the case of decryption, to potentially be able to catch non-constant time comparisons to a recomputed tag, but here it's a public output like the ciphertext.

chacha20poly1305/src/impl_aead_trait.rs
```diff
@@ -49,19 +52,26 @@ mod impl_chachapoly {
...
             );
 
             Ok(())
         }
 
         fn decrypt(
-            plaintext: &mut [u8],
-            key: &[u8; KEY_LEN],
-            nonce: &[u8; NONCE_LEN],
+            mut plaintext: &mut [U8],
```
jschneider-bensch, line 69: I see that the declassification doesn't work without making this a mutable binding, but that's probably a small wart in `libcrux-secrets`, no? Something to keep in mind for later.

chacha20poly1305/src/impl_aead_trait.rs
keks: This is what @franziskuskiefer and I had come up with in a discussion. There were some concerns around some MACs maybe not being perfectly hiding or so. I agree it shouldn't be a problem in the real world, though. The cost of making the user dereference is also low though, so I think it's fine for now.

chacha20poly1305/src/impl_aead_trait.rs
keks: Yes, exactly. Somehow the secrets traits require this to be mutable, not sure why.

chacha20poly1305/src/lib.rs
```diff
@@ -29,7 +30,7 @@
 //!       174, 123, 232 ]
 //! );
 //! assert_eq!(
-//!     tag.as_ref(),
+//!     &tag.as_ref().declassify(),
```
jschneider-bensch:
```suggestion
//!     tag.as_ref().declassify_ref(),
```
Otherwise I still get a failed build, now without the `check-secret-independence` feature.

keks:
Ugh, it looks like I forgot adding the check-secret-indepence stuff to the aead crate

PR #1185, opened 2025-09-30
Wysiwys/aes gcm cleanup

### LINKED ISSUE #1166: [AES-GCM] Cleanup to prepare crate for release

wysiwys:
[AES-GCM] Cleanup to prepare the new `libcrux-aesgcm` crate for release
- Addresses the points in https://github.com/cryspen/libcrux/issues/1166
- Reorganize module structure, so that the exported `Key`, `Nonce`, `Tag` type aliases can be found in a module that corresponds to each implementation
- Rename crate `libcrux_aesgcm` -> `libcrux-aesgcm` (for consistency with other crates in this repository)
- Apply clippy lints
- Update tests, benchmarks and examples to use `libcrux_traits`-based public APIs (the ones defined in `typed_owned` and `typed_refs`)
- Move `libcrux_traits` API implementation into separate module
- E.g. `libcrux_aesgcm::aes_gcm_128::neon::{Key, Nonce, Tag}`

Closes #1166


crates/algorithms/aesgcm/src/lib.rs
```diff
@@ -0,0 +1,306 @@
+//! AES-GCM.
+//!
+//! This crate provides separate modules for neon and x64.
```
franziskuskiefer: We should of course add examples etc. here.
But for these we should say something like "platform specific" and "only use these directly after performing runtime checks for the necessary CPU features.

crates/algorithms/aesgcm/src/aes_gcm_128.rs
```diff
@@ -0,0 +1,118 @@
...
+//!
+//! let k: Key = [0; AesGcm128::KEY_LEN].into();
+//! let nonce: Nonce = [0; AesGcm128::NONCE_LEN].into();
+//! let mut tag: Tag = [0; AesGcm128::TAG_LEN].into();
+//!
+//! let pt = b"the quick brown fox jumps over the lazy dog";
+//! let mut ct = [0; 43];
+//! let mut pt_out = [0; 43];
+//!
+//! k.encrypt(&mut ct, &mut tag, &nonce, b"", pt).unwrap();
```
franziskuskiefer: It would maybe be nice to know here what the errors could be. Something like "See [`EncryptError`] for errors".

franziskuskiefer (commented):
Looks like there's an `AesGcm128` (and 256) in the top level as well as in the `aes_gcm_128` module. That's a little odd? Looks like we have multiple APIs exposed here for the same thing.

crates/algorithms/aesgcm/src/traits_api.rs
```diff
@@ -0,0 +1,217 @@
...
+                impl_traits_public_api!($multiplexing, KEY_LEN, TAG_LEN, NONCE_LEN);
+
+                impl arrayref::Aead<KEY_LEN, TAG_LEN, NONCE_LEN> for $multiplexing {
+                    fn encrypt(
+                        ciphertext: &mut [u8],
+                        tag: &mut Tag,
+                        key: &Key,
+                        nonce: &Nonce,
+                        aad: &[u8],
+                        plaintext: &[u8],
```
jschneider-bensch, line 62: This, and `decrypt` should get an an additional check, that `plaintext.len() / AES_BLOCK_LEN < (u32::MAX - 1) as usize` to make sure we don't trigger the condition described here: https://github.com/cryspen/libcrux/pull/1167#discussion_r2372191735

crates/algorithms/aesgcm/benches/aesgcm.rs
```diff
@@ -0,0 +1,212 @@
...
+                                k.encrypt(&mut ciphertext, tag, nonce, &aad, &payload)
+                                    .unwrap();
+                            },
+                            BatchSize::SmallInput,
+                        )
+                    },
+                );
+
+                #[cfg(all(target_arch = "aarch64", target_feature = "aes"))]
+                group.bench_with_input(
```
franziskuskiefer, line 63: Looks like the benchmarks are not compiling.
We should at least build them on CI to catch it.

crates/algorithms/aesgcm/tests/wycheproof.rs
```diff
@@ -0,0 +1,84 @@
...
+                println!("  Test AES-GCM 128 {}", test.tc_id);
+
+                // Multiplexing
+                run(&test, libcrux_aesgcm::AesGcm128 {});
+
+                // Portable
+                run(&test, libcrux_aesgcm::PortableAesGcm128 {});
+
+                // Neon
+                #[cfg(all(target_arch = "aarch64", target_feature = "aes"))]
```
franziskuskiefer: We should use the features here as well instead of the `target_arch`.

crates/algorithms/aesgcm/src/lib.rs
franziskuskiefer: Looks like this is still open.
Also, the docs front page should give some examples.

franziskuskiefer (dismissed):

crates/algorithms/aesgcm/benches/aesgcm.rs
wysiwys: :+1: Should the CI workflow then run `cargo build` with the `--lib`, `--examples`, and `--benches` flags? Or would it be better to instead build the benchmarks in a separate step?

crates/algorithms/aesgcm/src/traits_api.rs
wysiwys: I added these checks, but along the way noticed that there is no input validation yet that checks the ciphertext length against the plaintext length. If the lengths are different, some debug assertions fail.

One way to address this would be to check that the lengths are equal inside the public `encrypt` and `decrypt` methods (if that is not too strict of a check).

crates/algorithms/aesgcm/src/lib.rs
jschneider-bensch: I think it would be good to copy any one of the examples from `aes_gcm_128/256` to the crate level docs in `lib.rs`. I think the owned version looks the nicest.

crates/algorithms/aesgcm/src/traits_api.rs
```diff
@@ -0,0 +1,294 @@
...
+
+                    fn encrypt(
+                        ciphertext: &mut [u8],
+                        tag: &mut Tag,
+                        key: &Key,
+                        nonce: &Nonce,
+                        aad: &[u8],
+                        plaintext: &[u8],
+                    ) -> Result<(), EncryptError> {
+                        if plaintext.len() / crate::aes::AES_BLOCK_LEN >= (u32::MAX - 1) as usize {
```
franziskuskiefer: There's lots of code duplication here. That's not great. We should move all checks into the common functions instead.
These implementations shouldn't do much more than passing things along to the actual functions, picking the right one, and converting arguments where necessary.

crates/algorithms/aesgcm/src/aes_gcm.rs
```diff
@@ -0,0 +1,128 @@
...
+                }
+            }
+        }
+    };
+}
+
+pub(crate) use aesgcm;
+
+/// Helper module for implementing platform-specific modules
+macro_rules! platform_mod {
```
franziskuskiefer: Do we really need all these types? The types aren't different for different platforms.

wysiwys: These type aliases are mainly provided for convenience, as opposed to using the underlying generic types. This is the same approach used in the `libcrux-chacha20poly1305` crate, where a separate set of type aliases is exported for each of `chacha20poly1305` and `xchacha20poly1305`. 

However, since there is overlap in the algorithm structs in this crate, using the generic structs directly may be better. Directly using the generic structs defined in `libcrux_traits::aead::typed_owned` would look like this:
```rust
    let k: Key<AesGcm128> = [0; AesGcm128::KEY_LEN].into();
    let nonce: Nonce<AesGcm128> = [0; AesGcm128::NONCE_LEN].into();
    let mut tag: Tag<AesGcm128> = [0; AesGcm128::TAG_LEN].into();
```

jschneider-bensch: Sorry for raising this late: We had some discussion of these comments [offline](https://chat.google.com/room/AAQA6lfp4Zw/dmp64YHnH8w/dmp64YHnH8w?cls=10) last week, and there we concluded that having the type aliases for the multiplexed implementation is good, but for the others we should remove the aliases and make the generic explicit, like you propose. So, `Key` will give you `Key<AesGcm128>` and anything else will have always have to make the underlying implementation explicit.

jschneider-bensch:
I think the public facing part of this is good and ready for review. For cleaning up the internals, I've created #1202 for a follow-up.

crates/algorithms/aesgcm/examples/bench_typed_owned.rs
```diff
@@ -0,0 +1,23 @@
+// structs for typed_owned
+use libcrux_aesgcm::{
+    aes_gcm_128::{Key, Nonce, Tag},
+    AeadConsts as _, AesGcm128,
+};
+
+fn main() {
+    const PAYLOAD_SIZES: usize = 3045;
```
keks: Any particular reason for this size? Should we make this a list of sizes that we iterator over or do we just use one? In that case, maybe this should be PAYLOAD_SIZE?

crates/algorithms/aesgcm/examples/bench_typed_refs.rs
```diff
@@ -0,0 +1,24 @@
+// structs for typed_owned
+use libcrux_aesgcm::{Aead as _, AeadConsts as _, AesGcm128};
+
+fn main() {
+    const PAYLOAD_SIZES: usize = 3045;
```
keks: same here

crates/algorithms/aesgcm/src/aes_gcm.rs
keks: In that case, maybe let's not call this `platform_mod`, but maybe `type_aliases` or so? It doesn't seem to be about platforms.

crates/algorithms/aesgcm/src/lib.rs
```diff
@@ -0,0 +1,529 @@
...
+    /// For more information on usage, see [`aes_gcm_128`].
+    pub struct NeonAesGcm128;
+    #[cfg(not(feature = "simd128"))]
+    /// Access to [lower-level AEAD APIs](libcrux_traits::aead) for portable AES-GCM 128 (no ARM NEON available).
+    ///
+    /// Should only be used directly after performing runtime checks for the necessary CPU
+    /// features.
+    ///
+    /// For more information on usage, see [`aes_gcm_128`].
+    pub type NeonAesGcm128 = PortableAesGcm128;
```
keks: I wonder if it makes sense to do something that the caller doesn't intend here, instead of letting this error if they didn't set the appropriate feature flags.

crates/algorithms/aesgcm/src/lib.rs
```diff
@@ -0,0 +1,529 @@
...
+    if plaintext.len() / crate::aes::AES_BLOCK_LEN >= (u32::MAX - 1) as usize {
+        return Err(DecryptError::PlaintextTooLong);
+    }
+
+    // ensure ciphertext and plaintext have same length
+    if ciphertext.len() != plaintext.len() {
+        return Err(DecryptError::WrongPlaintextLength);
+    }
+    debug_assert!(nonce.len() == NONCE_LEN);
+    debug_assert!(tag.len() == TAG_LEN);
```
keks: can we have these at the top of the function?

crates/algorithms/aesgcm/src/platform/x64/gf128_core.rs
```diff
@@ -0,0 +1,184 @@
...
+//     let m0 = unsafe { _mm_srli_si128(mid, 8) };
+//     let m1 = unsafe { _mm_slli_si128(mid, 8) };
+//     let low = unsafe { _mm_xor_si128(low, m0) };
+//     let high = unsafe { _mm_xor_si128(high, m1) };
+
+//     let low128: u128 = unsafe { core::mem::transmute(low) };
+//     let high128: u128 = unsafe { core::mem::transmute(high) };
+
+//     (FieldElement(low128), FieldElement(high128))
+// }
```
keks: Can we remove this?

crates/algorithms/aesgcm/src/lib.rs
jschneider-bensch: I believe we have to define these, since we have to refer to them in the runtime multiplexing (see the `traits_api::api` macro).

crates/algorithms/aesgcm/examples/bench_typed_owned.rs
jschneider-bensch: Thanks, I changed the name for now :+1:

crates/algorithms/aesgcm/src/lib.rs
keks: Ah, I see. Hm, can't we make the imports feature-gated over there? I mean that right now we do this

```
use crate::implementations::*;


api!(
    aes128,
    aes_gcm_128,
    AesGcm128,
    PortableAesGcm128,
    NeonAesGcm128,
    X64AesGcm128
);
```

and instead we could do

```
use crate::implementations::PortableAesGcm128;

#[feature = "simd128"]
use crate::implementations::NeonAesGcm128;
#[not(feature = "simd128")]
use crate::implementations::PortableAesGcm128 as NeonAesGcm128;

#[feature = "simd256"]
use crate::implementations::X64AesGcm128;
#[not(feature = "simd256")]
use crate::implementations::PortableAesGcm128 as X64AesGcm128;

api!(
    aes128,
    aes_gcm_128,
    AesGcm128,
    PortableAesGcm128,
    NeonAesGcm128,
    X64AesGcm128
);
```

That way we can just hide the value here. Does that make sense?

jschneider-bensch: Right, that works great. Thanks!

PR #1186, opened 2025-09-30
DHKEM from ECDH trait

jschneider-bensch:
This PR implements an [RFC 9180-style DHKEM(X25519, HKDF-SHA256)](https://www.rfc-editor.org/rfc/rfc9180#name-dh-based-kem-dhkem), sans Auth, on top of the ECDH trait in `libcrux_curve25519` (only that one, for now).

The implementation is behind a `kem-api` feature, since users may want to implement their own KEM on top of the ECDH trait, e.g. with a different KDF.

There are some things I'm not quite satisfied with:
- While it is possible to make the `arrayref::Kem` implementation generic over some `const SHARED_SECRET_LEN: usize`, it is not possible to make the macro-based implementation of `slice::Kem` generic. For this reason, I've fixed the `SHARED_SECRET_LEN == 32` in both for now.
- The implementation is not really specific to X25519, except for hardcoded (at the moment) KEM identifiers taken from RFC 9180. For this reason, I would really like to avoid duplicating it almost exactly in `p256`. Some things I've thought about/tried:
  - Blanket impl in `libcrux_traits`: Does not work because, `libcrux_hkdf` depends on `libcrux_traits` via `libcrux_sha2`. ❌
  - Blanket impl in `libcrux_kem`: I think this does not work because of the orphan rule, since both traits are foreign. ❌
  - Yet another macro in `libcrux_traits::kem`: Maybe this could work, but I dislike it, because it makes it harder to understand how things work, and because the macro would have to assume a crate using it has the HKDF dependency, which is not present in `libcrux_traits` (forbidden there, actually!).❓
- ~I've also had to touch the XWing implementation in `libcrux-kem`, since it was using the KEM trait API for X25519, assuming it gave you the raw ECDH shared secret. Instead it now uses the ECDH API, which one the one side _is_ what you want to use there. On the other hand the usage of `EcdhArrayref` is a bit inconvienient I found.~ (Errors handling can also be improved.)

### DIFF
--- a/curve25519/src/kem_api.rs
+++ b/curve25519/src/kem_api.rs
@@ -47,29 +47,27 @@
 use super::{DK_LEN, EK_LEN, X25519};
 use libcrux_hkdf::HkdfMode;
 use libcrux_secrets::{ClassifyRef as _, ClassifyRefMut as _};
-use libcrux_traits::ecdh::arrayref::EcdhArrayref;
 pub use libcrux_traits::kem::arrayref::Kem;
+use libcrux_traits::{
+    ecdh::arrayref::EcdhArrayref,
+    kem::arrayref::{DecapsError, EncapsError, KeyGenError},
+};
 
 /// An internal error that occurs during DHKEM operations.
 enum DhKemError {
-    GenerateKeyPair,
     Hkdf,
 }
 
 pub const SHARED_SECRET_LEN: usize = 32;
 
-impl libcrux_traits::kem::arrayref::Kem<DK_LEN, EK_LEN, EK_LEN, SHARED_SECRET_LEN, DK_LEN, DK_LEN>
-    for X25519
-{
+impl Kem<DK_LEN, EK_LEN, EK_LEN, SHARED_SECRET_LEN, DK_LEN, DK_LEN> for X25519 {
     fn keygen(
         ek: &mut [u8; DK_LEN],
         dk: &mut [u8; EK_LEN],
         rand: &[u8; DK_LEN],
-    ) -> Result<(), libcrux_traits::kem::arrayref::KeyGenError> {
-        generate_ecdh_pair(ek, dk, rand)
-            .map_err(|_| libcrux_traits::kem::arrayref::KeyGenError::Unknown)?;
-
-        Ok(())
+    ) -> Result<(), KeyGenError> {
+        X25519::generate_pair(ek, dk.classify_ref_mut(), rand.classify_ref())
+            .map_err(|_| KeyGenError::InvalidRandomness)
     }
 
     fn encaps(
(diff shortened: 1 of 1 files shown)


libcrux-kem/src/kem.rs
franziskuskiefer, line 962: These have been here before, but we really shouldn't unwrap here. The first one would be safe if we checked the length of the thing we split. But the second one would need checking as well.

libcrux-kem/src/kem.rs
```diff
@@ -962,7 +964,16 @@ mod xwing {
             let dk_x: &mut [u8; X25519_DK_LEN] = dk_x.try_into().unwrap();
 
             MlKem768::keygen(ek_m, dk_m, rand_m)?;
-            X25519::keygen(ek_x, dk_x, rand_x)?;
+
+            <X25519 as EcdhArrayref<X25519_RAND_KEYGEN_LEN, X25519_DK_LEN, X25519_EK_LEN>>::generate_secret(
+                dk_x.classify_ref_mut(),
```
franziskuskiefer: The classifications are only necessary for the secrets feature. If we don't expose that for this trait, we can skip them for now.

libcrux-kem/src/kem.rs
```diff
@@ -962,7 +964,16 @@ mod xwing {
             let dk_x: &mut [u8; X25519_DK_LEN] = dk_x.try_into().unwrap();
 
             MlKem768::keygen(ek_m, dk_m, rand_m)?;
-            X25519::keygen(ek_x, dk_x, rand_x)?;
+
+            <X25519 as EcdhArrayref<X25519_RAND_KEYGEN_LEN, X25519_DK_LEN, X25519_EK_LEN>>::generate_secret(
```
franziskuskiefer: Here we really need something else. We can't ask anyone to write something like this.

jschneider-bensch:
Rebased onto `jonas/ecdh-keygen` to use `generate_pair` in Xwing.

keks:
I think the approach here makes sense!

Regarding how to abstract this: Since the blanket impl (had it worked) would likely have lived in libcrux-ecdh, maybe that would be a good place for the impl macro, and then we call it everywhere where we also implement ECDH? I agree it doesn't make this more readable but (a) I don't have an idea for how we could get it to work with blanked impls and (b) at least personally, I don't find macros _that_ much worse than blanket impls from a readability perspective :)

Regarding the extract-then-expand function: Is that exactly one of the HPKE key schedules? Maybe we could call it that.

libcrux-kem/src/kem.rs
franziskuskiefer: These docs fail to build.

curve25519/Cargo.toml
```diff
@@ -15,12 +15,16 @@ libcrux-hacl-rs = { version = "=0.0.3", path = "../hacl-rs/" }
 libcrux-macros = { version = "=0.0.3", path = "../macros" }
 libcrux-traits = { version = "=0.0.3", path = "../traits" }
 libcrux-secrets = { version = "=0.0.3", path = "../secrets"}
+libcrux-hkdf = { version = "=0.0.3", path = "../libcrux-hkdf", optional = true }
 
 [features]
+kem-api = ["libcrux-hkdf"]
```
franziskuskiefer, line 21: Do we have naming convention for these? It could also just be `kem`. I'm fine either way.

curve25519/src/kem_api.rs
```diff
@@ -0,0 +1,177 @@
...
+        ek: &[u8; EK_LEN],
+        rand: &[u8; DK_LEN],
+    ) -> Result<(), libcrux_traits::kem::arrayref::EncapsError> {
+        let mut ephemeral_secret = [0u8; DK_LEN];
+
+        generate_ecdh_pair(ct, &mut ephemeral_secret, rand)
+            .map_err(|_| libcrux_traits::kem::arrayref::EncapsError::Unknown)?;
+
+        let mut derived_ecdh = [0u8; EK_LEN];
+        <X25519 as EcdhArrayref<DK_LEN, DK_LEN, EK_LEN>>::derive_ecdh(
```
franziskuskiefer: Wouldn't `X25519::` be enough? Same for the other calls.

curve25519/src/kem_api.rs
```diff
@@ -0,0 +1,177 @@
...
+
+/// An internal error that occurs during DHKEM operations.
+enum DhKemError {
+    GenerateKeyPair,
+    Hkdf,
+}
+
+pub const SHARED_SECRET_LEN: usize = 32;
+
+impl libcrux_traits::kem::arrayref::Kem<DK_LEN, EK_LEN, EK_LEN, SHARED_SECRET_LEN, DK_LEN, DK_LEN>
```
franziskuskiefer: If we import `arrayref` this entire thing gets a little more readable.

curve25519/src/kem_api.rs
```diff
@@ -0,0 +1,177 @@
...
+        let mut derived_ecdh = [0u8; EK_LEN];
+        <X25519 as EcdhArrayref<DK_LEN, DK_LEN, EK_LEN>>::derive_ecdh(
+            derived_ecdh.classify_ref_mut(),
+            ek,
+            ephemeral_secret.classify_ref(),
+        )
+        .map_err(|_| libcrux_traits::kem::arrayref::EncapsError::InvalidEncapsKey)?;
+
+        extract_and_expand(ct, ss, ek, derived_ecdh)
+            .map_err(|_| libcrux_traits::kem::arrayref::EncapsError::Unknown)
```
franziskuskiefer: Why do we do additional things with the secret? This will make this API unusable in most usecases.

franziskuskiefer (changes requested):

libcrux-kem/src/kem.rs
jschneider-bensch: Right, took out that feature for now.

libcrux-kem/src/kem.rs
jschneider-bensch: It became much simpler after disambiguation by just importing the `EcdhArrayref` trait and not the other ones. Making "the right trait" discoverable for users is a different question, though.

curve25519/src/kem_api.rs
jschneider-bensch: You mean, why do we do the HKDF?
The KEM implementation we had before gave out the raw DH shared secret, but we [warn against doing that](https://github.com/cryspen/libcrux/blob/af5d954baba6e255de0c74cd9d7179b80389a783/libcrux-ecdh/src/x25519.rs#L20-L24) in `libcrux-ecdh`. This here provides specifically an implementation of DHKEM(X25519, HKDF-SHA256) under the KEM traits, which requires that round of HKDF to make the shared secret safe to use. If you need the raw DH shared secret, like in XWing for example, you can use the ECDH API.

Or do you mean something else?

libcrux-kem/src/kem.rs
jschneider-bensch: I've updated the broken links.

curve25519/src/kem_api.rs
franziskuskiefer: But the point of having a KEM API for ECDH as well, is to make it possible to have a uniform way of calling KEMs, if that's an ECC based one or a PQ one. Like we do for example in Bertie. That won't be possible with this API.
When should people use this then?

jschneider-bensch: Okay, I think we had a misunderstanding of what this should achieve then.
For me, this was a way to deal with the warning I linked above and provide a DHKEM implementation under the KEM API. It would be used when you need a DHKEM.

The previous KEM API did what you want already, it just did so without using the ECDH trait. Doing that on top of the ECDH trait should be easy, however: Just remove the  HKDF-related bits here. In fact, removing the HKDF bits makes it possible to have a blanket impl in `libcrux-traits`.

Shall I do that instead?

jschneider-bensch:
As discussed offline, this doesn't quite go in the right direction. We _do_ want the EC KEMs to output the raw secret for uniformity of use with PQ-KEMs in protocol implementations.

Perhaps this can be salvaged into an easy fix for users that shouldn't use the raw secret, but it's not an immediate priority at the moment.

PR #344, opened 2024-07-02
Documentation pass

jschneider-bensch:
In this PR I've updated the docs of the verified code to mostly align with those of the `pre-verification` code.

I've also changed the visibility of the `unpacked` API in the verified code until we have #330.

- [x] Rust docs
- [x] Readme

PR #1194, opened 2025-10-01
Fix doc warnings in `libcrux-hacl-rs`

jschneider-bensch:
First step towards #344.

This fixes a lot of warnings resulting from literal translation of HACL comments, which clutter up the doc output for the  other crates.
I did not change the content of the comments beyond what's needed to silence the warning, in particular, I did not e.g. change `uint64_t[8]` to an equivalent Rust type.


keks (approved):
Ah, nice, thanks!

PR #1208, opened 2025-10-14
Fix Proof CI

karthikbhargavan:
This PR adjusts the proofs for ML-KEM and ML-DSA to fix a CI regression.
The changes try to make the proofs easier for F*, but do not substantially change them.
We also make a fix to enable F* extraction on Aarch64, which was broken by a prior PR.


karthikbhargavan:
ML-DSA Proofs: https://github.com/cryspen/libcrux/actions/runs/18503209195
ML-KEM ProofS: https://github.com/cryspen/libcrux/actions/runs/18507567981

PR #1210, opened 2025-10-16
Implement (de-)classification of immutable slices for `cfg(hax)`

jschneider-bensch:
The issue in hax is really only for `&mut`-returning functions, so `ClassifyRef` and `DeclassifyRef` should be fine.


keks (approved):
Ah, nice that this works!

PR #1211, opened 2025-10-16
Integrate AES-GCM in the `libcrux-aead` crate

jschneider-bensch:


crates/primitives/aead/src/lib.rs
```diff
@@ -31,7 +34,7 @@
 //! let key = Key::from(key_bytes);
 //! let nonce = Nonce::from(nonce_bytes);
 //!
-//! key.encrypt(&mut ciphertext, &mut tag, &nonce, &[/* no aad */], msg)
+//! key.encrypt(&mut ciphertext, &mut tag, &nonce, &[/* //! no aad */], msg)
```
keks: is this intended?

crates/algorithms/aesgcm/src/aes_gcm_128.rs
```diff
@@ -7,8 +7,6 @@ use crate::{
     DecryptError, NONCE_LEN, TAG_LEN,
 };
 
-// NOTE: hidden here in favor of being re-exported at crate root
-#[doc(hidden)]
```
keks, line 11: Is there a reason you removed this?

crates/primitives/aead/src/multiplexed.rs
```diff
@@ -16,19 +16,47 @@ pub enum Aead {
...
 }
 
 /// A reference to a multiplexed AEAD key.
-#[cfg(any(feature = "chacha20poly1305", feature = "xchacha20poly1305"))]
+#[cfg(any(
+    feature = "chacha20poly1305",
+    feature = "xchacha20poly1305",
+    feature = "aesgcm128",
+    feature = "aesgcm256"
+))]
```
keks, line 35: maybe we could have a `base-defs` feature (maybe there is a better name) that is automatically enable once any of the above are enabled, and then in the code we can just do `#[cfg(feature = "base-defs")]`. We can do that in a followup or skip it entirely, if you want to get this in quickly

crates/algorithms/aesgcm/src/aes_gcm_128.rs
jschneider-bensch: I removed it it's not being re-exported at the crate root, and currently doesn't show up in the docs at all.

crates/primitives/aead/src/lib.rs
jschneider-bensch: Good catch! No idea, where that came from

crates/primitives/aead/src/multiplexed.rs
jschneider-bensch: That would make this more concise, right! You'd trigger this with a `build.rs`?
I'll file a follow-up issue for it.

crates/algorithms/aesgcm/src/aes_gcm_128.rs
jschneider-bensch: I added the re-export at the crate root for the two constants.

crates/primitives/aead/src/multiplexed.rs
jschneider-bensch: https://github.com/cryspen/libcrux/issues/1213

jschneider-bensch:
I've also updated the Readme to list AES-GCM as an available AEAD, clarifying it's verification status.

PR #1222, opened 2025-11-03
[Intrinsics] Fix incorrect `_vxarq_u64` intrisic

jschneider-bensch:
The manual implementation of `vxarq_u64` for platforms where that intrisic isn't available was using the wrong input for one of the shift operands.

This PR also adds Linux ARM64 runners to the build & test matrices for ML-DSA and ML-KEM CI.


nadimkobeissi:
So much for formally verified

ctz:
Is there an advisory/analysis of this bug? What are the implications and security impact?

jschneider-bensch:
> Is there an advisory/analysis of this bug? What are the implications and security impact?

Thank you for creating https://github.com/rustsec/advisory-db/pull/2493. We've replied there with a suggestion for the advisory text.

PR #1225, opened 2025-11-03
ML-DSA ACVP KATs

wysiwys:
Addresses part of #1007 
* Replaces ML-DSA ACVP KAT files with latest versions
* Makes these KATs available via `libcrux-kats` (and moves the needed parsing code into `libcrux-kats`)
* Adapts the tests in the `libcrux-ml-dsa` crate to account for the new format and new test vector variants introduced in the latest versions of the KATs

### DIFF
--- a/crates/testing/kats/src/acvp/mldsa/common.rs
+++ b/crates/testing/kats/src/acvp/mldsa/common.rs
@@ -42,6 +42,6 @@ impl<T: TestResult> Results<TestGroupResults<T>> {
             .tests
             .iter()
             .find(|t| t.tc_id() == tcId)
-            .unwrap()
+            .unwrap_or_else(|| panic!("testCase {} should have been deserialized", tcId))
     }
 }


crates/testing/kats/src/acvp/mldsa/common.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn tc_id(&self) -> usize;
+}
+
+impl<T: TestResult> Results<TestGroupResults<T>> {
+    #[allow(non_snake_case)]
+    pub fn find_expected_result(&self, tgId: usize, tcId: usize) -> &T {
+        self.testGroups
+            .iter()
+            .find(|tg| tg.tgId == tgId)
+            .unwrap()
```
jschneider-bensch: I don't disagree with panicking here, but I think that an `unwrap_or_else(|| panic!("test group {} should have been deserialized", tgId))` might be helpful in tracking down the issue when it happens.

crates/testing/kats/acvp/mldsa-1_1_0_40/keygen/expectedResults.json
jschneider-bensch, line 1: Do we document somewhere where the JSON files come from?

crates/testing/kats/src/acvp/mldsa/common.rs
jschneider-bensch, line 1: This might module might even live in the `acvp` module, since ACVP KATs for other primitives likely follow the same overarching structure, no?

crates/testing/kats/src/acvp/mldsa.rs
```diff
@@ -0,0 +1,53 @@
+pub mod common;
+pub mod keygen_schema;
+pub mod sign_schema;
+pub mod verify_schema;
+
+macro_rules! impl_tests {
```
jschneider-bensch, line 48: Similar to the `common` module, I think in the future this macro could be extended to implement loading of ACVP KAT files for other primitives as well. But it doesn't have to be fully generic just yet.

libcrux-ml-dsa/tests/acvp.rs
```diff
@@ -157,33 +73,47 @@ fn siggen() {
...
+                        tcId = test.tcId,
+                    );
+                    continue;
+                }
+                siggen_inner_internal(test, &results, kat.tgId, &parameter_set)
+            } else {
+                let pre_hash = kat.preHash == Some("preHash".to_string());
+                if pre_hash {
+                    if test.hashAlg.as_ref().map(String::as_str) != Some("SHAKE-128") {
+                        continue;
```
jschneider-bensch, line 92: We could also `eprintln!` a message here that a test was skipped because the pre-hash is not supported.

libcrux-ml-dsa/tests/acvp.rs
```diff
@@ -238,55 +223,80 @@ fn sigver() {
...
+                        tgId = kat.tgId,
+                        tcId = test.tcId,
+                    );
+                    continue;
+                }
+                sigver_inner_internal(test, &results, kat.tgId, &parameter_set);
+            } else {
+                let pre_hash = kat.preHash == Some("preHash".to_string());
+                if pre_hash && test.hashAlg.as_ref().map(String::as_str) != Some("SHAKE-128") {
+                    continue;
```
jschneider-bensch, line 242: This could have an error message as well.

crates/testing/kats/src/acvp/mldsa/common.rs
wysiwys: Thanks, that's a good point! I added a message there

crates/testing/kats/acvp/mldsa-1_1_0_40/keygen/expectedResults.json
wysiwys: The documentation and readme now include links to the usnistgov/ACVP-Server repo, including a link to the specific commit.

crates/testing/kats/src/acvp/mldsa/common.rs
wysiwys: I agree! This module could either be moved in this PR already, or moved into the main `acvp` module later on, when other ACVP KATs are added in #1007

jschneider-bensch: We don't have to move it now, I think, but it makes sense when we get more KATs :+1:

crates/testing/kats/src/acvp/mldsa/common.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[allow(non_snake_case)]
+    pub fn find_expected_result(&self, tgId: usize, tcId: usize) -> &T {
+        self.testGroups
+            .iter()
+            .find(|tg| tg.tgId == tgId)
+            .unwrap_or_else(|| panic!("testGroup {} should have been deserialized", tgId))
+            .tests
+            .iter()
+            .find(|t| t.tc_id() == tcId)
+            .unwrap()
```
jschneider-bensch: This one could get a similar treatment to the one above.

wysiwys: :+1:  That line is updated now, too.

PR #1227, opened 2025-11-04
impl Clone for generic structs.

mmaker:
I'm having trouble being able to clone public keys without wacky work-around via `AsRef` and `From`. 
Is there any way they could be made clone-able?


mmaker:
@franziskuskiefer  friendly bump here!

libcrux-ml-kem/src/types.rs
```diff
@@ -1,5 +1,6 @@
 macro_rules! impl_generic_struct {
     ($name:ident, $doc:expr) => {
+        #[derive(Clone)]
```
franziskuskiefer, line 3: This is also used for the private keys, which shouldn't make it that easy. Can you move the struct definition out of here? Then you can add clone only for the public keys.

franziskuskiefer (changes requested):
Cloning here is generally fine, but we should only do it for non private keys.

libcrux-ml-kem/src/types.rs
mmaker: I can, just to understand better the desired feature tho: why are secret keys not clone-able? 

Currently it's not possible to use secret keys across threads, or have it be serialized with zerocopy. I don't think cloning secret keys explicitly (differently than, say, `Copy`) causes security issues, right? 

Anyways, going first with the external implementation.

franziskuskiefer: It's not a strong requirement. But copying any secret bytes may be a security issue. Making it easier is just not great. But it's not a huge deal, and this bit is changing anyway. So maybe let's just keep it as is for now.

PR #1233, opened 2025-11-07
Update dependencies in flake.lock

protz:
Both googletest and json had min_versions that were < 3.5, for which recent CMake dropped support


protz:
Specifically, the build error is here: https://github.com/AeneasVerif/eurydice/actions/runs/19154743461/job/54752896185

```
CMake Error at /nix/store/4l0nmijdkl1yb857pv6vgqmdvjzlh1k3-source/CMakeLists.txt:1 (cmake_minimum_required):
       >   Compatibility with CMake < 3.5 has been removed from CMake.
       >
       >   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
       >   to tell CMake that the project requires at least <min> but has been updated
       >   to work with policies introduced by <max> or earlier.
       >
       >   Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
       >
       > 
       > -- Configuring incomplete, errors occurred!
       > make: *** [Makefile:131: test-libcrux] Error 1
```

it looks like CMake 4.1 dropped support for CMakeLists.txt containing old min_versions and projects like google test followed suit, see e.g.  https://github.com/google/googletest/commit/62046339790b8fb8a37e8fde5d2edcbf4ffe2a3a

Issue #1160, opened 2025-09-22
Restructure repo


franziskuskiefer:
Move crates around.

franziskuskiefer:
@keks  What is needed to get this done?

keks:
The PR is probably pretty much out of date. One issue there was that it tries to do the entire repo at once, which makes merge conflicts likely. Let's instead make a list of crates and see what makes sense. In the parentheses we have the name of the subdirectory of the crates folder that the crate should go into.

- [x] libcrux-aead
- [x] libcrux-aesgcm
- [x] libcrux-ml-dsa
- [x] libcrux-ml-kem
- [x] libcrux-hkdf
- [x] libcrux-digest
- [x] libcrux-secrets
- [x] libcrux-traits
- [x] libcrux-kats
- [x] libcrux-hacl (sys)
- [x] libcrux-platform (sys)
- [x] libjade-sys (sys)
- [x] libcrux-pqclean (sys)
- [x] lib25519 (sys)
- [x] libcrux-blake2 (algorithms)
- [x] libcrux-poly1305 (algorithms)
- [x] libcrux-chacha20poly1305 (algorithms)
- [x] libcrux-curve25519 (algorithms) 
- [x] libcrux-p256 (algorithms)
- [x] libcrux-sha2 (algorithms)
- [x] libcrux-ed25519 (algorithms)
- [x] libcrux-hmac (algorithms)
- [x] libcrux-rsa (algorithms)
- [x] libcrux-ecdsa (algorithms)
- [x] libcrux-sha3 (algorithms)
- [x] libcrux-hacl-rs (utils)
- [x] libcrux-macros (utils)
- [x] libcrux-intrinsics (utils)
- [x] core-models (utils)
- [x] libcrux-test-utils (utils)
- [ ] libcrux-psq (protocols)
- [ ] libcrux (lets put this directly into crates/libcrux)
- [ ] libcrux-ecdh (algorithms)
- [ ] libcrux-kem (primitives)
- [ ] benchmarks (testing)
- [ ] cavp (testing)
- [ ] libcrux-fuzz (testing)
- [ ] libcrux-ml-kem-fuzz (testing)
- [ ] libcrux-aesgcm-fuzz (testing)

Thoughts on order:
- The sys crates are not changing at all
- Most of the extracted hacl-rs crates and the primitive crates are pretty static, except the signatures one
- Not entirely sure about the testing ones, I can imagine this is pretty much across the board.

[...]

PR #1243, opened 2025-11-17
Move some of the crates into crates/ directory

keks:
This is a batch of crates that we are moving into the crates/ directory.

This is a step towards #1160.


jschneider-bensch:
Looks like some paths are not updated yet on CI.

jschneider-bensch (approved):
Looks good!

jschneider-bensch:
The macOS 13 runners will soon be unsupported, and had a scheduled brown-out. I've made an issue for migrating them: #1245

PR #1244, opened 2025-11-18
eurydice arm

franziskuskiefer:
[skip changelog]


jschneider-bensch:
I fixed the conditional syntax and moved the flag after `--` to pass it to cargo.

jschneider-bensch (approved):

PR #1248, opened 2025-11-19
Documentation for PSQv2 & client-authentication proposal

jschneider-bensch:
This PR adds
- signature based authentication for the initiator
  - to enable this, I've added support for (de-)serialization of key and signature structs using `tls_codec` to ML-DSA
  - the initiator authenticator, be it a long-term DH public key or signature verification key can be retrieved from the `Responder` once the first initiator message was processed for out-of-band verification. I've added a function `Responder::abort_handshake` to reset the responder if this verification fails.
  - To enforce that the initiator authenticator and responder public keys are available for deserialization, I've made them required inputs in the serialization as well.
- AES-GCM 128 support for the AEAD
- Documentation on the different protocol modes in `README.md`
- I've added a way to export secrets for a given context from the `Session`. This will export a secret derived as `HKDF-SHA256(K_session, context || "PSQ secret export")` where `context` can be provided by the application.

~TODO:~ 
- [x] Update changelog


libcrux-psq/README.md
```diff
@@ -2,10 +2,235 @@
...
+    - query handler f,
+    - response_aad, 
+    - registration handler f
+    
+A: 
+    (esk_A = x, epk_A = g^x) <- DH.KeyGen()
+    tx0 = hash(0 | ctx | pk_B | epk_A)
+    dh_shared_secret_outer = DH.Derive(esk_A, pk_B)
+    K_0  = KDF(dh_shared_secret_outer, tx0)
+    if pqpk_S provided
```
franziskuskiefer: This is not an input. So where is it coming from?
Making indices for parties a little clearer overall, also who is playing which role, would help in general as well.

libcrux-psq/README.md
```diff
@@ -2,10 +2,235 @@
...
+A: 
+    tx2 = hash(2 | tx1 | epk_B)
+    dh_shared_secret_response_1 = DH.Derive(sk_A, epk_B)
+    dh_shared_secret_response_2 = DH.Derive(esk_A, epk_B)
+    K_2 = KDF(K_1 | dh_shared_secret_response_1 | dh_shared_secret_response_2, tx2)
+    response_payload = AEAD.Decrypt(K_2, enc_response, tag_response, response_aad)
+
+```
+
+### Client-authenticating Handshake (Proposal)
```
franziskuskiefer: Some high level description here on what is happening would be nice.

libcrux-psq/README.md
jschneider-bensch: Right, I've updated the indices and added more docs :+1:

libcrux-psq/README.md
jschneider-bensch: I've yet to provide a high-level description here.

jschneider-bensch: I've provided more documentation here, also on the inputs that we assume are provided/validated out-of-band to either participant.

jschneider-bensch:
Looks like the macOS ASAN nightly has gone a bit stale. Attempting to unpin that with #1258.

libcrux-psq/src/handshake/ciphersuite/types.rs
```diff
@@ -1,9 +1,60 @@
...
 use libcrux_ml_kem::MlKemSharedSecret;
 use tls_codec::{TlsDeserialize, TlsSerialize, TlsSerializeBytes, TlsSize};
 
 #[cfg(feature = "classic-mceliece")]
 use crate::classic_mceliece::{Ciphertext, PublicKey, SecretKey, SharedSecret};
+use crate::handshake::{ciphersuite::initiator::Auth, dhkem::DHPublicKey};
+
+#[derive(TlsSize, TlsDeserialize, TlsSerialize)]
+#[repr(u8)]
+pub enum Authenticator {
```
franziskuskiefer, line 23: Can you give all these public types some comments?

libcrux-psq/src/handshake/ciphersuite/responder.rs
```diff
@@ -1,44 +1,72 @@
...
-    pub longterm_ecdh_keys: &'a DHKeyPair,
+pub(crate) enum PqKemKeyPair<'a> {
+    None,
+    MlKem(&'a MlKem768PrivateKey, &'a MlKem768PublicKey),
+    #[cfg(feature = "classic-mceliece")]
+    CMC(&'a SecretKey, &'a PublicKey),
+    #[cfg(not(feature = "classic-mceliece"))]
+    #[allow(unused)]
+    CMC(
+        std::marker::PhantomData<&'a [u8]>,
```
franziskuskiefer: This is also in `core` (just to have less `std` deps) in case we want to go nostd at some point.

libcrux-psq/src/handshake/ciphersuite/responder.rs
```diff
@@ -47,155 +75,165 @@ impl<'a> CiphersuiteBase for ResponderCiphersuite<'a> {
...
-            #[cfg(not(feature = "classic-mceliece"))]
-            ResponderCiphersuite::X25519ClassicMcElieceChaCha20Poly1305HkdfSha256(_) => {
-                // We can never reach this because the ciphersuite can only be constructed with the feature turned on.
-                unreachable!("unsupported ciphersuite")
-            }
-            ResponderCiphersuite::X25519NoneChaCha20Poly1305HkdfSha256(suite) => {
-                &suite.longterm_ecdh_keys.sk
-            }
-        }
+    // fn tx0(&self, ctx: &[u8], peer_pk: &DHPublicKey) -> Result<Transcript, HandshakeError> {
```
franziskuskiefer: Remove?

libcrux-psq/src/handshake/ciphersuite/responder.rs
```diff
@@ -47,155 +75,165 @@ impl<'a> CiphersuiteBase for ResponderCiphersuite<'a> {
...
-                    .longterm_ecdh_keys
-                    .pk
+
+            (
+                SigVerificationKey::MlDsa65(mldsaverification_key),
+                Signature::MlDsa65(mldsasignature),
+            ) => libcrux_ml_dsa::ml_dsa_65::verify(
+                mldsaverification_key,
+                &payload,
+                b"PSQ",
```
franziskuskiefer: This is hard coded on both sides. If you move it out to a constant, it would be a little less error prone.

libcrux-psq/src/handshake/ciphersuite/builder.rs
```diff
@@ -81,6 +86,36 @@ impl<'a> CiphersuiteBuilder<'a> {
         self
     }
 
+    /// Provide a principal's long-term ML-DSA signing key.
+    pub fn longterm_mldsa_signing_key(mut self, signing_key: &'a MLDSA65SigningKey) -> Self {
```
franziskuskiefer: We have `ecdh`, `mlkem`, `mldsa`, `ed25519`. The `ecdh` is the only generic one. Is that on purpose?

libcrux-psq/src/aead.rs
```diff
@@ -26,34 +48,56 @@ pub(crate) enum AEADError {
...
+                    key: AEADKey::AesGcm128(key),
+                    nonce: [0u8; 12],
+                })
+            }
+        }
     }
 
     fn increment_nonce(&mut self) -> Result<(), AEADError> {
-        if self.1 == [0xff; NONCE_LEN] {
+        if self.nonce == [0xff; 12] {
```
franziskuskiefer: Is there constant anymore? 😔

libcrux-psq/src/handshake.rs
```diff
@@ -82,8 +100,58 @@ pub(crate) struct HandshakeMessage {
...
+        /// A Diffie-Hellman KEM public key
+        pk: DHPublicKey,
+        /// The AEAD-encrypted message payload
+        ciphertext: VLBytes,
+        /// AEAD tag authenticating the ciphertext and any AAD
+        tag: [u8; 16],
+        /// Associated data, covered by the AEAD message authentication tag
+        aad: VLBytes,
+        /// An optional post-quantum key encapsulation
+        pq_encapsulation: VLBytes,
```
franziskuskiefer: Would it make sense to share the common part here?

libcrux-psq/src/handshake.rs
```diff
@@ -128,51 +196,92 @@ pub(super) fn derive_k0(
...
     }
 
     let ecdh_shared_secret = DHSharedSecret::derive(own_longterm_key, peer_longterm_pk)?;
+    let ikm = K1IkmDh {
+        k0,
+        ecdh_shared_secret: &ecdh_shared_secret,
+        pq_shared_secret,
+    };
+
+    // XXX: This makes clippy unhappy, but is a lifetime error for feature `classic-mceliece` if we return directlyr
```
franziskuskiefer: Disable the lint here then.

```suggestion
    // XXX: This makes clippy unhappy, but is a lifetime error for feature `classic-mceliece` if we return directly
```

libcrux-psq/src/session.rs
```diff
@@ -189,7 +220,7 @@ impl Session {
     // the validation.
     pub fn deserialize(
         bytes: &[u8],
-        initiator_ecdh_pk: &DHPublicKey,
+        initiator_authenticator: &Authenticator,
```
franziskuskiefer, line 244: Into directly here doesn't work. But the into on the outside is a little ugly. You could add an `authenticator` function to the `DHKeyPair` that does it.

libcrux-psq/tests/registration.rs
```diff
@@ -47,26 +54,107 @@ impl From<SessionError> for TestError {
...
+            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_CHACHA20POLY1305_HKDFSHA256
+            | CiphersuiteName::X25519_NONE_MLDSA65_AESGCM128_HKDFSHA256
+            | CiphersuiteName::X25519_MLKEM768_MLDSA65_AESGCM128_HKDFSHA256
+            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_AESGCM128_HKDFSHA256 => {
+                Authenticator::Sig(SigVerificationKey::MlDsa65(
+                    self.initiator_mldsa_keys.verification_key.clone(),
+                ))
+            }
+        }
+        #[cfg(not(feature = "classic-mceliece"))]
```
franziskuskiefer: Do you need to duplicate all of these matches? Couldn't you just add the additional bits with the feature flag?

libcrux-psq/tests/registration.rs
```diff
@@ -47,26 +54,107 @@ impl From<SessionError> for TestError {
...
+        match name {
+            CiphersuiteName::X25519_NONE_X25519_CHACHA20POLY1305_HKDFSHA256
+            | CiphersuiteName::X25519_MLKEM768_X25519_CHACHA20POLY1305_HKDFSHA256
+            | CiphersuiteName::X25519_NONE_X25519_AESGCM128_HKDFSHA256
+            | CiphersuiteName::X25519_MLKEM768_X25519_AESGCM128_HKDFSHA256 => {
+                Authenticator::Dh(self.initiator_ecdh_keys.pk.clone())
```
franziskuskiefer: This could just be that `authenticator` function on the dh key pair.

libcrux-psq/src/handshake/responder.rs
```diff
@@ -167,46 +188,121 @@ impl<'a, Rng: CryptoRng> Responder<'a, Rng> {
...
+                ciphertext,
+                tag,
+                aad,
+                pq_encapsulation,
+            } => {
+                // Whether we attempt to do a decapsulation is decided by the working ciphersuite.
+                let pq_encapsulation_deserialized =
+                    working_ciphersuite.deserialize_encapsulation(pq_encapsulation.as_ref())?;
+
+                let tx1 = tx1_dh(
```
franziskuskiefer: Why not move the match into the `tx1` and `derive_k1` functions? Then everything else appears to be the same.

libcrux-psq/src/handshake/ciphersuite/responder.rs
jschneider-bensch: Good point!

libcrux-psq/src/handshake/ciphersuite/builder.rs
jschneider-bensch: Not really, no. X25519 is the only choice here. We can make it explicit as well.

libcrux-psq/src/handshake.rs
jschneider-bensch: Indeed, I've pulled out an `AuthMessage` component that deals with the authentication bits and the rest is shared now.

libcrux-psq/tests/registration.rs
jschneider-bensch: I don't think it's possible to feature flag part of a pattern match, so I tried separating the CMC patterns and doing the `cfg` inside the bodies. It doesn't work for some reason, but it also wouldn't look much nicer.

libcrux-psq/src/handshake/responder.rs
jschneider-bensch: I've addressed this for `tx1` which somewhat simplified the code.

libcrux-psq/tests/registration.rs
jschneider-bensch: Since the initiator and receiver both have to turn public keys into authenticators, I'd put that function on `DHPublicKey` instead. It would be best if I introduced a signing key pair wrapper as well, which can then also provide this `authenticator` function. That would make the outside API more uniform.

libcrux-psq/src/session.rs
jschneider-bensch: I've given it a try with the `ProvideAuthenticator` trait which lets you get the authenticator portions from various key materials.

franziskuskiefer: That works too. It would work without the trait too. But it's a nice forcing mechanism.

libcrux-psq/src/handshake/ciphersuite.rs
```diff
@@ -17,24 +18,124 @@ pub mod types;
...
+    /// AesGcm128 as payload AEAD and SHA-256 for HKDF. (requires
+    /// feature `classic-mceliece`)
+    X25519_CLASSICMCELIECE_MLDSA65_AESGCM128_HKDFSHA256,
 }
 
 impl CiphersuiteName {
+    /// Returns the default ciphersuite used for query mode.
+    pub(crate) fn query_ciphersuite() -> Self {
+        Self::X25519_NONE_X25519_CHACHA20POLY1305_HKDFSHA256
+    }
```
franziskuskiefer, line 138:
```suggestion
    }

```

libcrux-psq/tests/registration.rs
franziskuskiefer: Maybe I misunderstand, but sure it is. Here a solution could be for example the following. Or even simpler as above with all the enum variants being defined.

```rust

            #[cfg(feature = "classic-mceliece")]
            CiphersuiteName::X25519_CLASSICMCELIECE_X25519_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_X25519_AESGCM128_HKDFSHA256 => {
                Authenticator::Dh(self.initiator_x25519_keys.pk.clone())
            }

            #[cfg(feature = "classic-mceliece")]
            CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_AESGCM128_HKDFSHA256 => {
                Authenticator::Sig(SignatureVerificationKey::Ed25519(
                    self.initiator_ed25519_keys.1,
                ))
            }

            #[cfg(feature = "classic-mceliece")]
            CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_AESGCM128_HKDFSHA256 => {
                Authenticator::Sig(SignatureVerificationKey::Ed25519(
                    self.initiator_ed25519_keys.1,
                ))
            }

            #[cfg(feature = "classic-mceliece")]
            CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_AESGCM128_HKDFSHA256 => {
                Authenticator::Sig(SignatureVerificationKey::MlDsa65(Box::new(
                    self.initiator_mldsa_keys.verification_key.clone(),
                )))
            }

            #[cfg(not(feature = "classic-mceliece"))]
            CiphersuiteName::X25519_CLASSICMCELIECE_X25519_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_AESGCM128_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_AESGCM128_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_X25519_AESGCM128_HKDFSHA256 => {
                panic!("unsupported ciphersuite")
            }
```

libcrux-psq/tests/registration.rs
```diff
@@ -47,22 +54,168 @@ impl From<SessionError> for TestError {
...
+            | CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_AESGCM128_HKDFSHA256
+            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_CHACHA20POLY1305_HKDFSHA256
+            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_AESGCM128_HKDFSHA256 => {
                 panic!("unsupported ciphersuite")
             }
-            CiphersuiteName::X25519_NONE_AESGCM128_HKDFSHA256
-            | CiphersuiteName::X25519_MLKEM768_AESGCM128_HKDFSHA256
-            | CiphersuiteName::X25519_CLASSICMCELIECE_AESGCM128_HKDFSHA256 => {
-                unimplemented!("AES-GCM 128 ciphersuites are not implemented yet")
+        }
```
franziskuskiefer, line 84:
```suggestion
                PQEncapsulationKey::MlKem(self.responder_mlkem_keys.public_key()),
            ),
            CiphersuiteName::X25519_CLASSICMCELIECE_X25519_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_X25519_AESGCM128_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_ED25519_AESGCM128_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_CHACHA20POLY1305_HKDFSHA256
            | CiphersuiteName::X25519_CLASSICMCELIECE_MLDSA65_AESGCM128_HKDFSHA256 => {
                #[cfg(feature = "classic-mceliece")]
                return Some(PQEncapsulationKey::CMC(&self.responder_cmc_keys.pk));

                #[cfg(not(feature = "classic-mceliece"))]
                panic!("unsupported ciphersuite")
            }
        }
```

libcrux-psq/tests/registration.rs
jschneider-bensch: I put in a `SigningKeyPair` wrapper that makes things more uniform here, and which provides an `authenticator` function, like `DHPublicKey`.

libcrux-psq/src/handshake/responder.rs
jschneider-bensch: I'm leaving this internal cleanup for later.

libcrux-psq/tests/registration.rs
jschneider-bensch: Yeah, that's better, thanks!

franziskuskiefer (approved):
Thanks. Let's get this in.

PR #1249, opened 2025-11-19
Fix paths when copying to boring in ml-kem extract.sh

keks:
I wanted to update boring, but the paths were out of date. 

[skip changelog]


jschneider-bensch (approved):
Thanks!
We have a CI job now that checks for changelog updates unless you put `[skip changelog]` in the PR description somewhere.

PR #1263, opened 2025-12-08
Fix hax extraction of crates depending on some libcrux crates.

maximebuyse:
When extracting a crate with hax that depends on some libcrux crates, we need these libcrux crates to compile under the `hax` cfg. This can be prevented by:
- Some items being marked `cfg(not(hax))` (using `hax_lib::exclude` is better to remove the problematic items for verification of libcrux using hax).
- The use of `libcrux-intrinsics` instead of driect use of intrinsics from core_arch needs to be consistent across different libcrux crates.

### DIFF
--- a/crates/utils/secrets/src/int/classify_public.rs
+++ b/crates/utils/secrets/src/int/classify_public.rs
@@ -2,11 +2,7 @@
 //! We give definitions for all conversions so that they can be tested.
 //! We define no-ops here and force inlining, to ensure that these are free.
 
-#[cfg(not(hax))]
 use crate::traits::*;
-
-#[cfg(hax)]
-use crate::traits::{ClassifyRef, DeclassifyRef, Scalar};
 
 // TODO: Remove hax exemptions once this is supported.
 //       See https://github.com/cryspen/hax/issues/1674.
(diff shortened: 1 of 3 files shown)


maximebuyse:
There is still a problem with the use of `transmute` between `u128` and `__m128i` under the hax feature, but this doesn't block extraction.

maximebuyse:
This breaks the hax extraction of ml-kem because of [this macro](https://github.com/cryspen/libcrux/blob/77f9a6783e03585cc1709840bbc23ee27a28e6ad/traits/src/kem/slice.rs#L106). We need the impl produced by the macro to be excluded from the extraction, but adding `hax_lib::exclude` inside the macro means other crates using this macro need to have `hax-lib` as a dependency, which is not the case now. Maybe we could add this attribute conditionnally.

franziskuskiefer:
@maximebuyse what's the state here? We should try to get this resolved to allow extraction of dependent crates.

maximebuyse:
> @maximebuyse what's the state here? We should try to get this resolved to allow extraction of dependent crates.

I think I found a way to make it work! I opened a new PR (https://github.com/cryspen/libcrux/pull/1312), I don't want to break things for https://github.com/freedomofpress/securedrop-protocol/pull/131 by messing up this branch.

PR #1265, opened 2025-12-11
Simplify AArch64 feature checks, use atomics instead of `static mut`

jrose-signal:
- All AArch64 CPUs support NEON / AdvSIMD, including 128-bit vectors

- [All Apple AArch64 CPUs support AES and SHA-2 even though some don't declare it.](https://github.com/RustCrypto/utils/issues/378#issuecomment-826985574) (Also, providing a constant value helps strip dead code.)

- Atomics: ensures that if two threads race to perform initialization, they can't end up seeing "INITIALIZED = true" but the actual feature values set to false. (This could happen even on x86 if the compiler reorders loads.)

[skip changelog]

### DIFF
--- a/crates/sys/platform/src/macos_arm.rs
+++ b/crates/sys/platform/src/macos_arm.rs
@@ -1,60 +1,7 @@
 //! Obtain particular CPU features for AArch64 on macOS
 
-use core::sync::atomic::{AtomicBool, Ordering};
-use libc::{c_char, c_void, sysctlbyname, uname, utsname};
-
-/// Check that we're actually on an ARM mac.
-/// When this returns false, no other function in here must be called.
-pub(crate) fn actually_arm() -> bool {
-    let mut buf = utsname {
-        sysname: [c_char::default(); 256],
-        nodename: [c_char::default(); 256],
-        release: [c_char::default(); 256],
-        version: [c_char::default(); 256],
-        machine: [c_char::default(); 256],
-    };
-    let error = unsafe { uname(&mut buf) };
-    // buf.machine == "arm"
-    // It could also be "arm64".
-    let arm = buf.machine[0] == 97 && buf.machine[1] == 114 && buf.machine[2] == 109;
-    error == 0 && arm
-}
-
-#[inline(always)]
-fn check(feature: &[i8]) -> bool {
-    let mut ret = 0i64;
-    let mut size = core::mem::size_of::<i64>();
-    let error = unsafe {
-        sysctlbyname(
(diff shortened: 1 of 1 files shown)


crates/sys/platform/src/linux_arm.rs
```diff
@@ -1,59 +1,49 @@
...
 pub(super) fn aes() -> bool {
     init();
-    unsafe { AES }
-}
-
-#[inline(always)]
-pub(super) fn adv_simd() -> bool {
-    init();
-    unsafe { ADV_SIMD }
+    AES.load(Ordering::Relaxed);
```
jschneider-bensch:
```suggestion
    AES.load(Ordering::Relaxed)
```

Thank you! It looks like there's a superfluous `;` here.

jrose-signal: Ooops. That's what I get for not re-running my cross-compilation tests after continuing to change things.

crates/sys/platform/src/macos_arm.rs
```diff
@@ -50,78 +44,44 @@ fn sysctl() {
...
-    if check(&FEAT_AES_STR) {
-        unsafe { AES = true };
-    }
-
     // hw.optional.arm.FEAT_PMULL
     const FEAT_PMULL_STR: [i8; 27] = [
         0x68, 0x77, 0x2e, 0x6f, 0x70, 0x74, 0x69, 0x6f, 0x6e, 0x61, 0x6c, 0x2e, 0x61, 0x72, 0x6d,
         0x2e, 0x46, 0x45, 0x41, 0x54, 0x5f, 0x50, 0x4d, 0x55, 0x4c, 0x4c, 0x00,
     ];
     if check(&FEAT_PMULL_STR) {
```
jschneider-bensch: From [this comment](https://github.com/golang/go/issues/42747#issuecomment-732215259) linked in the RustCrypto discussion, it seems that it's also safe to always assume `pmull` on Apple aarch64.

jschneider-bensch:
(I've edited the PR description to include `[skip changelog]`, clearing the changelog reminder for CI.)

crates/sys/platform/src/macos_arm.rs
jrose-signal: Oh, good catch, I didn't read far enough! Changed, and now there are *no* dynamic feature checks on macOS so I left a note about looking in the revision history if they're needed again.

jschneider-bensch: Thanks!

PR #1270, opened 2025-12-17
[ML-KEM] Overhaul Readme and Crate documentation

### LINKED ISSUE #1172: [ML-KEM] Enable feature documentation for docs.rs
And we should enable some more features, including pqcp for the `docs.rs` build.

_Originally posted by @franziskuskiefer in https://github.com/cryspen/libcrux/pull/1165#pullrequestreview-3257462624_

jschneider-bensch:
This PR updates the Readme for ML-KEM to include more information on crate features and different APIs available.
It's essentially an updated version of the [PQCP Readme](https://github.com/pq-code-package/rust-libcrux/blob/main/README.md).

The crate-level documentation is also updated accordingly, and `codec` is added to the features included in the `docs.rs` build.

Fixes #1172 

[skip changelog]


libcrux-ml-kem/src/lib.rs
```diff
@@ -51,16 +74,104 @@
...
...IFNWRyBSZXBvIE1peGVyIFRvb2xzIC0tPg0KPHN2ZyB3aWR0aD0iODAwcHgiIGhlaWdodD0iODAwcHgiIHZpZXdCb3g9IjAgMCAyNCAyNCIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj4NCjxwYXRoIGQ9Ik05IDEyTDExIDE0TDE1IDkuOTk5OTlNMjAgMTJDMjAgMTYuNDYxMSAxNC41NCAxOS42OTM3IDEyLjY0MTQgMjAuNjgzQzEyLjQzNjEgMjAuNzkgMTIuMzMzNCAyMC44NDM1IDEyLjE5MSAyMC44NzEyQzEyLjA4IDIwLjg5MjggMTEuOTIgMjAuODkyOCAxMS44MDkgMjAuODcxMkMxMS42NjY2IDIwLjg0MzUgMTEuNTYzOSAyMC43OSAxMS4zNTg2IDIwLjY4M0M5LjQ1OTk2IDE5LjY5MzcgNCAxNi40NjExIDQgMTJWOC4yMTc1OUM0IDcuNDE4MDggNCA3LjAxODMzIDQuMTMwNzYgNi42NzQ3QzQuMjQ2MjcgNi4zNzExMyA0LjQzMzk4IDYuMTAwMjcgNC42Nzc2NiA1Ljg4NTUyQzQuOTUzNSA1LjY0MjQzIDUuMzI3OCA1LjUwMjA3IDYuMDc2NCA1LjIyMTM0TDExLjQzODIgMy4yMTA2N0MxMS42NDYxIDMuMTMyNzEgMTEuNzUgMy4wOTM3MyAxMS44NTcgMy4wNzgyN0MxMS45NTE4IDMuMDY0NTcgMTIuMDQ4MiAzLjA2NDU3IDEyLjE0MyAzLjA3ODI3QzEyLjI1IDMuMDkzNzMgMTIuMzUzOSAzLjEzMjcxIDEyLjU2MTggMy4yMTA2N0wxNy45MjM2IDUuMjIxMzRDMTguNjcyMiA1LjUwMjA3IDE5LjA0NjUgNS42NDI0MyAxOS4zMjIzIDUuODg1NTJDMTkuNTY2IDYuMTAwMjcgMTkuNzUzNyA2LjM3MTEzIDE5Ljg2OTIgNi42NzQ3QzIwIDcuMDE4MzMgMjAgNy40MTgwOCAyMCA4LjIxNzU5VjEyWiIgc3Ryb2tlPSIjMDAwMDAwIiBzdHJva2Utd2lkdGg9IjIiIHN0cm9rZS1saW5lY2FwPSJyb3VuZCIgc3Ryb2tlLWxpbmVqb2luPSJyb3VuZCIvPg0KPC9zdmc+
```
franziskuskiefer: Having this everywhere adds lots of blob. We should add these labels to the repo at some point and just link to it.

Anyway, not for this PR. Just a thought.

libcrux-ml-kem/src/lib.rs
```diff
@@ -1,17 +1,40 @@
...
-//!
-//! Functions in this crate use CPU feature detection to pick the most efficient version
-//! on each platform. To use a specific version with your own feature detection
-//! use e.g. one of the following
-//! - `mlkem768::avx2::generate_key_pair`,
-//! - `mlkem768::neon::generate_key_pair`,
-//! - `mlkem768::portable::generate_key_pair`,
-//!
-//! analogously for encapsulation and decapsulation."
+//! This crate implements all three ML-KEM ([FIPS 203](https://csrc.nist.gov/pubs/fips/203/final)) variants 512, 768, and 1024.
```
franziskuskiefer: Looks like you duplicate the readme here? Why not include the readme instead?

franziskuskiefer (approved):
lgtm
If you can de-duplicate by including the readme, even better.

libcrux-ml-kem/src/lib.rs
jschneider-bensch: Right, I forgot you could do `#![doc = include_str!("../README.md")]`! That would be much nicer.

One issue there is that we have the doc test here that is feature-guarded for `mlkem768`, which we couldn't do with the include. So e.g. `cargo test --no-default-features -F mlkem512` will fail on that then.

franziskuskiefer: I'm not sure if/how you can run the tests from the readme. That never worked for me. But if it does, you could add the feature guards in there. Or keep the test in here.

jschneider-bensch: I've pushed a version that works. Unfortunately, just putting a `#![cfg(feature = "mlkem768")]` does not work, so I had to put a guarded code block in there. But it's not so bad,  I think.

franziskuskiefer (approved):
Thanks

PR #1277, opened 2025-12-29
Add missing models and some hax annotations

karthikbhargavan:
Fixed up a few annotations to allow Minos to extract.
Am not entirely sure whether these annotations are sufficient, so leaving as draft.
We should experiment with more examples, notably Bertie to make sure we have a "recipe" for crates that depend on libcrux and want to use hax.


jschneider-bensch:
@karthikbhargavan Is this superceded by something else already?

jschneider-bensch:
I'm closing this. If necessary we can re-open.

Issue #1271, opened 2025-12-17
[PSQ] Allow importing session secrets


jschneider-bensch:
This can be used for key rotation.

## Proposed mechanism

An external secret `psk` that is at least 32 bytes long can be imported into the
session, replacing the original main session secret `K_S` with a fresh
secret derived from the `psk` and `K_S`.
The new session will be bound to the same principal public keys, but
it will derive a fresh session ID and update the running transcript
`tx` for future imports.

In detail:
```
K_import = KDF(K_S || psk, "secret import")
tx' = Hash(tx || session_ID)

// From here: treat K_import as though it was the outcome of a handshake
K_S' = KDF(K_import, "session secret" | tx')
session_ID' = KDF(K_S', "shared key id")
```

We can provide this e.g. with the following API:
```Rust
impl Session {
    pub fn import(
        self,
        psk: &[u8],
        initiator_authenticator: &Authenticator,
        responder_ecdh_pk: &DHPublicKey,
        responder_pq_pk: Option<PQEncapsulationKey>,
    ) -> Result<Self, SessionError> {
    ...
    }
}
```

jschneider-bensch:
Closed by #1278

PR #1278, opened 2026-01-05
[PSQ] Implement import for session secrets

jschneider-bensch:
This PR implements session secret import. It follows the approach outlined in #1271, but also makes binding sessions to principal public keys optional:

When a new session is created it is always bound to the long-term public keys of intitiator and responder. When importing an external secret as outlined in #1271 this binding can be stripped from the session, i.e. unless the same binding information from the original session is provided, the new, imported session key will not be bound to any public key material. This should be used with caution, and we disallow introducing a new binding for a session once the binding has been stripped off.


libcrux-psq/src/session.rs
```diff
@@ -138,28 +142,45 @@ fn derive_pk_binder(
...
+    pub(crate) fn new<'a>(
         tx2: Transcript,
         k2: AEADKeyNonce,
-        initiator_authenticator: &Authenticator,
-        responder_ecdh_pk: &DHPublicKey,
-        responder_pq_pk: Option<PQEncapsulationKey>,
+        session_binding: Option<SessionBinding<'a>>,
+        // initiator_authenticator: &Authenticator,
+        // responder_ecdh_pk: &DHPublicKey,
+        // responder_pq_pk: Option<PQEncapsulationKey>,
```
franziskuskiefer: remove?

libcrux-psq/src/session.rs
```diff
@@ -138,28 +142,45 @@ fn derive_pk_binder(
...
     /// the session key to any long-term public key material that was used during the
     /// handshake.
-    pub(crate) fn new(
+    pub(crate) fn new<'a>(
         tx2: Transcript,
         k2: AEADKeyNonce,
-        initiator_authenticator: &Authenticator,
-        responder_ecdh_pk: &DHPublicKey,
-        responder_pq_pk: Option<PQEncapsulationKey>,
+        session_binding: Option<SessionBinding<'a>>,
```
franziskuskiefer, line 164: Maybe use an `impl Into<Option<_>>` to make it less noisy. If `Some` is the usual case.

jschneider-bensch: That's a good idea, thanks!

PR #1285, opened 2026-01-15
Upgrade hax to 0.3.6

maximebuyse:
[skip changelog]

### DIFF
--- a/.github/workflows/mlkem-hax.yml
+++ b/.github/workflows/mlkem-hax.yml
@@ -66,7 +66,7 @@ jobs:
       - uses: hacspec/hax-actions@main
         with:
           hax_reference: ${{ github.event.inputs.hax_ref || needs.get-hax-ref.outputs.hax_ref }}
-          fstar: v2025.10.06
+          fstar: v2025.10.06 # using an older version here because of https://github.com/cryspen/hax/issues/1878
 
       - uses: actions/download-artifact@v7
         with:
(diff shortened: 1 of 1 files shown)


maximebuyse:
The failed lax-checking for ML-kem was due to https://github.com/cryspen/hax/issues/1878. I was trying to use latest F* which caused some bad interaction, reverting to the version just before solves it (still an upgrade compared to what was used here previously). Hopefully this should fix the CI job and we can merge this PR!

.github/workflows/mlkem-hax.yml
```diff
@@ -67,7 +66,7 @@ jobs:
       - uses: hacspec/hax-actions@main
         with:
           hax_reference: ${{ github.event.inputs.hax_ref || needs.get-hax-ref.outputs.hax_ref }}
-          fstar: v2025.02.17
+          fstar: v2025.10.06
```
jschneider-bensch: If you could add a comment here that this is intentionally a slightly older version of F* because of https://github.com/cryspen/hax/issues/1878, then we can try bringing it in line with the other workflows once that issue is closed.

crates/utils/libcrux-core-models/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "core-models"
+name = "libcrux-core-models"
```
jschneider-bensch, line 2: You need to rename, because otherwise there's a name collision in the F* extractions now, right?

jschneider-bensch (approved):
Looks good, thanks!

crates/utils/libcrux-core-models/Cargo.toml
maximebuyse: Yes, exactly. We will probably have some version of these models provided by hax some day so we will be able to remove this from the libcrux repo.

PR #1291, opened 2026-01-21
ML-KEM Proofs Cleanup

karthikbhargavan:
This PR does a few things:

* Switches all the annotations in traits, polynomial, ntt, and invert_ntt to mostly use Rust (with a few F*-specific annotations)
* Simplifies the proofs by using opaque predicates, removing the need for a lot of F* annotations in modules above the vector abstraction

We should discuss this style and stabilize it before going further.
One design point is the use of `cfg(hax)` for specs. 
I was doing that uniformly before, but it is inconvenient when we start using Rust annotations more uniformly.

But maybe now the code is less cluttered, we don't mind having hax enabled everywhere? Dunno.
So some discussion is needed around how and whether to use conditional flags.


libcrux-ml-kem/src/vector/portable/arithmetic.rs
```diff
@@ -51,11 +54,8 @@ pub(crate) fn get_n_least_significant_bits(n: u8, value: U32) -> U32 {
 
 #[inline(always)]
 #[hax_lib::fstar::options("--z3rlimit 150")]
-#[hax_lib::requires(fstar!(r#"forall i. i < 16 ==> 
-    Spec.Utils.is_intb (pow2 15 - 1) (v (Seq.index ${lhs}.f_elements i) + v (Seq.index ${rhs}.f_elements i))"#))]
-#[hax_lib::ensures(|result| fstar!(r#"forall i. i < 16 ==> 
-    (v (Seq.index ${result}.f_elements i) == 
-     v (Seq.index ${lhs}.f_elements i) + v (Seq.index ${rhs}.f_elements i))"#))]
+#[hax_lib::requires(fstar!(r#"${spec::add_pre} ${lhs}.f_elements ${rhs}.f_elements"#))]
```
jschneider-bensch, line 57:
```suggestion
#[hax_lib::requires(spec::add_pre(&lhs.elements, &rhs.elements))]
```
These could also be written in the Rust style now, right?

libcrux-ml-kem/src/vector/portable/arithmetic.rs
```diff
@@ -77,16 +77,16 @@ pub fn add(mut lhs: PortableVector, rhs: &PortableVector) -> PortableVector {
         "assert (forall i. v (Seq.index ${lhs}.f_elements i) ==
     			          v (Seq.index ${_lhs0}.f_elements i) + v (Seq.index ${rhs}.f_elements i))"
     );
+    hax_lib::fstar!(
+        r#"reveal_opaque (`%Spec.Utils.is_i16b_array_opaque) (Spec.Utils.is_i16b_array_opaque)"#
```
jschneider-bensch, line 81: Is this related to the introduction of opaque predicates? Does it mean while the higher-level modules will have less F* annotations, we will need to `reveal_opaque` in the lower-levels?

libcrux-ml-kem/src/invert_ntt.rs
```diff
@@ -1,53 +1,40 @@
...
-         {| i1: Libcrux_ml_kem.Vector.Traits.t_Operations v_Vector |}
-         (re: Libcrux_ml_kem.Polynomial.t_PolynomialRingElement v_Vector) =
-       forall (i:nat). i < 16 ==> Spec.Utils.is_i16b_array_opaque (4 * 3328)
-            (Libcrux_ml_kem.Vector.Traits.f_to_i16_array (re.f_coefficients.[ sz i ]))"
-)]
-#[hax_lib::requires(fstar!(r#"v ${*zeta_i} == 128 /\
-    invert_ntt_re_range_1 $re"#))]
-#[hax_lib::ensures(|result| fstar!(r#"invert_ntt_re_range_2 ${re}_future /\
-    v ${*zeta_i}_future == 64"#))]
+#[hax_lib::requires(spec::is_bounded_poly(4 * 3328, re).and(*zeta_i == 128))]
```
jschneider-bensch: The spec functions return `hax_lib::Prop` instead of `bool`, which means you use `.and` here as the connective. That could be cumbersome in case there are a larger number of preconditions that have to be `.and`ed together. Would it be possible to implement e.g. `BitAnd` for `Prop` to be able to `&` two of them. (The short-circuiting `&&` is sadly not overloadable.)

libcrux-ml-kem/src/invert_ntt.rs
```diff
@@ -1,53 +1,40 @@
...
     for round in 0..16 {
         hax_lib::loop_invariant!(|round: usize| {
-            fstar!(
-                r#"v zeta_i == v $_zeta_i_init - v $round * 4 /\
-          (v round < 16 ==> (forall (i:nat). (i >= v round /\ i < 16) ==>
-            Spec.Utils.is_i16b_array_opaque (4 * 3328)
-              (Libcrux_ml_kem.Vector.Traits.f_to_i16_array (re.f_coefficients.[ sz i ])))) /\
-          (forall (i:nat). i < v $round ==> Spec.Utils.is_i16b_array_opaque 3328
-              (Libcrux_ml_kem.Vector.Traits.f_to_i16_array (re.f_coefficients.[ sz i ])))"#
+            hax_lib::fstar_prop_expr!(r#"v $zeta_i == v $_zeta_i_init - v $round * 4"#).and(
```
jschneider-bensch: This invariant is much more readable (to me)! Any reason to use the `fstar_prop_expr!` instead of `(*zeta_i == _zeta_i_init - round * 4).to_prop()`?

libcrux-ml-kem/src/invert_ntt.rs
```diff
@@ -1,53 +1,40 @@
...
+                hax_lib::forall(|i: usize| {
+                    if i < 16 {
```
jschneider-bensch, line 28: Maybe there could be a convenience macro for loop invariants like this, e.g. `pivot!(array, round, prop_lower_equal, prop_higher)` that states `prop_lower_equal` for `array[i]` with `i in 0..=round` and `prop_higher` for `i in (round + 1).. array.len()`.

libcrux-ml-kem/src/invert_ntt.rs
```diff
@@ -62,150 +49,164 @@ pub(crate) fn invert_ntt_at_layer_1<Vector: Operations>(
...
     (a, b)
 }
 
 #[inline(always)]
-#[hax_lib::fstar::verification_status(lax)]
-#[hax_lib::requires(fstar!(r#"v $layer >= 4 /\ v $layer <= 7"#))]
+#[hax_lib::fstar::options("--z3rlimit 200 --ext context_pruning")]
+#[hax_lib::requires(
+    spec::is_bounded_poly(3328, re).and(
+        match layer {
```
jschneider-bensch, line 148: It's neat to be able to use a match here, but it could also be written in terms of `layer`, i.e. `*zeta_i == (1 << (8 - layer))` (keeping the boundedness condition for `layer`). Does it make a difference for the proofs?

jschneider-bensch (commented):
Can you say more about the inconvenience of `cfg(hax)` when using Rust specs?

karthikbhargavan:
> Can you say more about the inconvenience of `cfg(hax)` when using Rust specs?

I undid that, it basically needs a few more `cfg(hax)` sprinkled around to make sure specs are only referred to in hax mode.

libcrux-ml-kem/src/vector/portable/arithmetic.rs
karthikbhargavan: I did not change anything under the trait boundary, there are a bug chunk of F* in there that needs to be switched. I focused on the code above the traits boundary for now.

libcrux-ml-kem/src/vector/portable/arithmetic.rs
karthikbhargavan: Yes, the observation is that the verification of higher-level modules was taking an unnecessarily long time because the underlying math was not opaque, even if the math itself is not needed for the proof. So we make it opaque at (or near) the trait boundary, so modules above can be verified quickly and the modules below will need to reveal these predicates when needed.

libcrux-ml-kem/src/invert_ntt.rs
karthikbhargavan: Yes, that would be a good feature upgrade. For now, I am trying to live within the constraints of the current hax-lib until libcrux is upgraded to latest hax.

libcrux-ml-kem/src/invert_ntt.rs
karthikbhargavan: no good reason, we should change to your version (earlier, I may have been trying to avoid to_prop but it was not a good reason).

libcrux-ml-kem/src/invert_ntt.rs
karthikbhargavan: sure, yes, once we have a bunch of annotations like these, we can try to find common patterns. (Copilot is already able to suggest these patterns to me now.)

libcrux-ml-kem/src/invert_ntt.rs
karthikbhargavan: I don't like using shifts in specs. Maybe you mean `pow2(8 - layer)` which may be nicer if that is what the NIST spec says e.g.

karthikbhargavan:
Force-pushed a bunch of changes, this is now moving to a new style where the specs and fully in Rust, and we cross-test code against the spec in Rust before we prove that the code meets the spec. Allows for a more incremental workflow. All annotations are being moved to Rust as well. Still WIP.

jschneider-bensch:
@karthikbhargavan Do we need a draft PR for this, before things have stabilized?

PR #1292, opened 2026-01-21
KMAC-128 and KMAC-256 implementation

kraemv:
Hello everyone,
I implemented KMAC-128 and KMAC-256 in Rust and would like to contribute my code to libcrux. I am not familiar with hax and F*, hence my commit does not contain a correctness proof.
I have tested my code with the wycheproof project, the test vectors and the changes to the /tests repository are included in this PR.
I made changes to sha3/src/portable.rs to implement an iterative CShake API with the correct Keccak padding. An iterative API is required to make the code no_std compatible.
Could you please share your thoughts about the length values (e.g. key_length) in hacl/kmac.rs, as they will panic for inputs larger than 2^61 bytes.


crates/algorithms/kmac/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+repository.workspace = true
+
+[lib]
+path = "src/kmac.rs"
+
+[features]
+expose-hacl = []
+
+[dependencies]
+libcrux-hacl-rs = { version = "=0.0.4", path = "../../utils/hacl-rs/" }
```
jschneider-bensch: `libcrux-hacl-rs` and `libcrux-macros` seem unnecessary here, no?

crates/algorithms/kmac/src/impl_hacl.rs
jschneider-bensch, line 1: It's not necessary to stick to the `impl_hacl/hacl` module structure here, since this doesn't use any HACL code as far as I can see. It could all live in the top-level `kmac` module.

crates/algorithms/kmac/src/kmac.rs
```diff
@@ -0,0 +1,19 @@
+//! KMAC
+//!
+//! This crate implements KMAC.
+#![no_std]
+
+extern crate alloc;
```
jschneider-bensch: Do you need to allocate anywhere, actually?

crates/algorithms/kmac/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+license.workspace = true
+homepage.workspace = true
+edition.workspace = true
+repository.workspace = true
+
+[lib]
+path = "src/kmac.rs"
+
+[features]
+expose-hacl = []
```
jschneider-bensch: We typically use features named like this to expose otherwise internal functions that come from HACL. If it's used to expose the internals here you could name it `expose-internal`, since there is no HACL code here.
But do you need to expose the internal functions?

jschneider-bensch (commented):
Thank you!
I haven't had time for a detailed look yet, just a few comments.

crates/algorithms/kmac/Cargo.toml
kraemv: They are removed now.

crates/algorithms/kmac/src/impl_hacl.rs
kraemv: I changed to an internal structure with helper functions (and an API, that is closer to the verified code) and the expected KMAC API in the top-level module.

crates/algorithms/kmac/src/kmac.rs
kraemv: I removed this now. It is a leftover from a previous implementation.

crates/algorithms/kmac/Cargo.toml
kraemv: I removed the feature, the inner functions should not be interesting for any other code.

franziskuskiefer:
@kraemv do you think you could add runtime safety, i.e. that the code never panics, if you help you with the set up?

jschneider-bensch:
In this [PR](https://github.com/kraemv/libcrux/pull/1) to your branch, I've added a basic setup for doing runtime safety proofs. Let me know what you think!

kraemv:
The fstar runtime safety proof now goes through.

jschneider-bensch:
It's very nice that you can extract and typecheck this code for runtime safety without any annotations!
I came up with some ideas for refactoring the code and tried them out on this branch: https://github.com/cryspen/libcrux/tree/jonas/refactor-kmac-cshake ([comparison](https://github.com/kraemv/libcrux/compare/feature/kmac...cryspen:libcrux:jonas/refactor-kmac-cshake))
- I reduced code duplication a bit using a const generic for the `RATE` and passing in the CSHAKE implementation as a generic as well.
- I moved the CSHAKE-specific parts out of KMAC and into a general purpose CSHAKE implementation in `sha3`. 
- To do this I needed to add some annotations for runtime safety, but the extraction should typecheck on that branch as well.

Would you be okay with incorporating these changes here?

kraemv:
Hi Jonas,
I like your changes to the code, they benefit readability and structure! Also, the structure is now closer to the definitions of KMAC and CSHAKE in the standard.
Do you think it is easier to proof functional correctness at some future point for a more generic code?
With the new structure, you should be able to remove CShake128It in [crates/algorithms/sha3/src/portable/cshake.rs] as it is superseded by CShake128 in [crates/algorithms/sha3/src/portable/incremental.rs].

jschneider-bensch:
I think we have found in ML-KEM and ML-DSA as well that doing a functional correctness proof on a generic implementation can be advantageous since you only do the proof once and it should hold for all different (valid!) parameters the generic implementation can be instantiated with.

I made a few more small changes on the refactoring branch, updating changelogs and removing the obsolete code you found. I think with these also merged into your PR this looks very good to me already!
One thing we should do before merging the PR is add some more documentation around the new CSHAKE and KMAC APIs. Just let me know if you'd like to give that a shot or if not, I can do it.

kraemv:
Well, then I am very happy to have the generic code now :)
I merged the most recent changes into this PR already. I am also very happy with the state of the code!
I would be grateful if you could do the documentation, as I am currently quite busy with another project.

jschneider-bensch:
Okay, then I'll take care of the rest. :+1:  Thanks a lot for your efforts!

kraemv:
Thank you for your feedback on the code and the support to prove runtime safety!

jschneider-bensch:
I've rebased this again on current main, and added a re-export for KMAC to the main libcrux crate. Still need to add docs and update the Readme.

EDIT: Also changing the title here, since we do have runtime safety for the implementation now.

jschneider-bensch (approved):
I added a little bit more documentation, I think this is now good to merge.

jschneider-bensch:
I'll resolve conflicts and see whether the hax issue persists.

robinhundt:
Ok, I got the sha3 and kmac hax extraction and F* proving now working (also in CI :partying_face:). This was quite tricky to resolve, as there we issues with the interface extraction which required excluding some modules, various hax caching issues which made working on this painful, and a gap in the proof of runtime safety for the high-level generic `CShake` API which is also used in kmac.

To resolve the last point, I've added a super trait to `CShake`, which when compiled with hax allows to thread through the `keccak_xof_state_inv(RATE, self.state.buf_len)` required/ensures clause, which is necessary to call the `KeccakXofState` methods in the `impl<const RATE: usize> CShake<RATE> for CShakeIncremental<RATE>` and in the generic use of `CShake` in kmac.

Debugging this and resolving the issues was heavily assisted by claude code (Opus 4.8).

@jschneider-bensch I've disable auto-merge, as my changes might warrant another look, specifically [this commit](https://github.com/cryspen/libcrux/pull/1292/changes/5e688a91799c5521c70b7de615280244cb787398) where I'm touching the actual kmac code.

CHANGELOG.md
```diff
@@ -20,13 +20,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- [#1474](https://github.com/cryspen/libcrux/pull/1474): Add support for AES-CCM from `libcrux-aes`, rename feature `aes_gcm` to `aes_aead`
 - [#1382](https://github.com/cryspen/libcrux/pull/1382): Add support for HMAC-DRBG
 - (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Integrate valgrind requests when cfg `valgrind_ct_test` is set
 - (libcrux-ml-dsa) [#1457](https://github.com/cryspen/libcrux/pull/1457): Add dependency on libcrux-secrets for optional valgrind integration
+- (libcrux-kmac) [#1292](https://github.com/cryspen/libcrux/pull/1292): Add support for kmac
+- (libcrux-sha3) [#1292](https://github.com/cryspen/libcrux/pull/1292): Add support for incremental CShake
 
 ## [0.0.4] (2026-05-13)
 
+### Added
```
jschneider-bensch: This should go to the `Unreleased` section.

crates/algorithms/sha3/CHANGELOG.md
```diff
@@ -24,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.0.8] (2026-03-19)
```
jschneider-bensch, line 30: This should go to the `Unreleased` section.

crates/algorithms/kmac/src/kmac.rs
```diff
@@ -0,0 +1,30 @@
...
+use internals::kmac::*;
+use libcrux_sha3::portable::incremental::{CShake128, CShake256};
+
+/// Write the KMAC128 MAC of `data` under `key` into `tag`, using
+/// `customization` if provided.
+///
+/// C.f. [NIST SP 800-185](https://csrc.nist.gov/pubs/sp/800/185/final),
+/// section 4.
+pub fn kmac_128<'a>(tag: &'a mut [u8], key: &[u8], data: &[u8], customization: &[u8]) -> &'a [u8] {
+    // Assert that key is long enough, i.e. at least 128 bits
```
jschneider-bensch, line 19: We have these comments, but we don't have the assertions?

crates/algorithms/sha3/src/portable/incremental.rs
```diff
@@ -0,0 +1,389 @@
+use super::*;
+use generic_keccak::xof::KeccakXofState;
+
+#[cfg(hax)]
+use crate::proof_utils::keccak_xof_state_inv;
+
+// Bring the proof-only invariant predicate into scope so its method can be
```
jschneider-bensch:
```suggestion
// Bring the proof-only invariant trait into scope so its method can be
```

crates/algorithms/sha3/src/portable/incremental.rs
```diff
@@ -0,0 +1,389 @@
...
+        state.absorb(&[customization_encoding]);
+        state.absorb(&[customization]);
+
+        // Pad zeros.
+        // `buffer_len` is only ever used modulo `RATE` (to compute `n_zeros`),
+        // so we reduce `name.len()` modulo `RATE` here. This keeps the sum
+        // small enough to provably never overflow `usize` while leaving
+        // `buffer_len % RATE` unchanged.
+        let buffer_len = 2
+            + name.len() % RATE
```
jschneider-bensch, line 362: This change is fine, since `(a + (b % c)) % c = (a + b - k * c) % c` for some `k` and that is equal to `(a + b) % c` since the the multiple of `c`  is dropped by the reduction.

jschneider-bensch (changes requested):
Looks good! Just some weirdness with the changelogs.

crates/algorithms/sha3/CHANGELOG.md
robinhundt: Oh yes, I didn't notice the existing entry. That's why I added one in the Unreleased section.

jschneider-bensch (approved):

PR #1294, opened 2026-01-21
Implement importing DH key material

georgio:
Hello,

Here are the proposed changes for the DH Key types.
It would be nice if we can validate that an imported public key is a valid point.

This could also be done externally.

Thanks for taking a look!


jschneider-bensch (approved):
This looks good, thank you!
I've added clamping to private key import and updated the changelog.

jschneider-bensch:
Sorry, seems I'm not quite awake yet this morning, will fix and then this can go in.

PR #1302, opened 2026-01-26
aesgcm: allow hax extraction

franziskuskiefer:
This allows extracting F* and Lean from the aes gcm code.

- [x] actually build avx2 code
- [x] figure out platform dependent extraction

[skip changelog]


crates/algorithms/aesgcm/src/platform/x64/gf128_core.rs
```diff
@@ -153,11 +153,14 @@ impl crate::platform::GF128FieldElement for FieldElement {
 #[test]
 fn test_transmute() {
     let x = 1u128 << 64 ^ 2u128;
-    let xv: __m128i = unsafe { core::mem::transmute(x) };
-    let xv: __m128i = unsafe { _mm_slli_si128(xv, 8) };
+    let xv: Vec128 = unsafe { core::mem::transmute(x) };
```
jschneider-bensch: Hm, should this test not use the `FieldElement::transmute` and `FieldElement::from_vec128` functions as well?

crates/algorithms/aesgcm/hax.py
```diff
@@ -0,0 +1,195 @@
...
+            cargo_hax_into,
+            cwd="../../utils/secrets",
+            env=hax_env,
+        )
+
+        # Extract ml-kem
+        includes = [
+            "+**",
+            "-libcrux_aesgcm::traits_api::**",
+            # "-libcrux_ml_kem::kem::**",
```
jschneider-bensch: These commented lines should be removed since they're unrelated here.

crates/algorithms/aesgcm/src/platform/x64/gf128_core.rs
franziskuskiefer: yeah, it's not really testing anything anyway. But I made it an actual test.

Issue #1220, opened 2025-11-02
Incorrect ML-DSA output on Alpine Linux and Ampere Altra


FiloSottile:
I tried using libcrux-ml-dsa v0.0.3 to generate [accumulated test vectors](https://words.filippo.io/accumulated/) for ML-DSA. Since I wanted to generate a 60M iteration test vector, I cross-compiled the code to run on an Alpine Linux server with an Ampere Altra processor.

Surprisingly, the output of the binary on Linux was different from the one generated on my M2 laptop. Comparing with my implementation, it looks like the local macOS build is correct, while the Alpine Linux cross-compiled build is producing incorrect results.

I built a minimal reproducer and built it directly on the Alpine Linux machine with a rustup toolchain, no cargo.toml, and Alpine's gcc to rule out cross-compilation issues. It gave the same incorrect output.

Note that I am a complete Rust noob, and I had Claude write most of the code, but from eyeballing it I can't imagine how it should produce different results on different targets.

### System 1 (macOS - CORRECT OUTPUT)
- **OS**: macOS 15.6.1 (24G90)
- **Kernel**: Darwin 24.6.0 ARM64
- **CPU**: Apple Silicon M2
- **Rust**: rustc 1.86.0 (05f9846f8 2025-03-31)
- **Cargo**: 1.86.0 (adf9b6ad1 2025-02-28)

### System 2 (Alpine Linux - INCORRECT OUTPUT)
- **OS**: Alpine Linux edge (3.23.0_alpha20250612)
- **Kernel**: Linux 6.12.40-1-lts aarch64
- **CPU**: Ampere Altra Q64-22 (ARM64)
- **Target**: aarch64-unknown-linux-musl
- **Rust**: rustc 1.91.0 (f8297e351 2025-10-28)
- **Cargo**: 1.91.0 (ea2d97820 2025-10-10)

### Cargo.toml
```toml
[package]
name = "libcrux-ml-dsa-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
libcrux-ml-dsa = "0.0.3"
hex = "0.4"
```

### src/main.rs
```rust
use libcrux_ml_dsa::ml_dsa_44;

fn main() {
    println!("libcrux-ml-dsa v0.0.3 - Platform Determinism Test\n");

    // Test with a zero seed
    let seed = [0u8; 32];
    let kp = ml_dsa_44::generate_key_pair(seed);

    // Sign an empty message with deterministic (zero) randomness
[...]

PR #1309, opened 2026-01-28
Document and check `no_std` support

jschneider-bensch:
This PR adds documentation and checks to some crates about their `no_std` support.

I used [`cargo-no-std`](https://github.com/wojciech-graj/cargo-no-std) to check most crates in the workspace for `no_std` support, and if given, under which conditions, i.e. do you need a global allocator, or do you need to disable a `std` default feature. The tool tries to build dummy binaries using the given crate for a `no_std` target, and optionally builds against a dummy allocator as well.

The main result is (unsurprisingly) that most crates I tested have `no_std` support, but many need an allocator because they depend on `libcrux-hacl-rs`. Other than that:

- For `libcrux-traits` I've included a small fix that feature-gates `extern crate alloc;` behind the existing `alloc` feature, as was surely the original intent. This allows using a subset of the traits without an allocator.
- There is some inconsistency around feature, e.g. some crates have a `std` default feature while others make `alloc/std` a non-default feature. Also I'm not sure we pay too much attention at the moment to depend on `no_std/no_alloc` feature sets.
- For some crates like `libcrux-ml-kem`, `libcrux-ml-dsa` and `libcrux-aesgcm` I got unexpected error messages from `cargo-no-std`, which need some investigation.

I've also added CI checks to the existing workflows of crates that I determined to have `no_std` support. They also just `cargo check` against a `no_std` target. This doesn't test the allocator requirement though.


crates/utils/intrinsics/README.md
```diff
@@ -0,0 +1,11 @@
+# SIMD Intrinsics
+
+![pre-verification]
```
franziskuskiefer: Here the question is if there's ever something we want to verify about this crate. Or if there's anything else we want to do about testing it.

traits/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+- [#XXX](https://github.com/cryspen/libcrux/pull/XXX): Allow `no_alloc` support & add minimal `README.md`
```
franziskuskiefer: Update PR number.

crates/utils/intrinsics/README.md
jschneider-bensch: Yes, I thought about avoiding issues like #1220, which was rooted in an incorrect fallback implementation. I think the most realistic thing would be a testing approach like what Aniket did with intrinsics.

jschneider-bensch: I've taken out the `pre-verification` badge, since anything we do is more likely to be test-shaped.

PR #1311, opened 2026-02-02
Reduce CI time

### LINKED ISSUE #1133: [CI] Only run actions on changed code

### LINKED ISSUE #1134: Don't run everything on main again (after it ran on the merge queue already)

jschneider-bensch:
This PR aims to reduce CI times by
- removing the `on.push` triggers for `main`, which cause workflows to be run on `main` after they've already been run on the merge queue.
- being more precise with `pull_request.path`, meaning a workflow for a given crate should only run if there were changes in its dependencies.

[skip changelog]

### DIFF
--- a/.github/workflows/aead-build_test.yml
+++ b/.github/workflows/aead-build_test.yml
@@ -2,9 +2,6 @@ name: AEAD - Build & Test
 
 on:
   merge_group:
-  push:
-    branches: ["main", "dev"]
-    paths: ['crates/primitives/aead', 'chacha20poly1305', 'traits/src/aead']
   pull_request:
     branches: ["main", "dev", "*"]
     paths: ['crates/primitives/aead', 'chacha20poly1305', 'traits/src/aead']
--- a/.github/workflows/aead-build_test.yml
+++ b/.github/workflows/aead-build_test.yml
@@ -4,7 +4,17 @@ on:
   merge_group:
   pull_request:
     branches: ["main", "dev", "*"]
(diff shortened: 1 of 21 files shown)


.github/workflows/aead-build_test.yml
```diff
@@ -2,12 +2,19 @@ name: AEAD - Build & Test
...
 on:
   merge_group:
-  push:
-    branches: ["main", "dev"]
-    paths: ['crates/primitives/aead', 'chacha20poly1305', 'traits/src/aead']
   pull_request:
     branches: ["main", "dev", "*"]
-    paths: ['crates/primitives/aead', 'chacha20poly1305', 'traits/src/aead']
+    paths:
+        - 'crates/primitives/aead/**'
```
franziskuskiefer, line 8: We should probably do this programmatically at some point (like discussed a little in #1133). But let's get this in for now to make things faster.

franziskuskiefer (approved):
Thanks!
This should get CI time down a lot.

.github/workflows/aead-build_test.yml
jschneider-bensch: Yes, I thought about that too, but found it quicker to do this manually for now. But especially when paths are changed, or dependencies come and go it would be nice to generate the paths on the fly.

jschneider-bensch:
Hm, this needs more work. The `-status` jobs required by the merge queue are not triggered with this path setup (as #1133 actually already described...).

jschneider-bensch:
Closing this, since the issue needs more thought.

PR #1259, opened 2025-12-02
clean libcrux crate

franziskuskiefer:
Draft PR for discussion of cleaned up libcrux crate.

Open todos
- [x] update benchmarks
- [x] add features
- [ ] update CI (e.g. drop specs workflow)

cc @keks @jschneider-bensch 

[skip changelog]

PR #1312, opened 2026-02-04
Fix hax compilation of crates relying on classification traits

maximebuyse:
The fix is to use `hax_lib::exclude` instead of `cfg(not(hax))` to exclude the unsupported items from hax extraction, without breaking compilation under the `hax` cfg. 

This PR replaces https://github.com/cryspen/libcrux/pull/1263

[skip changelog]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,26 +54,32 @@ allow-branch = ["main"]
 
 [workspace.dependencies]
 hax-lib = { version = "=0.3.6" }
-libcrux-intrinsics = { version = "=0.0.5", path = "crates/utils/intrinsics" }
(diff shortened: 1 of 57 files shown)


franziskuskiefer (approved):
lgtm. @jschneider-bensch do we have other crates where we should fix this? Also, should we add a dummy crate (or get #1259 in) to have a test in the repo to make sure we don't break this again?

jschneider-bensch:
> do we have other crates where we should fix this? Also, should we add a dummy crate (or get #1259 in) to have a test in the repo to make sure we don't break this again?

It is worth checking to avoid dependency trouble in the future, for sure. We should get #1259 as well.

PR #1313, opened 2026-02-04
move ml-dsa crate into crates/algorithms/

keks:
This is a step towards #1160


jschneider-bensch:
@keks Can you comment on the state here?

jschneider-bensch:
@keks I'm closing this because it's stale now, but please re-open if this is still active.

PR #1315, opened 2026-02-05
ECDSA signature malleability via lack of low-S normalization

nadimkobeissi:
libcrux's ECDSA P256 implementation does not enforce low-S normalization in either signing or verification. This allows signature malleability.

## Background

ECDSA signatures consist of two values `(r, s)` where both are integers modulo the curve order `n`. For any valid signature `(r, s)`, the signature `(r, n - s)` is also mathematically valid for the same message and public key. Low-S normalization ensures that s is always in the "lower half" of the range, i.e. `s <= n/2` This prevents signature malleability by ensuring only one of the two possible `(r, s)` pairs is considered valid.

The absence of low-S normalization is catastrophic in several critical and widespread applications. Bitcoin's BIP 62 and BIP 146 mandate low-S signatures because malleable signatures can alter transaction IDs, breaking transaction chains and enabling theft. Virtually all cryptocurrency and blockchain implementations require low-S normalization for the same reason—signature malleability can undermine transaction integrity, replay protection, and smart contract security. Beyond blockchain, any system using signatures as unique identifiers (deduplication systems, audit logs, identity verification) faces severe risks, as an attacker can generate alternate valid signatures to bypass security controls.

## Current state in libcrux

- **Signing**: does not produce low-S signatures. In `crates/algorithms/p256/src/p256.rs`, the function `ecdsa_sign_msg_as_qelem` (line 1923) only checks that r and s are non-zero at lines 1981-1984, but there is no code that compares `s` to `n/2` or replaces `s` with `(n - s)` if `s > n/2`.

- **Verification**: accepts high-S signatures. In the same file, the function `ecdsa_verify_msg_as_qelem` (line 1880) validates bounds at lines 1895-1896 to ensure `0 < r < n` and `0 < s < n`, but there is no check that `s <= n/2`.

## Fix included

This vulnerability report includes a proposed fix which takes a conservative approach, surfacing a `verify_strict` function that checks for low-S normalization in order not to break any potential backwards compatibility with insecure signatures. However, for actual signing, all signatures are normalized by adding a call to `normalize_s` in the internal `_sign` function.

---

[Music that inspired this work](https://www.youtube.com/watch?v=N7M_QU7mL_c)

### DIFF
--- a/crates/algorithms/ecdsa/src/p256.rs
+++ b/crates/algorithms/ecdsa/src/p256.rs
@@ -10,6 +10,66 @@ use crate::DigestAlgorithm;
 
 use super::Error;
 
+/// P-256 curve order n in big-endian bytes.
+/// n = 0xFFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551
+const P256_ORDER: [u8; 32] = [
+    0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x00, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+    0xBC, 0xE6, 0xFA, 0xAD, 0xA7, 0x17, 0x9E, 0x84, 0xF3, 0xB9, 0xCA, 0xC2, 0xFC, 0x63, 0x25, 0x51,
+];
+
+/// Half of P-256 curve order (n-1)/2 in big-endian bytes.
+/// Used for low-S normalization to prevent signature malleability.
+const P256_HALF_ORDER: [u8; 32] = [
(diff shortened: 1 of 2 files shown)


franziskuskiefer:
Thank you for your contribution. However, I am closing this Pull Request for the following reasons:

Policy Violation: This PR bypasses our documented Security Policy. For the safety of our users, all potential vulnerabilities must be reported through our established responsible disclosure channels rather than a public PR.

Irresponsible Disclosure: Publicly disclosing security-sensitive code before a patch is ready puts the community at risk.

Good Faith Concerns: Based on the nature of this submission, it does not appear to be a good-faith attempt to improve the project's security posture.

We take security seriously, but we require researchers to follow standard industry ethics and our repository's specific guidelines. Future submissions that disregard these protocols will be handled similarly.

franziskuskiefer:
> libcrux's ECDSA P256 implementation does not enforce low-S normalization in either signing or verification. This allows signature malleability.

Yes, this is how ECDSA is defined for P256. Adding other APIs is a feature request. This is not a bug.

PR #1316, opened 2026-02-05
Unnecessary "double clamping" in Ed25519 reduces seed entropy by 5 bits

nadimkobeissi:
The `generate_key_pair` function in `impl_hacl.rs` was applying clamping operations to the raw 32-byte seed before storing it:

```rust
sk[0] &= 248u8;
sk[31] &= 127u8;
sk[31] |= 64u8;
```

This clamping appears to be designed to prepare a scalar for Ed25519 curve operations, but it was being applied at the wrong stage. In standard Ed25519 ([RFC 8032](https://www.rfc-editor.org/rfc/rfc8032)):

1. The private key is a 32-byte random seed stored as-is.
2. To sign, compute `h = SHA-512`(seed).
3. The scalar `a = clamp(h[0:32])` is derived from the hash, not the seed (see page 54).
4. Clamping ensures: multiple of 8 (cofactor), `< 2^255`, and fixed high bit.

The libcrux implementation already performs correct clamping in `secret_expand()` after hashing the seed. The clamping in `generate_key_pair` was redundant and applied to the wrong value.

## Impact: seed entropy reduced by 5 bits for no benefit

The original clamping reduced the seed's entropy by 5 bits through three distinct operations. The first operation, masking the first byte with 0xF8, clears the three lowest bits, removing 3 bits of entropy. The second operation, masking the final byte with 0x7F, clears the highest bit, removing 1 bit of entropy. The third operation, setting bit 6 of the final byte, forces that bit to always be 1, removing another bit of entropy. Together, these operations constrain **what should be a 256-bit random value down to an effective entropy of 251 bits**.

While 251 bits of entropy remains likely cryptographically sufficient for practical purposes, this reduction serves no useful purpose in the context of Ed25519 key generation. A proper Ed25519 seed should contain the full 256 bits of entropy because the seed undergoes SHA-512 hashing before any scalar derivation takes place. The SHA-512 output is uniformly distributed regardless of any patterns in the input, and the proper clamping operations are designed to be applied after this hash, not before. The constrained seed is functionally equivalent to a different random seed that happens to have certain bit patterns fixed.

The fundamental issue is that these constraints provide no security benefit whatsoever while unnecessarily reducing the key space. The clamping operations make sense when applied to the hash output as part of scalar derivation, where they ensure the scalar is a multiple of 8 for the cofactor, is less than 2^255, and has a fixed high bit for consistent timing. Applying them to the raw seed before hashing accomplishes none of these goals and merely introduces an artificial limitation on the set of valid seeds.

## Proposed fix

Removed the clamping operations from generate_key_pair in `src/impl_hacl.rs`. The function now simply:

1. Generates a random 32-byte seed.
2. Rejects all-zero seeds.
3. Derives the public key via `secret_to_public` (which internally hashes and clamps).

This aligns with RFC 8032's key format where seeds are arbitrary 32-byte values.

---

[Music that inspired this work](https://www.youtube.com/watch?v=nIpajV0XctU)


nadimkobeissi:
Oh, wait, this merged the other PR somehow.

nadimkobeissi:
OK, pushed a `git revert` that removed the code that was meant strictly as part of a different PR (#1315)

franziskuskiefer:
Thank you for your contribution. However, I am closing this Pull Request for the following reasons:

Policy Violation: This PR bypasses our documented Security Policy. For the safety of our users, all potential vulnerabilities must be reported through our established responsible disclosure channels rather than a public PR.

Irresponsible Disclosure: Publicly disclosing security-sensitive code before a patch is ready puts the community at risk.

Good Faith Concerns: Based on the nature of this submission, it does not appear to be a good-faith attempt to improve the project's security posture.

We take security seriously, but we require researchers to follow standard industry ethics and our repository's specific guidelines. Future submissions that disregard these protocols will be handled similarly.

karthikbhargavan:
This PR reported a bug on Thursday, 5th Feb.
The bug was fixed on Monday, 9th Feb, in https://github.com/cryspen/libcrux/pull/1320
New crates with the fix (version 0.0.6) were published on Thursday, 12th Feb on crates.io.
A [security advisory](https://github.com/cryspen/libcrux/security/advisories/GHSA-435g-fcv3-8j26) noting bug-fixes in 0.0.6 was published on Friday, 13th Feb on GitHub,in accordance with a new advisory policy as documented in [SECURITY.md](https://github.com/cryspen/libcrux/blob/main/SECURITY.md)

PR #1379, opened 2026-04-02
Combined Header-Only C extraction for ML-KEM 768 & ML-DSA 65

### LINKED ISSUE #1240: Combined C Extraction of ML-KEM & ML-DSA

### LINKED ISSUE #1328: Update Eurydice for combined header-only C extraction of ML-KEM & ML-DSA
Would it be possible to make a `c_header_only` extract for ML-DSA, similar to the existing one for ML-KEM.

AFAIK the existing `libcrux-ml-dsa/cg` extract targets C++, but I'd like to use libcrux' ML-DSA support in OpenSSH, which is a vanilla C project.

jschneider-bensch:
[skip changelog]

Closes #1240 
Closes #1328


franziskuskiefer (approved):
Thanks! There's a bunch of duplication here of course. But that's fine for now.
Let's get this in and consolidate later, depending on how we want to maintain this kind of code.

PR #1382, opened 2026-04-09
Add Hmac_DRBG

### LINKED ISSUE #1396: Add HMAC_DRBG
Also note #36 for CTR DRBG

keks:
This PR adds a DRBG and a wrapper that allows it to implement rand::CryptoRng.


crates/algorithms/hmac-drbg/src/rand.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    /// Somewhat safely generates a bit or randomness:
+    ///  - ensures we don't pass in too much additional_input (none, in fact)
+    ///  - reseeds if needed
+    ///
+    /// The caller MUST ensure that they don't ask fo too much data. Ideally enforce this using
+    /// hax.
+    ///
+    ///
+    //hax: requires dst.len() < crate::MAX_GENERATE_BYTES
```
franziskuskiefer, line 329: I played with some of these comments but we may want to drop them or translate them into some other style.

crates/algorithms/hmac/src/hmac.rs
```diff
@@ -20,8 +20,37 @@ pub mod hacl {
 }
 
 mod impl_hacl;
+mod incremental;
```
franziskuskiefer, line 23: We should update the docs around this new, not yet verified version.

crates/algorithms/drbg/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "libcrux-drbg"
+
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+license.workspace = true
+readme.workspace = true
```
franziskuskiefer: We should add a readme here.

Cargo.toml
```diff
@@ -123,6 +126,9 @@ libcrux-kats = { path = "crates/testing/kats" }
 libcrux-macros = { version = "=0.0.3", path = "crates/utils/macros" }
 libcrux-secrets = { version = "=0.0.5", path = "crates/utils/secrets" }
 
+# 3rd Party
+rand = { version = "0.9", default-features = false }
```
franziskuskiefer, line 130: We may bump to 0.10 before merging this and need to update here then as well.

Cargo.toml
```diff
@@ -169,6 +176,7 @@ mlkem = []
 mldsa = []
 rsapss = []
 poly1305 = []
+drbg = []
```
jschneider-bensch: This feature doesn't do anything in the main crate yet, it looks like?

crates/testing/kats/src/wycheproof/hmac.rs
```diff
@@ -0,0 +1,67 @@
+//! Wycheproof HMAC Known Answer Tests
+//!
+//! The JSON files were taken from <https://github.com/C2SP/wycheproof>, as of commit
+//! [cd136e97040de0842c3a198670b1c5e4f423c940](https://github.com/C2SP/wycheproof/tree/cd136e97040de0842c3a198670b1c5e4f423c940).
```
jschneider-bensch, line 4: Is the referenced commit here correct? It's the same as ML-KEM and ML-DSA, but perhaps it's just copy paste?

crates/algorithms/hmac-drbg/src/rand.rs
jschneider-bensch: Where can I find more information on these comments?

crates/algorithms/hmac-drbg/src/errors.rs
```diff
@@ -0,0 +1,213 @@
...
+    #[cfg(feature = "health-tests")]
+    HealthCheckFailed,
+}
+
+/// Errors returned by [`HmacDrbg::generate`].
+///
+/// [`HmacDrbg::generate`]: crate::HmacDrbg::generate
+#[derive(Debug, PartialEq)]
+pub enum GenerateError {
+    /// The reseed counter has exceeded [`RESEED_INTERVAL`]; call `reseed` before generating.
```
jschneider-bensch, line 86:
```suggestion
    /// The reseed counter has exceeded [`RESEED_INTERVAL`](crate::RESEED_INTERVAL); call `reseed` before generating.
```

crates/algorithms/hmac-drbg/src/errors.rs
```diff
@@ -0,0 +1,213 @@
...
+    ReseedRequired,
+
+    /// The requested output length is zero or exceeds [`MAX_GENERATE_BYTES`].
```
jschneider-bensch, line 89:
```suggestion
    /// The requested output length is zero or exceeds [`MAX_GENERATE_BYTES`](crate::MAX_GENERATE_BYTES).
```

crates/algorithms/hmac-drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+    ///
+    /// `additional_input` is optional; pass `None` or `Some(&[])` for none.
+    /// Returns `Err(ReseedRequired)` if the reseed counter has been exhausted.
+    ///
+    /// With feature `health-tests`, four continuous tests run on every call:
+    /// - **Intra-call** (NIST 800-90C §9.3.3): `V = HMAC(K, V_old)` must satisfy
+    ///   `V ≠ V_old`; equality signals a catastrophic HMAC fixed-point.
+    /// - **Inter-call / T4** (AIS 20/31): each new block must differ from the last
+    ///   block of the previous `generate` call.
+    /// - **RCT** (SP 800-90B §4.4.1): no run of ≥ [`RCT_C`] identical bytes.
```
jschneider-bensch, line 275: The doc links here are broken if the `health-tests` feature is not enabled.

crates/algorithms/drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+//! use libcrux_drbg::HmacDrbgSha256;
+//!
+//! // In production obtain entropy from the OS (e.g. `new_from_sys_rng`).
+//! let entropy = {
+//!     let mut e = [0u8; 32];
+//!     for (i, b) in e.iter_mut().enumerate() { *b = if i % 2 == 0 { 0x55 } else { 0xaa }; }
+//!     e
+//! };
+//! let nonce = [0u8; 16];
+//! let mut drbg = HmacDrbgSha256::new(&entropy, &nonce, &[]).unwrap();
```
jschneider-bensch: I think it would be nice for the example to use the personalization and also the optional extra input below, with a comment stating the these can also be omitted. Otherwise there are unexplained empty parameters in the example, which could be confusing.

jschneider-bensch (dismissed):
Just some initial comments on docs, will give it a more thorough review later.

crates/algorithms/hmac-drbg/src/rand.rs
franziskuskiefer: It's just `hax: ...` with a precondition on lengths. Not much more to it.

crates/algorithms/drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+mod utils;
+
+mod hmac;
+use hmac::{HmacAlgorithm, HmacSha256, HmacSha384, HmacSha512};
+
+/// Maximum number of generate calls before a reseed is required (SP 800-90A §10.1 Table 2).
+///
+/// After this many calls to [`HmacDrbg::generate`], the next call returns
+/// [`GenerateError::ReseedRequired`]. Use [`HmacDrbg::needs_reseed`] to check proactively.
+pub const RESEED_INTERVAL: u64 = 10_000;
```
jschneider-bensch: Isn't this 2^48 according to the referenced table?

crates/algorithms/hmac-drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+/// [`GenerateError::ReseedRequired`]. Use [`HmacDrbg::needs_reseed`] to check proactively.
+pub const RESEED_INTERVAL: u64 = 10_000;
+
+/// Maximum number of bytes that can be requested in a single [`HmacDrbg::generate`] call
+/// (SP 800-90A §10.1 Table 2).
+///
+/// Requesting more returns [`GenerateError::RequestTooLarge`].
+pub const MAX_GENERATE_BYTES: usize = 65_536;
+
+/// Minimum entropy input length in bytes (SP 800-90A Table 2: security strength).
```
jschneider-bensch, line 75: Should we reference Table 1 here as well? That is where the justification is that a maximum designed security strength of 256 bits allows for instantiating all the lower security strengths as well. I found it a bit confusing since the concrete number stated here is not found in Table 2, which instead references another recommendation document for the security strength.

crates/algorithms/drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    /// HMAC_DRBG_Update (§10.1.2.2).
+    ///
+    /// `provided_parts` is a slice of data fragments that are logically
+    /// concatenated to form the provided_data argument.  Pass `&[]` when
+    /// there is no additional data (equivalent to `provided_data = Null`
+    /// in the spec).
+    fn update(&mut self, provided_parts: &[&[u8]]) -> Result<(), UpdateError> {
+        let has_data = provided_parts.iter().any(|s| !s.is_empty());
+        let mut new_key = [0u8; OUTLEN];
```
jschneider-bensch: I think `new_key` isn't strictly necessary, since we could just finalize directly into `self.key` instead.

crates/algorithms/drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    // -----------------------------------------------------------------------
+    // §10.1.2.3  Instantiate
+    // -----------------------------------------------------------------------
+
+    /// Instantiate an HMAC_DRBG from explicit entropy material (SP 800-90A §10.1.2.3).
+    ///
+    /// - `entropy_input`: must contain at least [`MIN_ENTROPY_BYTES`] bytes of entropy.
+    /// - `nonce`: at least `OUTLEN / 2` bytes (the spec minimum); using `OUTLEN` bytes is
```
jschneider-bensch: It looks like we don't check for this requirement on the nonce.

crates/algorithms/hmac-drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+        Ok(())
+    }
+
+    // -----------------------------------------------------------------------
+    // §10.1.2.3  Instantiate
+    // -----------------------------------------------------------------------
+
+    /// Instantiate an HMAC_DRBG from explicit entropy material (SP 800-90A §10.1.2.3).
+    ///
+    /// - `entropy_input`: must contain at least [`MIN_ENTROPY_BYTES`] bytes of entropy.
```
jschneider-bensch, line 184: We do check the length of the provided entropy with the `health-tests` feature, but should we maybe do it in any case since it's relatively inexpensive?

I see that we don't have Error variants for these checks conditions failing outside of the `health-tests` feature...

crates/algorithms/hmac-drbg/src/errors.rs
```diff
@@ -0,0 +1,213 @@
...
+    RequestTooLarge,
```
jschneider-bensch, line 90: I find the naming here a bit confusing since it includes the zero-length case. Maybe something like this?
```suggestion
    RequestInvalid,
```

crates/algorithms/hmac-drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+        Ok(())
+    }
+
+    // -----------------------------------------------------------------------
+    // §10.1.2.5  Generate
+    // -----------------------------------------------------------------------
+
+    /// Generate pseudorandom bytes into `output` (§10.1.2.5).
+    ///
+    /// `additional_input` is optional; pass `None` or `Some(&[])` for none.
```
jschneider-bensch, line 267: Perhaps I'm missing something, but wouldn't it simplify things to accept `additional_input` as a `&[u8]` and perform the first update `self.update(additional_input)` only when `!additional_input.is_empty()`?

The final update can then be an unconditional `self.update(additional_input)`.

crates/algorithms/drbg/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+    //
+    // #hax: requires personalization_string.len() <= MAX_SEED_BYTES - 2 * OUTLEN
+    // #hax: ensures result.is_ok() ==> result.unwrap().reseed_counter == 1
+    pub fn new_from_sys_rng(
+        personalization_string: &[u8],
+    ) -> Result<Self, InstantiateFromRngError> {
+        Self::new_from_rng(&mut rand::rngs::SysRng, personalization_string)
+    }
+
+    /// Reseed from a [`rand::CryptoRng`] (generates entropy internally).
```
jschneider-bensch: Is there a reason why instantiation from RNG accepts a fallible RNG, and reseeding does not?

crates/algorithms/drbg/src/rand.rs
```diff
@@ -0,0 +1,379 @@
...
+///
+/// [`SeedableRng::from_seed`] intentionally bypasses the AIS 31 startup
+/// health tests that [`HmacDrbg::new`] applies to the entropy input.
+/// `SeedableRng` is a *deterministic* interface: the caller supplies an exact
+/// seed, which may include all-zero test vectors.  The continuous health tests
+/// that run during [`HmacDrbg::generate`] are still active.
+///
+/// ## `from_rng` / `try_from_rng` and startup health tests
+///
+/// Both overrides draw `OUTLEN` bytes of entropy **and** `OUTLEN` bytes of
```
jschneider-bensch: Typo?
```suggestion
/// Both methods draw `OUTLEN` bytes of entropy **and** `OUTLEN` bytes of
```

crates/algorithms/drbg/src/rand.rs
```diff
@@ -0,0 +1,379 @@
...
+            Ok(drbg) => drbg,
+            Err(error) => match error {
+                crate::InstantiateError::InputTooLarge => unreachable!(),
+            },
+        };
+
+        Self { drbg, rng }
+    }
+
+    /// Somewhat safely generates a bit or randomness:
```
jschneider-bensch: So this is safe if the caller observes the request limit?

```suggestion
    /// Somewhat safely generates a bit of randomness:
```

crates/algorithms/drbg/src/rand.rs
```diff
@@ -0,0 +1,379 @@
...
+    ///  - ensures we don't pass in too much additional_input (none, in fact)
+    ///  - reseeds if needed
+    ///
+    /// The caller MUST ensure that they don't ask fo too much data. Ideally enforce this using
```
jschneider-bensch:
```suggestion
    /// The caller MUST ensure that they don't ask for too much data. Ideally enforce this using
```

crates/algorithms/hmac-drbg/src/rand.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    fn try_next_u64(&mut self) -> Result<u64, Self::Error> {
+        let mut out = [0u8; 8];
+        self.safe_generate_small(&mut out);
+        Ok(u64::from_le_bytes(out))
+    }
+
+    fn try_fill_bytes(&mut self, dst: &mut [u8]) -> Result<(), Self::Error> {
+        let (chunks, rest): (&mut [[u8; MAX_GENERATE_BYTES]], _) = dst.as_chunks_mut();
```
jschneider-bensch, line 373: I think this assumes that that we can always generate `MAX_GENERATE_BYTES` without reseeding, which isn't the case if the `HmacDrbgRng` has been called before with `try_next_u32` or `try_next_u64`, right?

Meaning calling any of those before will advance the reseed counter, but not to the point where the `need_reseed` check in `safe_generate_small` is triggered.

crates/algorithms/hmac-drbg/src/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+    fn drop(&mut self) {
+        // Overwrite key material with zeros. Using write_volatile prevents the
+        // compiler from optimising away the writes as "dead stores".
+        for b in self.key.iter_mut() {
+            unsafe { core::ptr::write_volatile(b, 0) };
+        }
+        for b in self.v.iter_mut() {
+            unsafe { core::ptr::write_volatile(b, 0) };
+        }
+        atomic::compiler_fence(atomic::Ordering::SeqCst);
```
jschneider-bensch, line 30: For my understanding: What's the purpose of this `atomic::compiler_fence` here?

crates/algorithms/hmac-drbg/src/health_tests.rs
```diff
@@ -0,0 +1,261 @@
...
+            unsafe { core::ptr::write_volatile(b, 0) };
+        }
+        unsafe { core::ptr::write_volatile(&mut self.rct_last_byte, 0) };
+        unsafe { core::ptr::write_volatile(&mut self.apt_first_byte, 0) };
+        atomic::compiler_fence(atomic::Ordering::SeqCst);
+    }
+
+    /// Block-level tests run on each generated V block.
+    ///
+    /// - **Test 1** (NIST 800-90C §9.3.3): `new_block ≠ v_before` (HMAC fixed-point).
```
jschneider-bensch, line 133: Is that reference correct? I can't find section 9.3.3 in [NIST SP-800-90C](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-90C.pdf).

crates/algorithms/hmac/src/hmac.rs
jschneider-bensch: Should include a changelog entry, too.

jschneider-bensch (dismissed):
Still didn't look too closely at the health checks yet, but have enough comments on the rest.

crates/algorithms/hmac-drbg/src/rand.rs
keks: When I saw this, I thought the tooling changed while I was away, so I adopted it :sweat_smile: I can also change it to hax attributes, if that is better. I don't know the reasoning behind the new syntax (but it is similar to how things work in Go: https://go.dev/wiki/Comments#directives

keks:
How do you all feel about renaming the crate and module to hmac_drbg? It seems DRBG would be the scheme/primitive, but the construction we put into the algorithms folder is HMAC_DRBG.

jschneider-bensch:
> How do you all feel about renaming the crate and module to hmac_drbg? It seems DRBG would be the scheme/primitive, but the construction we put into the algorithms folder is HMAC_DRBG.

That sounds reasonable :+1:

crates/algorithms/drbg/src/lib.rs
keks: I double checked the document, the comment is wrong. The wording is:

The nonce shall be either:
a. A value with at least (security_strength/2) bits of entropy, or
b. A value that is expected to repeat no more often than a (security_strength/2)-bit random string would be expected to repeat.

So if you only do the operation once, even no nonce is fine, because it doesn't repeat (at least that is my reading). I replaced it with saying it needs to be unique and a note explaining the technicalities.

crates/algorithms/hmac-drbg/src/lib.rs
keks: I was thinking about this as well. This is how it's done in the standard and it was given to me like this by @franziskuskiefer, so I decided against changing it. I liked that the health check error was so opaque and didn't want to open it up. Maybe we can add the check to only run if we _don't_ do the health checks? Then we only need the additional error variant in that case, and the error variants don't overlap.

We'll have to do the same changes in Reseed.

crates/algorithms/hmac-drbg/src/lib.rs
keks: Hm. I am not sure. On the one hand we have [this passing test][test] that checks that we get the same for  `Some(&[])` and `None`. That would suggest the distinction doesn't matter. On the other hand we do [update v in the update function][update v].

I tried your suggestion and it produces wrong results. I'll push it to a separate branch.

[test]: https://github.com/cryspen/libcrux/pull/1382/files#diff-60e05d4ca548b45bf666f71d275f2245a9a6f616a5132640dc3a7796137dd53cR318-R325
[update v]: https://github.com/cryspen/libcrux/pull/1382/files#diff-74227025daae59eaa9ad25a1f015e5fc6b88cdce048eb03d155e28a2b1482607R120

crates/algorithms/drbg/src/rand.rs
keks: Ah, this is because the SeedableRng trait requires implementing only one function and the others have a default impl that we are overriding here. Making this more explicit.

crates/algorithms/drbg/src/rand.rs
keks: Yes. This function ensures that we auto-reseed if needed and that we don't have too much input data for hmac.

crates/algorithms/hmac-drbg/src/rand.rs
keks: My understanding as that every invocation of `generate` can generate

`max_number_of_bits_per_request` = $2^{19}$ bits, i.e. $2^{16}$ bytes.

crates/algorithms/hmac-drbg/src/utils.rs
keks: It tells the compiler to not move a read before the zeroization when it reorders for optimization.

I think.

crates/algorithms/hmac-drbg/src/rand.rs
jschneider-bensch: Oops, my bad! I confused `MAX_GENERATE_BYTES` somehow with `RESEED_INTERVAL` and thought it was a cross call limit.

crates/testing/kats/src/wycheproof/hmac.rs
keks: @franziskuskiefer do you remember what commit you took them from?

keks: nevermind, since that commit there only was one, which formatted the files. These files are not formatted, so I assume the hash is correct

crates/algorithms/hmac-drbg/src/health_tests.rs
keks: It also is not in SP-800-90Ar1, or in the Health Test section of that document. Not sure where that came from.

keks: @franziskuskiefer I think you added this. Do you remember?

crates/algorithms/hmac-drbg/src/lib.rs
keks: https://github.com/cryspen/libcrux/tree/keks/drbg-additional-input

crates/algorithms/hmac-drbg/src/utils.rs
jschneider-bensch: I'm not sure about the reasoning here. From the documentation for [`std::sync::atomic::compiler_fence`](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html) I understand that it applies to intra-thread synchronization only and for this to work the read/write operations should be atomic ones. But as far as I can tell we don't do any atomic reads. The [`write_volatile`](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html) is guaranteed to not to be re-ordered wrt to other externally observable events but can not be considered atomic.

@franziskuskiefer What are the requirements for zeroization here?

crates/algorithms/hmac-drbg/src/lib.rs
jschneider-bensch: The reason for the failing test is that the [first update](https://github.com/cryspen/libcrux/blob/8d9d5789c8c0b796684c5d160fc807c34af05ba5/crates/algorithms/drbg/src/lib.rs#L286) with the additional input should only be done when `additional_input` is not empty.

If you change that to
```rust
        if !additional_input.is_empty() {
            self.update(&[additional_input])?;
        }
```
the tests pass.

crates/algorithms/hmac-drbg/src/rand.rs
jschneider-bensch: I think it's fine to keep them around until we have a better idea of the style of proof annotations that we'll use in the future. I did not review any of these comments, though.

crates/algorithms/hmac/src/hmac.rs
jschneider-bensch: The changelog entry should be under an `### Added` heading. (We started doing a bit more fine-grained tracking of the change categories at some point, but didn't adapt do the categorization for earlier versions retroactively.)

crates/algorithms/drbg/Cargo.toml
jschneider-bensch: We can fill in more information in a follow-up.

Cargo.toml
jschneider-bensch: I've updated the dependabot PR for this here: #1385

crates/algorithms/hmac-drbg/src/health_tests.rs
franziskuskiefer: Oh, that looks hallucinated. 800-90A §10.1.2 may be more appropriate here.

crates/algorithms/hmac-drbg/src/lib.rs
franziskuskiefer: yeah sounds reasonable to me.

crates/algorithms/hmac-drbg/src/utils.rs
franziskuskiefer: This is stupid zeroization. I'd be totally fine with dropping this here, or moving into a separate crate to allow zerization across crates at some point.

crates/algorithms/hmac-drbg/src/health_tests.rs
keks: That's the thing. I can't find the check anywhere in the document.Not in 10.1.2 and not in 11.3.

jschneider-bensch:
@keks Are the remaining comments all related to health tests? If so, we could also address those in a follow up.

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+
+impl<const OUTLEN: usize, const BLOCK: usize, Digest> HmacState<OUTLEN>
+    for HmacStream<OUTLEN, BLOCK, Digest>
+where
+    Digest: arrayref::DigestIncremental<OUTLEN>,
+    Digest::IncrementalState: InitializeDigestState,
+{
+    /// Initialize a new streaming HMAC computation with the given key.
+    fn new(key: &[u8]) -> Result<Self, Error> {
+        if key.len() > u32::MAX as usize {
```
jschneider-bensch, line 30: Please correct me if I'm wrong here, but this limit is specific to the `Digest` implementations for SHA2, right? I think for SHA3 the same limit would apply, and I don't know if we are ever likely to instantiate with anything else but we should still document that this is a limit that comes from the specific implementations here. (I don't think we have a way at the moment to get the maximum supported payload length for a `Digest` impl?).

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+        if key.len() <= BLOCK {
+            key_block[..key.len()].copy_from_slice(key);
+        } else {
+            // Key longer than one block: hash it down to OUTLEN bytes.
+            let mut tmp = Digest::IncrementalState::new();
+            Digest::update(&mut tmp, key)
+                .expect("This can't fail because we checked that key.len() <= u32::MAX");
+            let mut hashed = [0u8; OUTLEN];
+            Digest::finish(&mut tmp, &mut hashed);
+            key_block[..OUTLEN].copy_from_slice(&hashed);
```
jschneider-bensch, line 44: Do we assert somewhere that `OUTLEN <= BLOCK`?

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut opad = [0u8; BLOCK];
+
+        for i in 0..BLOCK {
+            ipad[i] = key_block[i] ^ 0x36;
+            opad[i] = key_block[i] ^ 0x5c;
+        }
+
+        // Prime the inner state with ipad so subsequent updates feed data directly.
+        let mut inner = Digest::IncrementalState::new();
+        Digest::update(&mut inner, &ipad).expect("HMAC: ipad update failed");
```
jschneider-bensch: Can we say anything about why this shouldn't panic? For our instantiations it's clear that, e.g. `BLOCK < u32::MAX`, but we also don't check that. In practice here I don't think it matters, but `Digest` does not give a good way of checking beforehand that errors will be avoided. (Perhaps worth a follow-up issue to overhaul `Digest` at some point.)
Alternatively, we could propagate the error here.

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        let mut ipad = [0u8; BLOCK];
+        let mut opad = [0u8; BLOCK];
+
+        for i in 0..BLOCK {
+            ipad[i] = key_block[i] ^ 0x36;
+            opad[i] = key_block[i] ^ 0x5c;
+        }
+
+        // Prime the inner state with ipad so subsequent updates feed data directly.
```
jschneider-bensch:
```suggestion
        // Prime the inner state with K0 xor ipad so subsequent updates feed data directly.
```

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+/// and an incremental digest implementation `D`.
+///
+/// Use the type aliases [`HmacSha256`], [`HmacSha384`], or [`HmacSha512`]
+/// rather than spelling out the type parameters directly.
+pub struct HmacStream<const OUTLEN: usize, const BLOCK: usize, D>
+where
+    D: arrayref::DigestIncremental<OUTLEN>,
+    D::IncrementalState: InitializeDigestState,
+{
+    /// Inner hash state already primed with ipad.
```
jschneider-bensch:
```suggestion
    /// Inner hash state already primed with K0 xor ipad.
```

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        Ok(Self { inner, opad })
+    }
+
+    fn update(&mut self, data: &[u8]) -> Result<(), Error> {
+        Digest::update(&mut self.inner, data).map_err(|_| Error::InvalidInputLength)
+    }
+
+    fn finalize(mut self, dst: &mut [u8; OUTLEN]) {
+        // Inner hash: finalize SHA-2(ipad || data).
```
jschneider-bensch: Comments here could not specific to SHA2
```suggestion
        // Inner hash: finalize H(K0 xor ipad || data).
```

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut inner_hash = [0u8; OUTLEN];
+        Digest::finish(&mut self.inner, &mut inner_hash);
+
+        // Outer hash: SHA-2(opad || inner_hash).
```
jschneider-bensch:
```suggestion
        // Outer hash: H(K0 xor opad || inner_hash).
```

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+        let mut outer = Digest::IncrementalState::new();
+        Digest::update(&mut outer, &self.opad).expect("self.opad.len() is always <= u32::MAX");
+        Digest::update(&mut outer, &inner_hash).expect("inner_hash.len() is always <= u32::MAX");
```
jschneider-bensch, line 79: Again, a technicality, but I don't think we check that `OUT_LEN <= u32::MAX`.

crates/algorithms/hmac/src/incremental.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// Streaming HMAC-SHA-384 state.  Initialize with [`HmacSha384::new`].
+pub type HmacSha384 = HmacStream<48, 128, libcrux_sha2::Sha384Hash>;
+
+/// Streaming HMAC-SHA-512 state.  Initialize with [`HmacSha512::new`].
+pub type HmacSha512 = HmacStream<64, 128, libcrux_sha2::Sha512Hash>;
+
+/// Compute HMAC-SHA-256 over logically concatenated, non-contiguous data slices.
+#[inline]
+pub fn hmac_sha2_256_slices(dst: &mut [u8; 32], key: &[u8], slices: &[&[u8]]) -> Result<(), Error> {
```
jschneider-bensch: Could you implement these generically as e.g. `hmac_slices` for `HmacStream<OUTLEN, BLOCK, Digest> : HmacState<OUTLEN>`? That could potentially simplify things on the DRBG side too?

crates/testing/kats/src/wycheproof/hmac/schema.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// Test result.
+    pub result: MacResult,
+
+    /// A list of flags.
+    pub flags: Vec<Flag>,
+}
+
+#[derive(PartialEq, Serialize, Deserialize, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum MacResult {
```
jschneider-bensch: This could be `schema_common::TestResult` instead, now.

crates/algorithms/hmac/tests/wycheproof.rs
```diff
@@ -0,0 +1,76 @@
...
+            "tcId={} should be valid",
+            t.tc_id
+        ),
+        MacResult::Invalid => assert_ne!(
+            computed,
+            t.tag.as_slice(),
+            "tcId={} should be invalid",
+            t.tc_id
+        ),
+        MacResult::Acceptable => {} // skip
```
jschneider-bensch: We should make sure we're running some tests here, e.g. by counting the number of Valid/Invalid assertion cases and checking that it's not 0.

crates/algorithms/hmac-drbg/fuzz/fuzz_targets/generate.rs
```diff
@@ -0,0 +1,61 @@
...
+//! Input layout:
+//!   [0]      algorithm selector: 0 → SHA-256, 1 → SHA-384, 2 → SHA-512
+//!   [1..3]   output_len (little-endian u16; the full range 0..=65537 is explored)
+//!   [4]      has_additional_input: 0 = None, non-zero = Some(rest of data)
+//!   [5..]    additional_input bytes (used when has_additional_input != 0)
+//!
+//! Checks:
+//!   - No panic or undefined behaviour for any input.
+//!   - `RequestTooLarge` iff output_len == 0 or output_len > MAX_GENERATE_BYTES.
+//!   - On success the output buffer is fully written (no uninitialised bytes leak).
```
jschneider-bensch, line 15: How is this checked below?

crates/algorithms/hmac-drbg/src/utils.rs
jschneider-bensch: I think we can leave this, but I've created #1450 to review it again in the broader context of zeroization for the whole library.

jschneider-bensch (approved):
Thanks for addressing the comments! I think I have only some take-it-or-leave-it suggestions for the incremental HMAC API now.

I've also created some follow-up issues:
- For running the tests on CI: #1451, #1452
- For revisiting zeroization: #1450 
- For revisiting health-tests: #1441

crates/algorithms/hmac-drbg/src/health_tests.rs
jschneider-bensch: Follow-up: #1441

crates/algorithms/hmac/src/incremental.rs
keks: No, but (so far) that holds for our implementations of the trait. Not sure where to best assert this, and not sure if this can be enforced by the type system (I didn't find a way that is not basically using S(S(S(Z))) style numbers)

crates/algorithms/hmac/src/incremental.rs
keks: Yes, I think that is correct. I suppose we could add the maximum supported length as a constant on the Digest trait. Then we could do that check properly. Not sure it's worth it, though

crates/algorithms/hmac/src/incremental.rs
keks: I am adding a debug_assert here.

crates/algorithms/hmac/src/incremental.rs
keks: I am adding another debug_assert that `BLOCK <= u32::MAX`.

crates/algorithms/hmac/src/incremental.rs
keks: hm, the variable `ipad` here is alread `K0 xor 0x36`. I am updating the variable name.

crates/algorithms/hmac-drbg/fuzz/fuzz_targets/generate.rs
keks: - the fuzzer catches panics.
- not sure about UB, but I think this is all safe rust, so there shouldn't be any? Maybe we can just remove that?
- fixed RequestTooLarge to RequestInvalid - this is checked in the Err case of the match in the run macro.
- Buffer written: yeah, not sure.

jschneider-bensch: The buffer written one is the one I was wondering about. Maybe just remove it, since I don't think that is something that's checked here.

jschneider-bensch (approved):
Thanks!

crates/algorithms/hmac/src/incremental.rs
jschneider-bensch: Right, I'd say that would be for a follow-up in any case!

jschneider-bensch:
I think it looks good to merge now. I'll resolve the conflicts.

PR #1386, opened 2026-04-14
[chacha20poly1305] Fix potential encryption panic on overlong ciphertext buffer

jschneider-bensch:
This PR fixes an issue in `libcrux_chacha20poly1305::encrypt` where an application that passes in a ciphertext buffer of length greater than `ptxt.len() + TAG_LEN` would experience a panic.
The fix makes it so that `libcrux_chacha20poly1305::encrypt` no longer panics in this case, but instead writes out the ciphertext and tag into the first `ptxt.len() + TAG_LEN` bytes of the provided buffer.

The issue was reported by @fg0x0.

### DIFF
--- a/crates/algorithms/chacha20poly1305/CHANGELOG.md
+++ b/crates/algorithms/chacha20poly1305/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- [#1386](https://github.com/cryspen/libcrux/pull/1386): Fix potential panic in `libcrux_chacha20poly1305::encrypt (reported by @fg0x0)
+- [#1386](https://github.com/cryspen/libcrux/pull/1386): Fix potential panic in `libcrux_chacha20poly1305::encrypt` (reported by @fg0x0)
 
 ## [0.0.7] (2026-03-19)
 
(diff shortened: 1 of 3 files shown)


crates/algorithms/chacha20poly1305/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- [#1386](https://github.com/cryspen/libcrux/pull/1386): Fix potential panic in `libcrux_chacha20poly1305::encrypt (reported by @fg0x0)
```
franziskuskiefer:
```suggestion
- [#1386](https://github.com/cryspen/libcrux/pull/1386): Fix potential panic in `libcrux_chacha20poly1305::encrypt` (reported by @fg0x0)
```

crates/algorithms/chacha20poly1305/tests/chachapoly.rs
```diff
@@ -309,6 +309,10 @@ fn chachapoly_test_invalid_buffer_lengths() {
     let mut ctxt = [0; 29];
     libcrux_chacha20poly1305::encrypt(&key, msg, &mut ctxt, aad, &nonce).unwrap();
 
+    // encrypt correctly on ciphertext buffer longer than `ptxt.len() + TAG_LEN`
+    let mut long_ctxt = [0; 30];
+    libcrux_chacha20poly1305::encrypt(&key, msg, &mut long_ctxt, aad, &nonce).unwrap();
```
franziskuskiefer, line 314: Shall we check that is worked correctly?

franziskuskiefer (approved):
Thanks!

crates/algorithms/chacha20poly1305/tests/chachapoly.rs
jschneider-bensch: Sure! At that point the caller will have to know what originally encrypted plaintext length was though.

PR #1332, opened 2026-02-14
feat(ml-kem): add zeroize support and tests

rainmitch:
### Description:
This PR introduces memory hardening for ML-KEM secret keys by implementing the Zeroize and ZeroizeOnDrop traits. The goal is to ensure sensitive key material is securely wiped from memory when it goes out of scope. This facilitates FIPS 140-3 compliance (AS05.10) regarding the zeroization of plaintext secret keys.

## Scope:
This implementation focuses on the secret-holding structs and their underlying vector types. It does not modify the underlying core logic or seed inputs (which very likely needs more work to fully implement zeroize). I believe this covers all primary secret-holding structs in a non-intrusive way required for memory hardening.

This is implemented as an optional feature that requires being enabled.

## Changes:
I have implemented Zeroize for the following types:
- High-Level Keys: MlKemPrivateKey, MlKemKeyPair
- Internal Components: PolynomialRingElement
- Unpacked Variants: MlKemPrivateKeyUnpacked, IndCpaPrivateKeyUnpacked, MlKemKeyPairUnpacked
- Vector Backends: PortableVector, SIMD128Vector, SIMD256Vector

I have implemented ZeroizeOnDrop for the following types:
- MlKemPrivateKey, MlKemKeyPair
- PolynomialRingElement
- MlKemPrivateKeyUnpacked, IndCpaPrivateKeyUnpacked, MlKemKeyPairUnpacked

> Note: Vector types do not implement ZeroizeOnDrop to avoid performance penalties in non-secret contexts; they are zeroed explicitly by their parent structs.

## Technical Verification:
To ensure the security features are not optimized away by the compiler, I performed the following checks:
- Hax Compatibility: All zeroize logic is gated behind #[cfg(not(hax))] to prevent interference with the formal verification pipeline.
- Dead Store Elimination (DSE): Used core::hint::black_box in SIMD implementations and testing to force the compiler to treat the memory as "used," preventing it from deleting the zeroing instructions.

Assembly Audit:
- SIMD128Vector test: Verified that stp (Store Pair) instructions are generated for clearing, and vmaxvq is used for zero-verification in tests.
- SIMD256Vector test: Verified that vmovdqa is generated for clearing, and vptest is used for zero-verification.

## Performance:
Benchmarks show a negligible performance impact (performed on a Ryzen 5 7430u):

Operation | Key Size | Platform | API | Ratio | Zeroize-Enabled | Original
-- | -- | -- | -- | -- | -- | --
[...]

PR #1387, opened 2026-04-22
[ml-dsa]Implement Zeroize for MLDSASigningKey/[traits,secrets,chacha20poly1305]Implement Zeroize for AEAD Key

### LINKED ISSUE #1370: Feature request: implement Zeroize for MLDSASigningKey (and related key types)
I'm building a peer-to-peer encrypted file transfer tool over LAN(for now). It uses a custom
handshake protocol (X25519 + ML-KEM-1024 hybrid key exchange, ML-DSA-87 for peer
identity signatures) no X.509, no PKIX, no certificate authorities. I'm using
`libcrux-ml-dsa` as the signing implementation.

## What I ran into

My signing code looks roughly like this:

```rust
pub fn sign(&self, message: &[u8]) -> Result<[u8; SIGNATURE_LEN]> {
    // self.signing_key is a Box<[u8; 4896]> that I zeroize on Drop.
    let typed_sk = libcrux_ml_dsa::MLDSASigningKey::new(*self.signing_key); // <- 4896-byte stack copy

    let mut randomness = Zeroizing::new([0u8; 32]);
    crate::crypto::csprng().fill_bytes(&mut *randomness);

    let sig = ml_dsa_87::sign(&typed_sk, message, ctx, *randomness)?;
    // typed_sk drops here but without a Zeroize impl
    // sit on the stack until the next frame overwrites them.
    Ok(*sig.as_ref())
}
```

I zeroize my own `Box<[u8; 4896]>` on drop via the `zeroize` crate, so the
long-lived copy is taken care of. But `MLDSASigningKey` doesn't implement `Zeroize`,
which means the short-lived stack copy that gets created for the duration of the
`sign()` call isn't explicitly cleared when it drops. It'll eventually be overwritten
by normal stack reuse, but there's no guarantee on when.

## Why I care

RFC 9881 §9 specifically calls out side-channel risk in ML-DSA signing:

> *"Some amount of side-channel leakage has been demonstrated in parts of the signing
> algorithm (specifically the bit-unpacking function), from which a demonstration of
> key recovery has been made over a large sample of signatures."*

It also recommends randomized signing to harden against fault and hardware side-channel
attacks and I'm already doing that (hedged signing with a 256-bit RBG string). But the
complementary defense and then not leaving key material in memory longer than necessary is
something I can't currently achieve without `Zeroize` support on `MLDSASigningKey`.

In practice the window is tiny: it's the time between `drop(typed_sk)` and the stack
frame getting reused. I'm not claiming this is an imminent catastrophe. But in a
security-sensitive context, "tiny window" isn't the same as "no window", and I'd like
to be able to document that I've done everything possible on my side.

## What would help me

Would it be feasible to add `Zeroize` (and ideally `ZeroizeOnDrop`) for
`MLDSASigningKey` and the equivalent types for ML-DSA-44 and ML-DSA-65 while
we're at it?

Something like:

```rust
impl zeroize::Zeroize for MLDSASigningKey {
    fn zeroize(&mut self) {
        self.0.zeroize(); // assuming the inner type is a plain byte array
    }
}

impl zeroize::ZeroizeOnDrop for MLDSASigningKey {}
```

I just don't want to touch the internals blindly without knowing whether there are constraints I'm
not aware of (inner representation, etc.).

## A few questions

- Is this something that's already on your radar, or intentionally left out for a
  reason I'm missing?
- Does the formally verified internals story complicate adding a `Zeroize` impl, or
  is it straightforward at the boundary?
- Is there anything I should know for this context?

## Environment

- `libcrux-ml-dsa`: 0.0.8 (feature mldsa87)
- `libcrux-secrets`: 0.0.5
- Rust: 1.85.1

Thanks for your time and again.

### LINKED ISSUE #1375: Feature Request: implement Zeroize for Key in chacha20poly1305 (and related key types)
The same concern I raised for `libcrux-ml-dsa` https://github.com/cryspen/libcrux/issues/1370 applies here.

## What I ran into

My encrypt/decrypt code looks roughly like this:

```rust
let derived_key = derive_key_from_password(password, &salt)?; // Zeroizing<[u8; 32]>

let aead_key = Key::from((*derived_key).classify()); // <- 32-byte copy into Key
aead_key.encrypt(&mut ctxt_buf, &mut tag, &nonce, &[], plaintext.classify_ref())?;
// aead_key drops here but without Zeroize, the 32 bytes may linger
```

I wrap my Argon2id output in `Zeroizing<[u8; 32]>`, so the long-lived copy is taken care of. But `Key` doesn't implement `Zeroize`, which means the copy that lives inside the `Key` struct for the duration of the encrypt/decrypt call isn't explicitly cleared on drop.

## Why I care

This key is derived from a user password and used to protect a long-lived ML-DSA-87 identity keypair (4896-byte signing key + 2592-byte verification key).  The derived key itself is short-lived it exists only for one encrypt or decrypt call but:

Without `Zeroize` on `Key`, the 32-byte derived key copy lives in memory for an unspecified duration after the call returns. My own `Zeroizing<[u8; 32]>` takes care of the long-lived copy, but the copy inside `Key` is outside my control.

## What would help

A `Zeroize` (and ideally `ZeroizeOnDrop`) implementation for `Key`.

Something like:

```rust
impl zeroize::Zeroize for Key {
    fn zeroize(&mut self) {
        self.0.zeroize(); // assuming inner representation is a plain byte array
    }
}
impl zeroize::ZeroizeOnDrop for Key {}
```

## Environment

- `libcrux-aead`: 0.0.7 (feature: chacha20poly1305)
- `libcrux-secrets`: 0.0.5
- Rust: 1.85.1

Thanks for your time.

MavenRain:
Add an optional `zeroize` feature implementing Zeroize, Drop, and
  ZeroizeOnDrop for MLDSASigningKey<SIZE> and MLDSAKeyPair.  MLDSAKeyPair
  clears only the signing key; the verification key is public material.

  Gated on `not(hax)` so hax extraction is unaffected.

  Closes #1370.

  Add an optional `zeroize` feature:

  - libcrux-secrets: `Zeroize for Secret<T> where T: Zeroize`, so `[U8; N]` zeroizes under both default (U8 = u8) and `check-secret-independence` (U8 = Secret<u8>) configurations.
  - libcrux-traits: `Zeroize for typed_owned::Key<Algo>` where `Algo::Key: Zeroize`.  Covers `chacha20poly1305::Key` and any future AEAD whose key bytes are Zeroize.
  - libcrux-chacha20poly1305: `zeroize` feature forwarding to both.

  Drop and ZeroizeOnDrop are not implemented in this PR.  Adding them
  requires `Algo::Key: Zeroize` as a bound on `typed_owned::Aead` (or a
  sealed helper trait), which is a larger trait-level change.  Happy to
  follow up if maintainers want one of those approaches.  In the meantime,
  callers wanting drop-on-zero can wrap in `zeroize::Zeroizing<Key<...>>`.

  Closes #1375.


jschneider-bensch:
Thanks! I will review when I have capacity for it.
I'll say upfront (as mentioned in #1332) that I'm a bit apprehensive about adding new cargo features and I'm thinking about different ways to manage array-backed key types (cf. #1390).

MavenRain:
I feel like this is still relevant, but open to feedback.

jschneider-bensch:
This is still on hold since we haven't decided yet on a particular approach to zeroization for the workspace.

PR #1389, opened 2026-04-22
Fix 1362 xof streaming squeeze

MavenRain:


jschneider-bensch:
Thank you! I don't have capacity for a full review right now.
But I noticed that you also updated the hax annotations. Did you run it through hax already?

MavenRain:
> Thank you! I don't have capacity for a full review right now. But I noticed that you also updated the hax annotations. Did you run it through hax already?

Sorry for the delay . . . I was behind on my Github email notifications.

I installed hax at the workspace's pinned `hax-lib-v0.3.6` revision and ran `cargo hax into ... fstar` on `crates/algorithms/sha3` locally.  The extract succeeds, and all three of my annotation changes translate cleanly into the resulting `Libcrux_sha3.Generic_keccak.Xof.fst`:                            
                                                                             
  - `t_KeccakXofState` gains `f_squeeze_buf` and `f_squeeze_pos`.             
  - `impl__new`'s ensures clause carries `result.f_squeeze_pos =. v_RATE`.    
  - `impl_1__squeeze` carries `self.f_squeeze_pos <=. v_RATE` on requires and `future(self).f_squeeze_pos <=. v_RATE` on ensures; the loop's bounds assert shifts to `out_offset + i * v_RATE <=. out.len()` as expected.
                                                                              
The `SHA3 - hax` workflow on this PR is sitting in `action_required` (fork, no prior merged PR).  Approving it should exercise lax + type-check on top of what I confirmed locally.  I'd be happy to run `prove --admit` here too if you'd rather, it just needs F* v2025.12.15, which I haven't set up yet.

robinhundt:
Thank you for the PR and sorry that the review is taking so long!

The implementation looks good and fixes the issue with partial squeezes. The F* proofs were broken and largely fixed by me in the latest commits. I merged the main branch into this one to resolve some conflicts. Unfortunately, the pre-/post-condition changes also broke the newly added cshake implementation proofs (#1292). I'll fix these up next week.

robinhundt (approved):
Thank you again for the PR!

Due to the kmac and cshake changes I merged into this branch, I had to do a bit of additional work to get the hax extraction and F* proof to go through. 

@jschneider-bensch I would be great if you could also review this, as I made some more substantive changes to the proofs (but hopefully made some more legible rather than less). I put the commit for updating the c extractions at the end, so that you can filter them out in the review UI.

crates/algorithms/sha3/tests/portable.rs
jschneider-bensch, line 1: The later tests are only for SHAKE-128, looks like. We could make sure that we have all tests for SHAKE-128 and SHAKE-256.

crates/algorithms/sha3/src/generic_keccak/xof.rs
jschneider-bensch, line 1: Could we re-use `buf[0]` and `buf_len` for tracking the squeeze buffering ? Meaning `buf_len` is set to `RATE` in `absorb_final` and now serves the role of `squeeze_pos` with `buf[0]` used instead of `squeeze_buf` for buffering. I don't know if the compiler would do that on its own already, perhaps, but it could reduce some of the bookkeeping for `cfg(hax)`?

jschneider-bensch (commented):
I think the squeeze logic looks good, but I wonder if we can reduce some bookkeeping by re-using the parts of the state that are used for absorb buffering.

crates/algorithms/sha3/tests/portable.rs
robinhundt: I've added variants of the tests for SHAKE-256. The code with the bug was generic over the rate, which is why I omitted the other tests for shake256. But it doesn't hurt to have them, so I've added them.

crates/algorithms/sha3/src/generic_keccak/xof.rs
robinhundt: Mh, interesting idea. I think that should work.

But I'm wondering whether we should rather get rid of the buffering completely, for absorption and squeezeing (see #1501), than to only partially remove it. But that should not be part of this PR I think.

jschneider-bensch: Right, I would be fine to keep it like this for the PR now, and address #1501 later. :+1:

crates/algorithms/sha3/tests/portable.rs
jschneider-bensch: Thanks!

PR #1391, opened 2026-04-23
sha2: remove alloc requirement

franziskuskiefer:
- drop hmac-sha1


jschneider-bensch (changes requested):
The changes all look good to me, but
- CI shows some crates still reference `hmac_sha1`
- dropping HMAC-SHA 1 should have a changelog entry
- Documentation should be updated with the new `no_std` status
- CI could be updated with a build for a no-alloc target so we maintain that status

jschneider-bensch (approved):

PR #1392, opened 2026-04-23
tracing: fix warning

franziskuskiefer:
[skip changelog]


jschneider-bensch (approved):
Ah, yes, thanks!
That was made into a warning relatively recently I believe.

PR #1395, opened 2026-04-27
[ML-DSA] Fully reduce iNTT inputs on AVX2

jschneider-bensch:
The reduction of inputs to the inverse NTT that was introduced in https://github.com/cryspen/libcrux/pull/1006 was only performed for four of 32 SIMD vectors for AVX2. This is incongruent with the portable implementation and the test introduced in #1006 for portable produces a different value after inverse NTT than on AVX2.

I have not found this to be an issue in any of the KAT tests, but I can't give an estimation of how likely or unlikely it is to result in differences between portable and AVX2 for real signatures.


jschneider-bensch:
I've updated the tests for this edge case on portable to check that unreduced inputs do produce panics (in debug mode) or disagree with reduced inputs (release mode).
I've also added tests that AVX2 and portable agree on the results for iNTT of these edge cases.

franziskuskiefer (approved):
Thanks, lgtm

libcrux-ml-dsa/src/simd/avx2.rs
```diff
@@ -158,9 +158,8 @@ impl Operations for AVX2SIMDUnit {
 
     #[inline(always)]
     fn reduce(simd_units: &mut [Self; SIMD_UNITS_IN_RING_ELEMENT]) {
-        shift_left_then_reduce::<0>(&mut simd_units[0].value);
-        shift_left_then_reduce::<0>(&mut simd_units[8].value);
-        shift_left_then_reduce::<0>(&mut simd_units[16].value);
-        shift_left_then_reduce::<0>(&mut simd_units[24].value);
+        for i in 0..simd_units.len() {
+            shift_left_then_reduce::<0>(&mut simd_units[i].value);
```
karthikbhargavan: It is unclear why we call "shift_left_then_reduce::<0>" here. Wouldn't it be cleaner to have a "reduce" function at the level of simd_units that you can call directly? Then shift_left_the_reduce could also use this function internally.

karthikbhargavan: In this function, we could factor out the reduction (everything except the first line) into a "reduce" function, which could then be used from both shift_then_reduce and the direct reduce: https://github.com/cryspen/libcrux/blob/abf60f5c7774a2f8d1738be333c57e9bb0a00f35/libcrux-ml-dsa/src/simd/avx2/arithmetic.rs#L142

karthikbhargavan (dismissed):

libcrux-ml-dsa/src/simd/avx2.rs
jschneider-bensch: I factored out a function that does a Barrett reduction at the SIMD unit level. In portable, this means that there are now two loops for `shift_then_reduce`, but I would hope the compiler can do that in one pass instead.

karthikbhargavan (approved):

PR #1397, opened 2026-04-29
chore: extract anonymous implementations of `{Public,Private}Key` traits

cfm:
Marks these traits for extraction per <https://hax.cryspen.com/manual/faq/include-flags/#6-including-anonymous-items-using-hax_libinclude>.  But just let me know if there's a better way to make these available for extraction by crates that use them.  :-)

### DIFF
--- a/libcrux-kem/src/kem.rs
+++ b/libcrux-kem/src/kem.rs
@@ -398,6 +398,7 @@ pub enum Ss {
     MlKem1024(MlKemSharedSecret),
 }
 
+#[hax_lib::include]
 impl PrivateKey {
     /// Encode a private key.
     pub fn encode(&self) -> Vec<u8> {
@@ -449,6 +450,7 @@ impl PrivateKey {
     }
 }
 
+#[hax_lib::include]
 impl PublicKey {
     /// Encapsulate a shared secret to the provided `pk` and return the `(Key, Enc)` tuple.
     pub fn encapsulate(&self, rng: &mut impl CryptoRng) -> Result<(Ss, Ct), Error> {


maximebuyse:
@cfm So you will extract `libcrux-kem` on your side to verify a crate that depends on it, and you need only those 2 impls, is that correct? 

Did you try extracting more than just these two items? A `-i` flag would allow you to select the whole module for example. If this is a problem for extraction or F* type-checking, would extracting with interface-only work for you?

cfm:
Thanks, @maximebuyse.  We've been starting by extracting as narrowly as possible, which has worked well for us except for anonymous items like these trait implementations.  But you're right that these do extract fine with:

```sh-session
$ cargo hax into -i '-** +!libcrux_kem::*' fstar --interfaces '+!**'
```

However, it looks like the F* modules this produces depend on cryspen/hax#1899, which hasn't yet been released:
- https://github.com/freedomofpress/securedrop-protocol/actions/runs/25196467077/job/73878101627#step:6:366

```
[CHECK] Libcrux_kem.fsti 
Error: (118) * Error 118 at ../models/Libcrux_kem.fsti(255,4-255,18):
  - No field 'f_try_into_pre' in record type 'Core_models.Convert.t_TryInto'.
```

I'll see if I can wire this up directly from cryspen/hax@c2f85199b049caa1d9eaf8e7953bb84d356bac0d, and I'll follow up with you if I run into trouble that way.

maximebuyse:
> Thanks, @maximebuyse. We've been starting by extracting as narrowly as possible, which has worked well for us except for anonymous items like these trait implementations. But you're right that these do extract fine with:
> 
> ```
> $ cargo hax into -i '-** +!libcrux_kem::*' fstar --interfaces '+!**'
> ```
> 
> However, it looks like the F* modules this produces depend on [cryspen/hax#1899](https://github.com/cryspen/hax/pull/1899), which hasn't yet been released:
> 
>     * https://github.com/freedomofpress/securedrop-protocol/actions/runs/25196467077/job/73878101627#step:6:366
> 
> 
> ```
> [CHECK] Libcrux_kem.fsti 
> Error: (118) * Error 118 at ../models/Libcrux_kem.fsti(255,4-255,18):
>   - No field 'f_try_into_pre' in record type 'Core_models.Convert.t_TryInto'.
> ```
> 
> I'll see if I can wire this up directly from [cryspen/hax@c2f8519](https://github.com/cryspen/hax/commit/c2f85199b049caa1d9eaf8e7953bb84d356bac0d), and I'll follow up with you if I run into trouble that way.

@cfm Ok, we can make a release soon. If there is anything else you want for this next release I can also make sure we include it.

PR #1398, opened 2026-05-05
fix(ml-dsa): avx2 use_hint function

robinhundt:
This PR updates the wycheproof ML-DSA verification test vectors and fixes an issue uncovered by the updated vectors:

The updated vectors include signatures which result in a `use_hint(1, 0)` call. As per the spec, `r0 <= 0`, should return `(r1 - 1) mod m`.   However, the AVX2 implementation of use_hint used the `vec256_blendv_epi32` intrinsic to check whether `r0 < 0`, and otherwise return `(r1 + 1) mod m`.

In the case that r0 = 0, this resulted in differing behavior between the spec/portable implementation and the AVX2 backend.

This fixes the issue by first comparing r0 against 0 and using the result as the mask for vec256_blendv_epi32. The fix is adapted from https://github.com/pq-code-package/mldsa-native/blob/45ba4b3e87aba0e6681f256a3e5f90e01b0e3af1/dev/x86_64/src/poly_use_hint_32_avx2.c#L71-L77


jschneider-bensch (dismissed):
Thank you! The only thing that's missing is a short changelog entry under the "Fixed" heading, e.g. like [here](https://github.com/cryspen/libcrux/blob/jonas/intt-avx2-fix/libcrux-ml-dsa/CHANGELOG.md#fixed).

jschneider-bensch (approved):
Looks good, thanks!

PR #1399, opened 2026-05-06
Correctness Proofs for SHA-3 (Part 1): A Rust Specification for SHA-3

karthikbhargavan:
This PR adds a specification of SHA-3 in a purely functional subset of Rust (sometimes called hacspec). 

The spec uses some specific proof-friendly styles.
1) For code that creates or updates arrays point-wise, it uses createi (array::from_fn) as a way of creating arrays using an initialization function instead of initializing the array using a loop.
2) For code that loops over unbounded inputs, it prefers recursion over loops
3) When reasoning about integer arithmetic, it sometimes uses hax_lib::int

The spec itself verifies in F* (`hax.sh extract; hax.sh prove`) and is tested against the typical test vectors.
We also added cross-spec tests to the SHA-3 implementation in libcrux, testing that the code matches the spec.
As a drive-by change, we aligned the array layout in the implementation state to match that of the specification state.
This eases all subsequent proofs.

[skip changelog]


abentkamp:
I ran `hax into aeneas-lean` on these new Rust specs. Here is what I found:

One issue with extraction: Aeneas says `Unimplemented` for `xor_block_into_state` in `sponge.rs` (https://github.com/AeneasVerif/aeneas/issues/924). An easy workaround is to use `core::array::from_fn` directly instead of `createi`.

The extracted Lean code cannot be compiled yet, mostly because of missing defs in core models: 
    - Insts.Core_modelsOpsIndexIndexMut.index_mut
    - Slice.Insts.Core_modelsOpsIndexIndex.index
    - Array.Insts.Core_modelsOpsIndexIndex.index

There is also one slightly more serious issue with Lean compilation: https://github.com/AeneasVerif/aeneas/issues/984
This can probably be worked around by editing the Lean file (e.g., with a python script).

So overall, I think these specs are suitable for `hax into aeneas-lean`, even though we might have to make some small corrections in the future. I'll now proceed to read the code and review it.

specs/sha3/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+    r#"
+let createi
+      (#v_T: Type0)
+      (v_N: usize)
+      (#v_F: Type0)
+      (f: (x:usize{x <. v_N}) -> v_T)
+    : t_Array v_T v_N
+    = Rust_primitives.Arrays.createi v_N f
+"#
+)]
```
abentkamp, line 21: We need to use `createi` instead of using `array::from_fn` directly for F*, right?

Maybe we could call it `array_from_fn` instead of `createi`? That would be more self-explanatory.

specs/sha3/src/sponge.rs
```diff
@@ -0,0 +1,165 @@
...
+#[hax_lib::requires(rate > 0 && rate <= 200 && rate % 8 == 0)]
+#[hax_lib::decreases(message.len().to_int())]
+pub fn absorb_rec(state: State, rate: usize, delim: u8, message: &[u8]) -> State {
+    if message.len() < rate {
+        absorb_final(state, message, 0, message.len(), rate, delim)
+    } else {
+        let state = absorb_block(state, &message[0..rate], rate);
+        absorb_rec(state, rate, delim, &message[rate..])
+    }
+}
```
abentkamp, line 104: In this form, a recursive function doesn't seem to be more useful for verification than a for-loop.

It would help if the recursive function would only call `absorb_block`, and leave the `absorb_final` call to be implemented in the `absorb` function below. Then we could use `absorb_rec` to write down a loop invariant on the implementation's for-loop, I think.

crates/algorithms/sha3/src/generic_keccak.rs
```diff
@@ -308,3 +308,368 @@ impl<const N: usize, T: KeccakItem<N>> Index<(usize, usize)> for KeccakState<N,
...
+    }
+
+    #[test]
+    fn store_block_matches_spec() {
+        let state = test_state();
+        let mut impl_out = [0u8; 200];
+        crate::simd::portable::store_block::<136>(&state, &mut impl_out, 0, 136);
+        let mut spec_out = [0u8; 200];
+        spec_out = spec_sponge::squeeze_state(&state, spec_out, 0, 136);
+        assert_eq!(impl_out[..136], spec_out[..136]);
```
abentkamp: We can even assert `assert_eq!(impl_out, spec_out);`, can't we?

crates/algorithms/sha3/src/generic_keccak.rs
```diff
@@ -308,3 +308,368 @@ impl<const N: usize, T: KeccakItem<N>> Index<(usize, usize)> for KeccakState<N,
...
+        for &len in &[8, 16, 64, 72, 104, 136] {
+            let mut impl_out = [0u8; 200];
+            crate::simd::portable::store_block::<136>(&state, &mut impl_out, 0, len);
+            let mut spec_out = [0u8; 200];
+            spec_out = spec_sponge::squeeze_state(&state, spec_out, 0, len);
+            assert_eq!(
+                impl_out[..len],
+                spec_out[..len],
+                "store_block mismatch at len={len}"
+            );
```
abentkamp: Again, can't we test equality of the entire array?

crates/algorithms/sha3/src/generic_keccak.rs
```diff
@@ -308,3 +308,368 @@ impl<const N: usize, T: KeccakItem<N>> Index<(usize, usize)> for KeccakState<N,
...
+        let len = 72;
+        let offset = 64;
+        let mut impl_out = [0u8; 200];
+        crate::simd::portable::store_block::<136>(&state, &mut impl_out, offset, len);
+        let mut spec_out = [0u8; 200];
+        spec_out = spec_sponge::squeeze_state(&state, spec_out, offset, len);
+        assert_eq!(
+            impl_out[offset..offset + len],
+            spec_out[offset..offset + len]
+        );
```
abentkamp: Here, too.

specs/sha3/tests/cavp.rs
```diff
@@ -0,0 +1,281 @@
...
+                msg: msg.clone(),
+                output,
+            });
+        }
+    }
+    (input_length_bits, tests)
+}
+
+// ---------------------------------------------------------------------------
+// Path to the test vector files (shared with reference implementation)
```
abentkamp: When I read "reference implementation", I'd think of the hacspec version. So maybe better:
```suggestion
// Path to the test vector files (shared with implementation)
```

abentkamp:
Other than the small nitpicks above, this looks good! I haven't compared the hacspec implementation with the official FIPS spec, though. Let me know if I should do that.

specs/sha3/NIST.FIPS.202.pdf
jschneider-bensch, line 1: Is it necessary to have this in the repository?

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
```
jschneider-bensch, line 21: This doesn't need to be part of the changelog for the libcrux crate, since we don't expose the sha3 spec via that crate.

specs/sha3/.gitignore
jschneider-bensch, line 1: This file is not necessary, since the spec is part of the `specs` workspace, the cargo output directory would be `specs/target`, and that is already in the `.gitignore` for the workspace.

specs/sha3/hax_aeneas.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+
+import subprocess
+import re
+import sys
+from pathlib import Path
+
+import os
+
+HAX_VERSION = "1f85fc13b9967080cc657863e2000ba5d4aa8647"
```
jschneider-bensch, line 10: Could you add a `VERIFICATION.md` or a `README.md` with a section explaining how to obtain these verification dependencies? I.e. are these revisions of main branch hax and main branch aeneas, or some fork?

specs/sha3/hax.sh
```diff
@@ -0,0 +1,182 @@
...
+function cli() {
+    if [ -z "$1" ]; then
+        help
+        exit 1
+    fi
+
+    case "$1" in
+        --help) #> Show help message
+            help;;
+        extract) #> Extract the F* code for the proofs.
```
jschneider-bensch:
```suggestion
        extract_fstar) #> Extract the F* code for the proofs.
```

specs/sha3/hax.sh
```diff
@@ -0,0 +1,182 @@
...
+            extract_all
```
jschneider-bensch:
```suggestion
            extract_all_fstar
```

specs/sha3/hax.sh
```diff
@@ -0,0 +1,182 @@
...
+            msg "$GREEN" "done"
+            ;;
+        extract_lean) #> Extract Lean code for the proofs.
+            extract_all_lean
+            msg "$GREEN" "done"
+            ;;
+        prove) #> Run F*. This typechecks the extracted code. To lax-typecheck use --admit.
```
jschneider-bensch:
```suggestion
        prove_fstar) #> Run F*. This typechecks the extracted code. To lax-typecheck use --admit.
```

specs/sha3/hax.sh
```diff
@@ -0,0 +1,182 @@
...
+
+function extract_all_lean() {
+    extract_to_lean specs/sha3 \
+        into -i "+**" \
+        lean
+
+    patch_lean_extractions
+}
+
+function prove() {
```
jschneider-bensch:
```suggestion
function prove_fstar() {
```

specs/sha3/hax.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
+set -ex
+
+function extract_all() {
```
jschneider-bensch:
```suggestion
function extract_all_fstar() {
```

specs/sha3/hax.sh
```diff
@@ -0,0 +1,182 @@
...
+    extract specs/sha3 \
```
jschneider-bensch:
```suggestion
    extract_to_fstar specs/sha3 \
```

specs/sha3/hax.sh
```diff
@@ -0,0 +1,182 @@
...
+    ROOT="$SCRIPT_DIR/../.."
+    cd "$ROOT"
+    cd "$1"
+}
+
+function msg() {
+    echo -e "$1[$SCRIPT_NAME]$RESET $2"
+}
+
+function extract() {
```
jschneider-bensch:
```suggestion
function extract_to_fstar() {
```

specs/sha3/hax.sh
jschneider-bensch, line 1: This script does a different lean extraction than `hax_aeneas.py`, right? Do we need the lean extraction here, or can this script be purely to drive the F* extraction?

specs/sha3/tests/cavp.rs
jschneider-bensch, line 1: This file should be updated to use the `libcrux-kats` crate instead of manually parsing test vectors from hardcoded file paths. The current version of `/crates/algorithms/sha3/tests/cavp.rs` in on the main branch of libcrux can be the basis.

specs/sha3/proofs/fstar/extraction/Hacspec_sha3.fst
jschneider-bensch, line 1: Should we commit the extracted files to the repo?

specs/sha3/proofs/aeneas-lean/HacspecSha3/Extraction/Funs.lean
jschneider-bensch, line 1: This is the only extracted file for lean, right? Should we put it in `.gitignore`?

specs/sha3/proofs/aeneas-lean/HacspecSha3.lean
jschneider-bensch, line 1: Would this be the entrypoint for any theorems?

specs/sha3/src/lib.rs
jschneider-bensch: I would also prefer a more descriptive name, like the one suggested by @abentkamp, as well as a short comment explaining why we can't use `array::from_fn` directly.

specs/sha3/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+#[cfg(not(hax_backend_lean))] // https://github.com/AeneasVerif/aeneas/issues/924
```
jschneider-bensch: What is the relevance of the issue that is referenced here? I saw that in some other place you should use `array::from_fn` directly, because `createi` causes an Aeneas issue, but that's not clear from the comment here.

specs/sha3/src/keccak_f.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Read lane `A[x, y]`.
+#[inline]
+#[hax_lib::requires(x < 5 && y < 5)]
+pub fn get(state: &State, x: usize, y: usize) -> u64 {
+    state[5 * y + x]
+}
+
+// =========================================================================
+// Constants — FIPS 202, Section 3.3 / Algorithm 5
```
jschneider-bensch:
```suggestion
// Constants — FIPS 202, Section 3.2
```

specs/sha3/src/keccak_f.rs
```diff
@@ -0,0 +1,156 @@
...
+    0x8000_0000_8000_000A,
+    0x8000_0000_8000_8081,
+    0x8000_0000_0000_8080,
+    0x0000_0000_8000_0001,
+    0x8000_0000_8000_8008,
+];
+
+/// Rotation offsets for ρ step — FIPS 202, Algorithm 2 / Table 2.
+///
+/// Indexed as `RHO_OFFSETS[5*y + x]`.
```
jschneider-bensch:
```suggestion
/// Indexed as `RHO_OFFSETS[5*y + x]`. Already reduced by the lane size (64 bits).
```

specs/sha3/src/sponge.rs
```diff
@@ -0,0 +1,165 @@
...
+    let bytes: [u8; 200] = createi(|i| state[i / 8].to_le_bytes()[i % 8]);
+    output[out_offset..out_offset + len].copy_from_slice(&bytes[0..len]);
+    output
+}
+
+/// Absorb one full block: XOR it into the state, then apply Keccak-f.
+///
+/// Corresponds to one iteration of the absorb loop in Algorithm 8 (step 6).
+#[hax_lib::requires(rate <= 200 && rate % 8 == 0 && block.len() == rate)]
+pub fn absorb_block(state: State, block: &[u8], rate: usize) -> State {
```
jschneider-bensch, line 48: Perhaps we can add a `debug_assert_eq!(block.len(), rate)` here as well.

specs/sha3/src/sponge.rs
```diff
@@ -0,0 +1,165 @@
...
+use crate::createi;
+use crate::keccak_f::{keccak_f, State};
+
+#[cfg(hax)]
+use hax_lib::int::*;
+
+/// XOR a block of message bytes into the state (little-endian, lane-interleaved).
+///
+/// Corresponds to the `S ⊕ (Pi || 0^c)` step of Algorithm 8.
+#[hax_lib::requires(rate <= 200 && rate % 8 == 0 && block.len() >= rate)]
```
jschneider-bensch, line 16: I think it would be worth it to have a `debug_assert` for this precondition as well.

specs/sha3/src/sponge.rs
```diff
@@ -0,0 +1,165 @@
...
+pub fn xor_block_into_state(state: State, block: &[u8], rate: usize) -> State {
+    createi(|i| {
+        if i < rate / 8 {
+            // The slice is exactly 8 bytes (since `i < rate / 8` and
+            // `block.len() >= rate`), so `try_into::<[u8; 8]>` cannot fail.
+            state[i] ^ u64::from_le_bytes(block[8 * i..8 * i + 8].try_into().unwrap())
```
jschneider-bensch, line 22: This `unwrap` should be an `expect` with the content of the comment above it.

specs/sha3/src/sponge.rs
```diff
@@ -0,0 +1,165 @@
...
+    buffer[remaining] = delim;
+    buffer[rate - 1] = buffer[rate - 1] | 0x80;
+    buffer
+}
+
+/// Absorb the final (possibly partial) block: pad it, XOR into state, and
+/// apply Keccak-f.
+///
+/// Combines `pad_last_block` + `absorb_block`.
+#[hax_lib::requires(rate > 0 && rate <= 200 && rate % 8 == 0 && remaining < rate
```
jschneider-bensch, line 77: Is it necessary for the proofs or the cross-implementation testing to have a extra parameters for `msg_offset` and `remaining`?

specs/sha3/src/sponge.rs
```diff
@@ -0,0 +1,165 @@
...
+                     && msg_offset <= message.len() && remaining <= message.len() - msg_offset)]
+pub fn pad_last_block(
+    message: &[u8],
+    msg_offset: usize,
+    remaining: usize,
+    rate: usize,
+    delim: u8,
+) -> [u8; 200] {
+    let mut buffer = [0u8; 200];
+    buffer[0..remaining].copy_from_slice(&message[msg_offset..msg_offset + remaining]);
```
jschneider-bensch, line 67: We only use this for the last block, so this should always just be the full slice, no?
```suggestion
    buffer[0..remaining].copy_from_slice(message);
```

specs/sha3/src/sha3.rs
```diff
@@ -0,0 +1,149 @@
+/// SHA-3 and SHAKE functions — FIPS 202, Section 6.
+use crate::sponge::keccak;
+
+// Rate constants in bytes: rate = (1600 - 2*capacity) / 8
```
jschneider-bensch:
```suggestion
// Rate constants in bytes: rate = (1600 - capacity) / 8
```

specs/sha3/src/sponge.rs
```diff
@@ -0,0 +1,165 @@
...
+///    state, and apply Keccak-f. The final partial block is padded with
+///    the domain separation byte `delim` and the pad10*1 terminator `0x80`.
+/// 2. Squeeze: extract `OUTPUT_LEN` bytes from the state, applying
+///    Keccak-f between each `rate`-byte block of output.
+///
+/// The `OUTPUT_LEN < usize::MAX - 200` precondition is a Rust implementation
+/// artifact to prevent arithmetic overflow; FIPS 202 places no upper bound
+/// on the output length.
+#[hax_lib::requires(rate > 0 && rate <= 200 && rate % 8 == 0 && OUTPUT_LEN < usize::MAX - 200)]
+pub fn keccak<const OUTPUT_LEN: usize>(rate: usize, delim: u8, message: &[u8]) -> [u8; OUTPUT_LEN] {
```
jschneider-bensch, line 163: Just a minor question regarding style: Why did you use a const generic for `OUTPUT_LEN`, but not for `rate` or `delim`?

specs/sha3/tests/sponge_decomposition.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    // SHA3-224: rate=144, delim=0x06, out=28
+    check::<28>(144, 0x06, &empty);
+    check::<28>(144, 0x06, short);
+    check::<28>(144, 0x06, &long);
+    // SHA3-256: rate=136, delim=0x06, out=32
+    check::<32>(136, 0x06, &empty);
+    check::<32>(136, 0x06, short);
+    check::<32>(136, 0x06, &long);
+    // SHA3-384: rate=104, delim=0x06, out=48
```
jschneider-bensch, line 31: Why are the others not all tested on `empty` and `long` as well?

jschneider-bensch (changes requested):
I've looked in more detail at the spec now.

jschneider-bensch:
We should think about what we want to run on CI too, when anything in the spec changes. At least the tests for the spec would be good, and also the verification, if they run in reasonable time.

specs/sha3/hax.sh
abentkamp: Right, the Lean extraction here is outdated. I could add a redirect, so `hax.sh extract_lean` would run `hax_aeneas.py`. Does that make sense? Or should I rather rename the file to `hax_fstar.sh` (Then the edits you suggest above would be obsolete)?

jschneider-bensch: I'm fine with just renaming the `hax.sh` into `hax_fstar.sh` and removing the lean bits from it. You can ignore my renamings then. That way it's clear which script is responsible for which extraction.

specs/sha3/proofs/aeneas-lean/HacspecSha3/Extraction/Funs.lean
abentkamp: I would like to commit it if that's okay. It makes it much easier to use the Lean package as a dependency elsewhere.

specs/sha3/proofs/aeneas-lean/HacspecSha3.lean
abentkamp: Yes, this is the main entrypoint. Also, it's a Lean convention to have this main file named the same as the Lean package.

specs/sha3/proofs/aeneas-lean/HacspecSha3/Extraction/Funs.lean
jschneider-bensch: Okay, that's a good reason to keep it :+1: (Is there a way to ensure that a dependant on the Lean package would get the same extraction if they were to extract it, or is that not a concern?)

With F* there's an occasional annoyance that the extraction tooling and Makefile may or may not clean up already existing state, which can produce weird errors, which is why I'd prefer not checking in the F* files. And the F* setups, here and elsewhere, should be fixed to flush any caches/old states by default (but that's for us to fix in a different PR).

specs/sha3/proofs/aeneas-lean/HacspecSha3.lean
jschneider-bensch: Thanks!

specs/sha3/src/lib.rs
abentkamp: Done.

specs/sha3/src/sponge.rs
abentkamp: Done.

crates/algorithms/sha3/src/generic_keccak.rs
abentkamp: Done.

crates/algorithms/sha3/src/generic_keccak.rs
abentkamp: Done.

specs/sha3/tests/cavp.rs
abentkamp: Done.

specs/sha3/NIST.FIPS.202.pdf
abentkamp: Removed.

CHANGELOG.md
abentkamp: Removed.

specs/sha3/.gitignore
abentkamp: Removed.

specs/sha3/src/sponge.rs
abentkamp: I don't think it was necessary, but now I've rewritten the `absorb_rec` function, so that these extra parameters actually serve a purpose in Rust. My motivation for this change was that `absorb_rec` will then be easier to verify.  (see https://github.com/cryspen/libcrux/pull/1399#discussion_r3199710527)

specs/sha3/src/sponge.rs
abentkamp: As I wrote above, I changed the implementation of `absorb_rec`, so that this is actually needed now.

specs/sha3/src/sha3.rs
abentkamp: Done.

specs/sha3/tests/sponge_decomposition.rs
abentkamp: I've added them.

specs/sha3/hax.sh
abentkamp: Done.

specs/sha3/hax.sh
abentkamp: Obsolete with https://github.com/cryspen/libcrux/pull/1399#discussion_r3434619387

specs/sha3/hax.sh
abentkamp: Obsolete with https://github.com/cryspen/libcrux/pull/1399#discussion_r3434619387

specs/sha3/hax.sh
abentkamp: Obsolete with https://github.com/cryspen/libcrux/pull/1399#discussion_r3434619387

specs/sha3/hax.sh
abentkamp: Obsolete with https://github.com/cryspen/libcrux/pull/1399#discussion_r3434619387

specs/sha3/hax.sh
abentkamp: Obsolete with https://github.com/cryspen/libcrux/pull/1399#discussion_r3434619387

specs/sha3/hax.sh
abentkamp: Obsolete with https://github.com/cryspen/libcrux/pull/1399#discussion_r3434619387

specs/sha3/hax.sh
abentkamp: Obsolete with https://github.com/cryspen/libcrux/pull/1399#discussion_r3434619387

specs/sha3/tests/cavp.rs
abentkamp: Done.

specs/sha3/src/sponge.rs
karthikbhargavan: We only ever use const generics for things that need to be used in array lengths, and this often only happens for array outputs of functions. We could switch to a style where we use Vec everywhere instead of fixed-sized arrays and in that case, we would even eliminate OUTPUT_LEN as a const parameter. Having a const parameter in the spec is actually quite awkward when mapping the implementation to the spec.

specs/sha3/proofs/aeneas-lean/HacspecSha3/Extraction/Funs.lean
abentkamp: No, I wouldn't know how to ensure that a dependant would get the same extraction. Maybe we could add a CI job that reruns the extraction and checks that the checked in files are correct?

But for F*, how do dependants reextract the files?

specs/sha3/src/lib.rs
abentkamp: I've elaborated on why the issue is relevant here.

The other place that you saw was in a previous version of this PR. I've moved the `#cfg(not(hax_backend_lean))]` switch to here instead.

specs/sha3/src/sponge.rs
abentkamp: Done.

specs/sha3/src/sponge.rs
abentkamp: done.

specs/sha3/src/sponge.rs
abentkamp: @karthikbhargavan ?

specs/sha3/src/sponge.rs
abentkamp: Done.

crates/algorithms/sha3/src/generic_keccak.rs
abentkamp: Done.

specs/sha3/src/sponge.rs
abentkamp: Apologies, didn't see your response...

specs/sha3/proofs/fstar/extraction/Hacspec_sha3.fst
abentkamp: I've removed them.

specs/sha3/NIST.FIPS.202.pdf
jschneider-bensch: Thanks!

specs/sha3/hax.sh
jschneider-bensch: Thank you!

specs/sha3/proofs/aeneas-lean/HacspecSha3/Extraction/Funs.lean
jschneider-bensch: Typically the extraction would be re-run, I'd say.

specs/sha3/src/sponge.rs
jschneider-bensch: If having a vector is easier for mapping then that would be fine as well, I think, but might incur more runtime length checks?

specs/sha3/tests/cavp.rs
jschneider-bensch: Great, thanks!

karthikbhargavan:
Right now, this spec is in a very delicate balance, since we have working proofs in F* and in Lean against this spec. I would really love to get the spec merged without too many changes, fix up the proofs if needed, merge those, and then we can iterate if needed. Once we have 3 specs in, we may also have enough data to commit to a uniform style.

specs/sha3/src/sponge.rs
karthikbhargavan: Its a style thing, I would not change this now. All the proofs would need updating. But we can try to switch uniformly from arrays to Vecs in the future and see if we can adapt the proofs accordingly. There are many different styles in which we can write the spec, and so we need to agree on a preferred style after seeing a few examples.

abentkamp:
I've undone all changes to the Rust code that would change the extraction.

specs/sha3/hax_aeneas.py
abentkamp: They are from `hax-evit`, but they are part of our current submission that will hopefully be approved soon.

abentkamp: I've added a Readme.

abentkamp:
I think I addressed all the comments, except that I've undone some of them in
[[sha3] undo Rust code changes](https://github.com/cryspen/libcrux/pull/1399/commits/a394dda4eb16c3abbfd874d1ffa932437f02d639), so that we don't break our existing proofs. Let's do that in a later iteration if needed.

.github/workflows/sha3-spec-test.yml
```diff
@@ -0,0 +1,53 @@
+name: SHA3 Spec - Build & Test
+
+on:
+  merge_group:
+    paths:
+      - "specs/sha3/**"
+      - "crates/algorithms/sha3/**"
+      - "crates/testing/kats/**"
```
jschneider-bensch:
```suggestion
```
Unfortunately you can't give paths for `merge_group` triggers.

jschneider-bensch (approved):
I filed #1486 for keeping track of the changes that were reverted. Other than the small CI nit below I think this is good to go now, thanks!

PR #1373, opened 2026-03-23
Add comprehensive test vectors across all algorithm crates

karthikbhargavan:
I used Claude to generate Wycheproof tests for all the algorithms, and to add other test vectors.
This PR is the result.

- New Wycheproof tests: hkdf (SHA-256/384/512), hmac (SHA-1/256/384/512), curve25519 (X25519 ECDH), p256 (ECDH)
- Migrate chacha20poly1305 and ecdsa from custom JSON parsing to typed wycheproof crate API; add ECDSA SHA-512 coverage and invalid vector testing
- Add CAVP test vectors for sha2 (SHA-224/256/384/512 ShortMsg + LongMsg)
- Add test vectors for blake2 (BLAKE2b-256 and BLAKE2s-256, keyed + unkeyed)
- Add test vectors for poly1305 (24 vectors with various input sizes)

```
 ┌───────────────────┬─────────────────────────────┬────────────────────────────────────────────────────────┐
  │     Algorithm     │        Test Vectors         │                         Source                         │
  ├───────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────┤
  │ HKDF              │ 252 (86 + 83 + 83)          │ Wycheproof (SHA-256/384/512)                           │
  ├───────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────┤
  │ HMAC              │ 692 (170 + 174 + 174 + 174) │ Wycheproof (SHA-1/256/384/512)                         │
  ├───────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────┤
  │ X25519            │ 518                         │ Wycheproof                                             │
  ├───────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────┤
  │ P-256 ECDH        │ 355                         │ Wycheproof                                             │
  ├───────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────┤
  │ ChaCha20-Poly1305 │ 622 (316 + 306)             │ Wycheproof (migrated from custom JSON)                 │
  ├───────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────┤
  │ ECDSA P-256       │ 1,012 (471 + 541)           │ Wycheproof SHA-256 + SHA-512 (migrated, added SHA-512) │
  ├───────────────────┼─────────────────────────────┼────────────────────────────────────────────────────────┤
  │ SHA-2             │ 68 (17 × 4 variants)        │ CAVP-format (SHA-224/256/384/512)                      │
[...]

PR #1400, opened 2026-05-06
kats: add sha2 CAVP test vectors

robinhundt:
Test vectors are from
https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/secure-hashing .

This is based on #1373, keeping the parsing code for the test vector files but updating the vectors to those issued by NIST.

Currently, I have not included the Monte Carlo tests. Should we track that as a future work in an issue?

[skip changelog]


crates/testing/kats/Cargo.toml
```diff
@@ -15,7 +15,9 @@ repository.workspace = true
 hex = { version = "0.4.3", features = ["serde"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0" }
+cavp = { path = "../../../cavp", optional = true }
```
jschneider-bensch: We could add `cavp` to the workspace dependencies and use it as such here. In case the path changes at some point in the future.

jschneider-bensch (approved):
I think this looks good, with one optional suggestion.

I created #1402 and its sub-issue #1403 since we don't run these on CI yet at the moment. But we should figure out #1133 first.

robinhundt:
@jschneider-bensch I added cavp as a workspace dependency. I also briefly refactored the sha2 testing code to be slightly more idiomatic.

Thanks for creating the issues!

Issue #1407, opened 2026-05-06
[kats/CI] Check for wycheproof changes in CI


robinhundt:
Updating wycheproof ML-DSA test vectors has uncovered an issue with our AVX2 verify implementation (see #1398). To reliably catch issues that might be surfaced by new test vectors, it would be helpful if we knew when there are new versions of the test vectors we use.

In the wycheproof-rs repo, there is an [open PR](https://github.com/randombit/wycheproof-rs/pull/27) which adds a script that checks upstream test vectors against local ones. We could adapt this script to our needs and run it on a regular interval in CI to be notified of changes to the test vectors.

PR #1409, opened 2026-05-06
[kats/ml-kem]: update ml-kem wycheproof vectors

robinhundt:
This updates the ml-kem wycheproof test vectors to the latest state, adapts the schema and the libcrux-ml-kem tests.

I'm not super happy with the code duplication introduced in the loading test vector and testing code but also didn't want to overengineer this.

I created #1407 to track potentially regularly checking for wycheproof updates. If we decide to do that, and then notice increased churn by schema changes, we could think about generating the rust types from the [json schemas](https://github.com/C2SP/wycheproof/tree/6d9d6de30f02e229dfc160323722c3ddac866181/schemas) in the wycheproof repo.

The wycheproof repo also contains keygen seed test files [such as this](https://github.com/C2SP/wycheproof/blob/6d9d6de30f02e229dfc160323722c3ddac866181/testvectors_v1/mlkem_512_keygen_seed_test.json), but as for as I can tell, these tests are also included in the keygen/decaps test files.

[skip changelog]


libcrux-ml-kem/tests/wycheproof.rs
```diff
@@ -41,26 +50,81 @@ macro_rules! wycheproof_test {
...
+                        };
                         let encapsulation_key: MlKemPublicKey<_> = bytes.into();
 
-                        // validate the key (should fail for ModulusOverflow cases)
-                        assert!(!validate_public_key(&encapsulation_key));
+                        let is_valid = validate_public_key(&encapsulation_key);
+                        match test.result {
+                            MlKemResult::Invalid => assert!(
+                                !is_valid,
+                                "tc_id {} failed. key is valid but should be invalid",
```
jschneider-bensch: I think these  messages are a bit ambiguous in that it's not immediately clear what's the ground truth and what's our hypothetical implementation behavior.

```suggestion
                                "tc_id {} failed. invalid public key passes key validation",
```

libcrux-ml-kem/tests/wycheproof.rs
```diff
@@ -41,26 +50,81 @@ macro_rules! wycheproof_test {
...
+                                test.tc_id,
+                            ),
+                            MlKemResult::Valid => assert!(
+                                is_valid,
+                                "tc_id {} failed. key is invalid but should be valid",
```
jschneider-bensch:
```suggestion
                                "tc_id {} failed. valid public key rejected by key validation",
```

jschneider-bensch (approved):
I think these look good, thanks!
Just one minor suggestion around the assertion failure messages.

Also worth noting that these cover one set of APIs for ML-KEM, where we have statically known array sizes for keys, ciphertexts etc, so it's fine to skip invalid length test cases. We also have the `libcrux-traits` based APIs, on the `MlKem*` structs, e.g. on [`MlKem512`](https://docs.rs/libcrux-ml-kem/latest/libcrux_ml_kem/mlkem512/struct.MlKem512.html), which includes a slice-based API that does runtime length-checks. When we eventually add tests for these (not for this PR) we should not skip the invalid length test cases.

robinhundt:
> Also worth noting that these cover one set of APIs for ML-KEM, where we have statically known array sizes for keys, ciphertexts etc, so it's fine to skip invalid length test cases. We also have the `libcrux-traits` based APIs, on the `MlKem*` structs, e.g. on [`MlKem512`](https://docs.rs/libcrux-ml-kem/latest/libcrux_ml_kem/mlkem512/struct.MlKem512.html), which includes a slice-based API that does runtime length-checks. When we eventually add tests for these (not for this PR) we should not skip the invalid length test cases.

Maybe we could change the tests in a future PR to only test the slice based API, as they're a thin wrapper around the array based APIs?

PR #1413, opened 2026-05-07
[kats / chacha20poly1305]: Add wycheproof chacha20poly1305 tests

robinhundt:
This PR is split into two commits:

- kats: add {x}chacha20poly1305 wycheproof vectors
- chacha20poly1305: use libcrux-kats vectors

Working on the second commit, I realized that the previous test code was not working as intended.

```rust
let mut ctxt = msg.clone();
let tag = match libcrux_chacha20poly1305::encrypt(key, msg, &mut ctxt, aad, nonce) {
    Ok((_v, t)) => t,
    Err(_) => {
        *tests_run += 1;
        continue;
    }
};
```

This will always silently fail, because `encrypt` requires the `ctxt` buffer to be big enough to hold the ciphertext **and tag**. `encrypt` will always return an `Err(AeadError::CiphertextTooShort)` which is silently ignored in the `Err` branch of the `match` and treated as a success.

The second commit also handles the `TestResult::Invalid` cases differently than before. These test are either due to wrong Nonce sizes, or due to a modified tag. As per the [flag description](https://github.com/C2SP/wycheproof/blob/6d9d6de30f02e229dfc160323722c3ddac866181/testvectors_v1/chacha20_poly1305_test.json#L42):
> The test vector contains a ciphertext where the tag has been modified. The goal of the test vector is to detect implementations with partial or incorrect tag verification.
Which is why the new test code now tries to decrypt these ciphertexts+tags, and fails if decryption succeeds.

At the moment, these tests are not executed in CI. I created #1411 to track this.

[skip changelog]

### DIFF
--- a/crates/testing/kats/src/wycheproof/schema_common.rs
+++ b/crates/testing/kats/src/wycheproof/schema_common.rs
@@ -3,6 +3,14 @@
 
 use serde::{Deserialize, Serialize};
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum TestResult {
+    Valid,
+    Invalid,
+    Acceptable,
+}
+
 #[derive(PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Source {
--- a/crates/testing/kats/src/wycheproof/schema_common.rs
+++ b/crates/testing/kats/src/wycheproof/schema_common.rs
@@ -3,7 +3,7 @@
 
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "snake_case")]
 pub enum TestResult {
     Valid,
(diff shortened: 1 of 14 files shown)


crates/testing/kats/src/wycheproof/schema_common.rs
```diff
@@ -3,6 +3,14 @@
 
 use serde::{Deserialize, Serialize};
 
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum TestResult {
```
jschneider-bensch, line 8: I think this is the same result schema as for the ML-KEM tests. Do you think it makes sense to use that there as well?

jschneider-bensch (approved):
I think these look good, and the fixes and changes to the test logic are most welcome!

jschneider-bensch:
I would hold off on merging for a little bit until we have the current round of releases out. Just to avoid CI noise.

crates/testing/kats/src/wycheproof/schema_common.rs
robinhundt: The ML-KEM and ML-DSA result structs did not have an `Acceptable` variant. However, the [wycheproof schema](https://github.com/C2SP/wycheproof/blob/6d9d6de30f02e229dfc160323722c3ddac866181/schemas/common.json#L57-L65) uses the same result type for all test sets. I've added a commit that removes the individual result enums and instead uses the `TestResult` for all.

PR #1417, opened 2026-05-11
[psq] Remove PSQv1

jschneider-bensch:
This removes PSQv1 and the related features `v1` and `test-utils`, as well as example and benchmarking code for PSQv1.


franziskuskiefer (approved):
:shipit:

PR #1418, opened 2026-05-11
[kats/p256] Add wycheproof test vectors and tests for p256 ECDH/ECDSA

### LINKED ISSUE #1416: [kats] Update p256 test vectors

robinhundt:
First commit adds the [ecdh_secp256r1_ecpoint_test.json](https://github.com/cryspen/libcrux/compare/robin/kats-p256-add-test-vector?expand=1#diff-4523de4043484f45365838081d13a988789e01bf5042a7807819f987b6b320a4) wycheproof test vector file and accompanying parsing code in libcrux-kats (based on #1373).

The second commit adds a test in `crates/algorithms/p256` that evaluates the test vectors against the libcrux-p256 ECDH implementation.

The tested crate also contains an ECDSA implementation which is not tested as part of this PR.
@jschneider-bensch Should this be done in a follow-up or should I include it here?

closes #1416 

[skip changelog]

### DIFF
--- a/crates/algorithms/p256/Cargo.toml
+++ b/crates/algorithms/p256/Cargo.toml
@@ -30,6 +30,8 @@ check-secret-independence = [
 
 
 [dev-dependencies]
+libcrux-p256 = { workspace = true, features = ["expose-hacl"] }
 libcrux-traits = { workspace = true, features = [
     "generic-tests",
 ] }
+libcrux-kats = { workspace = true, features = ["ecdh"] }
--- a/crates/algorithms/p256/Cargo.toml
+++ b/crates/algorithms/p256/Cargo.toml
@@ -30,8 +30,9 @@ check-secret-independence = [
 
 
 [dev-dependencies]
+der = "0.8.0"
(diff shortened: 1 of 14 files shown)


jschneider-bensch:
Feel free to include the ECDSA testing here as well.

jschneider-bensch (approved):
These look good to me! The ECDSA ones are really a bit tricky with the necessary decoding beforehand, but at least that logic should stay the same when we port these tests to `libcrux-ecdsa`.
I wonder where it would make the most sense to document that there are not wycheproof vectors for ECDSA-P256-SHA384, probably in `libcrux-ecdsa`?

robinhundt:
> These look good to me! The ECDSA ones are really a bit tricky with the necessary decoding beforehand, but at least that logic should stay the same when we port these tests to `libcrux-ecdsa`. I wonder where it would make the most sense to document that there are not wycheproof vectors for ECDSA-P256-SHA384, probably in `libcrux-ecdsa`?

Yes, I would document it there.

PR #1419, opened 2026-05-11
[workspace] Bump to release versions

jschneider-bensch:
This bumps all pre-release versions to release and updates the changelogs.


franziskuskiefer (approved):
We should still test some consumer before merging (hpke, openmls) but lgtm

jschneider-bensch:
There are more crates that need a release for the `rand` bump.

PR #1435, opened 2026-05-13
[CT/ML-DSA] Add valgrind constant time tests and ML-DSA valgrind requests 

### LINKED ISSUE #1111: [secrets] Check constant-time assembly on CI
- How should the check work?
- Is it useful to check on toy code vs actual application?

robinhundt:
This PR adds a `ctgrind_test` crate that performs constant time tests of ML-DSA, ML-KEM, and SHA-3 by using the [valgrind memcheck tool](https://valgrind.org/docs/manual/mc-manual.html). 

Each binary marks secret data as "undefined" with [`crabgrind::memcheck::mark_memory`](https://docs.rs/crabgrind/latest/crabgrind/memcheck/fn.mark_memory.html), runs a cryptographic operation, then marks the output as defined. Valgrind flags any use of a value containing undefined bits which would result in observable differences[^1] in program behaviour, indicating a potential timing side channel.

For libcrux-ml-dsa, appropriate valgrind requests are already issued in the crate itself, if compiled with the `valgrind_ct_test` cfg enabled.

These tests are only on a best effort basis. They cannot, e.g., guarantee the absence of cache side channel attacks.

resolves #1111 

[^1]: > Checks on definedness only occur in three places: when a value is used to generate a memory address, when control flow decision needs to be made, and when a system call is detected, Memcheck checks definedness of parameters as required. https://valgrind.org/docs/manual/mc-manual.html#mc-manual.value

----

Reasoning for declassifications in ML-DSA by @jschneider-bensch:

## Declassifications in ML-DSA

The signing operation in ML-DSA includes some operations that
technically depend on secret data, but are in fact safe under the
assumptions of the Dilithium security proof [1][dilithium-round3]. In these cases we
explicitly mark the memory as `MemState::Defined` for valgrind, in
order to avoid false positives.

### Verifier Challenge

Revealing the verifier challenge `commitment_hash_candidate` is safe
in the random oracle model.

> The challenge reveals information about H(μ||w₁) also
> in the case of rejected y, but this does not reveal any
> information about the secret key when H is modelled as
> a random oracle and w₁ has high min-entropy.

-- Section 5.5 of the [Dilithium Specification for Round
3 of the NIST Post-Quantum Cryptography
Standardization][dilithium-round3].

### Infinity Norm Checks

It is safe to leak the index of a violating coefficient during ML-DSA
signature generation.

See section 5.5 of the [Dilithium Specification for Round 3 of the NIST
Post-Quantum Cryptography Standardization][dilithium-round3].

### `w0` after norm checks have passed

After the norm checks have passed it is safe to serialize the
signature, so any value that could be derived from the signature and
the public key is safe to leak.

At this point `w0` = w - c * s2, and

    A * `mask` - `verifier_challenge` * t = w - c * s2

where `mask` and `verifier_challenge` are hold values that are part of
the signature to be serialized, and A and t are part of the public
key.

### `challenge_times_t0` after norm checks have passed

After the norm checks have passed it is safe to serialize the
signature, so any value that could be derived from the signature and
the public key is safe to leak.

`t0` is technically part of the signing key, but only for compression
of the public key. It can be considered public. `verifier_challenge`
will be part of the signature that is now safe to serialize, cf. [FIPS
204, section
6.1](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf)

[dilithium-round3]: https://pq-crystals.org/dilithium/data/dilithium-specification-round3-20210208.pdf

[skip changelog]

### DIFF
--- a/.github/workflows/libcrux-ct-test.yml
+++ b/.github/workflows/libcrux-ct-test.yml
@@ -1,0 +1,47 @@
+name: Libcrux - Constant Time Tests
+
+on:
+  merge_group:
+  push:
+    branches: ["main", "dev"]
+  pull_request:
+    branches: ["main", "dev"]
+  workflow_dispatch:
+
+permissions: {}
+
+env:
+  CARGO_TERM_COLOR: always
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+
+jobs:
+  ct-test:
+    name: Constant Time Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      
+      - name: Install valgrind / libclang-dev
+        run: sudo apt-get update && sudo apt-get install -y valgrind libclang-dev
+
+      - name: Compile ctgrind-test crate
+        env:
+          # needed for the ML-DSA CT tests
+          RUSTFLAGS: '--cfg valgrind_ct_test'
+        run: cargo build --package ctgrind-test --profile release-debug --bins
+
+      - name: Run valgrind ML-DSA
+        run: valgrind --tool=memcheck --error-exitcode=1 --track-origins=yes target/release-debug/mldsa
+
+      - name: Run valgrind ML-KEM
+        run: valgrind --tool=memcheck --error-exitcode=1 --track-origins=yes target/release-debug/mlkem
+
+      - name: Run valgrind SHA-3
+        run: valgrind --tool=memcheck --error-exitcode=1 --track-origins=yes target/release-debug/sha3
+
--- a/.github/workflows/libcrux-ct-test.yml
+++ b/.github/workflows/libcrux-ct-test.yml
@@ -3,9 +3,9 @@ name: Libcrux - Constant Time Tests
 on:
   merge_group:
   push:
-    branches: ["main", "dev"]
+    branches: ["main"]
   pull_request:
-    branches: ["main", "dev"]
+    branches: ["main"]
   workflow_dispatch:
 
 permissions: {}
--- a/crates/utils/ctgrind-test/Cargo.toml
+++ b/crates/utils/ctgrind-test/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "ctgrind-test"
+version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+edition.workspace = true
+repository.workspace = true
+readme.workspace = true
+
+[dependencies]
+crabgrind = "0.2"
+libcrux-sha3 = { workspace = true }
+libcrux-ml-kem = { workspace = true, features = ["mlkem512"] }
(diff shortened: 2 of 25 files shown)


jschneider-bensch:
I have a branch [here](https://github.com/cryspen/libcrux/tree/jonas/ct-mldsa) that hopefully fixes the hax issue and adds more documentation around the declassifications. Feel free to take anything from that branch, but note that the links that I added in the ctgrind-test readme should be updated before we merge.

EDIT: And we probably need to re-run C extraction as well, still.

jschneider-bensch:
Huh, I did not realize you had already updated the C extraction, and did it again. Sorry! But I wonder why I get diff to your extraction then. :thinking:

jschneider-bensch:
Ah, nevermind, my extraction setup was in a bad state, I'll revert that.

jschneider-bensch:
The changes here mean we'd have to update the extraction config for the combined extraction as well, but I'm postponing that until we do the Eurydice update required for #1328.

libcrux-ml-dsa/cg/libcrux_mldsa65_portable.c
jschneider-bensch, line 1: This file is new, and means we need to update the extraction config here as well, since this is supposed to be a header only extraction. Like with the combined extraction I would also like to defer this for a moment, since we need to clean up the configs in any case for #1328.

.github/workflows/libcrux-ct-test.yml
```diff
@@ -0,0 +1,47 @@
+name: Libcrux - Constant Time Tests
+
+on:
+  merge_group:
+  push:
+    branches: ["main", "dev"]
```
jschneider-bensch: We don't actually have a long-standing `dev` branch anymore, so we can drop it here, I think.
```suggestion
    branches: ["main"]
```

.github/workflows/libcrux-ct-test.yml
```diff
@@ -0,0 +1,47 @@
...
+  pull_request:
+    branches: ["main", "dev"]
```
jschneider-bensch:
```suggestion
    branches: ["main"]
```

crates/utils/ctgrind-test/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "ctgrind-test"
+version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+edition.workspace = true
+repository.workspace = true
+readme.workspace = true
```
jschneider-bensch: We're not likely to publish this, but in any case, there's a local readme here.
```suggestion
readme = "README.md"
```

jschneider-bensch (approved):
I think this looks good, with a few really minor suggestions.

I'll create follow-up issues for things not covered here:
- ML-DSA Key Generation: #1443 
- ML-KEM Key Generation: #1444 
- Updating C extraction toolchain & cleaning up ML-DSA and combined extractions: #1328 
- Running the regular tests under valgrind: #1445

PR #1439, opened 2026-05-18
[ML-DSA]: improve wycheproof test robustness

robinhundt:
Previously, the test would ignore unexpected test failures. We now panic should this happen after an update of the test vectors.

I've also documented in the test that we do not perform checking of s1/s2 during signing. There are test vectors for this in wycheproof, but these checks are not mandated or recommended by the specs. I've also linked to this [ml-dsa-native PR](https://github.com/pq-code-package/mldsa-native/pull/1003) which provides a discussion of this topic. They also do not perform these checks.

I noticed this while working on https://github.com/cryspen/libcrux-iot/pull/139.

[skip changelog]


jschneider-bensch (approved):
Thanks!

PR #1446, opened 2026-05-19
[secrets] valgrind integration

### LINKED ISSUE #1440: [secrets] Integrate valgrind requests with libcrux-secrets
In #1435 we implemented an initial version of constant time checks for ML-DSA, ML-KEM, and SHA-3 using valgrind client requests and the memcheck tool.

As ML-DSA contains some code that is inherently not constant time, we added [`ct_classify/ct_declassify` operations](https://github.com/cryspen/libcrux/blob/e9184c4e786e7c8fba829e3acb05c197987aa0b0/libcrux-ml-dsa/src/ct_test.rs) that use [crabgrind](https://github.com/cryspen/libcrux/blob/e9184c4e786e7c8fba829e3acb05c197987aa0b0/libcrux-ml-dsa/src/ct_test.rs) (when compiled with the `valgrind_ct_test` cfg) to mark memory as undefined/defined where needed.

Libcrux also contains the [libcrux-secrets](https://docs.rs/libcrux-secrets/latest/libcrux_secrets/index.html) crate, which provides a new-type + trait approach to checking for secret independence. It has APIs for de-/classifying values (e.g. the [`Classify`](https://docs.rs/libcrux-secrets/latest/libcrux_secrets/trait.Classify.html) trait).

I propose we integrate the valgrind client requests to mark memory with the libcrux-secrets API. Each `classify` would mark memory as `Undefined` and each `declassify` would mark it as defined.

This would give us the benefit of the API based approach of `libcrux-secrets`, combined with the runtime validation of valgrind. The latter could catch issues with the compilation of the portable constant time swap/select operations.

robinhundt:
This PR integrates valgrind client requests into the libcrux-secrets crate.

When the crate is compiled with the `--cfg valgrind_ct_test` the following operations mark memory as `Undefined`:
- conversions from non-secret types to secret types
- `classify`, `classify_ref`, `classify_ref_mut` and `classify_mut_slice` when the feature `check-secret-independence` is enabled and they convert to the secret type

The following operations mark memory as `Defined`:
- `declassify`, `declassify_ref`, `declassify_ref_mut` and `declassify_mut_slice`

In order to properly track the memory states in the select and swap implementations when running under valrind, I've refactored the portable swap/select implementations and added explicit `ct_classify` calls necessary `declassify` calls in the `aarch64` implementations.

I've also added a CI job that runs the crate's tests on ubuntu-latest and ubuntu-24.04-arm under valgrind to validate that valgrind reports no errors on these platforms (This commit largely generated by claude code with Opus 4.7).

Open questions:
- Currently, the cargo feature `check-secret-independence` and the `--cfg valgrind_ct_test` are independent of each other. You could enable just one or the other. While there are mostly no client requests issued when the feature is not present (as these mostly happen in the classify calls) there are some in the swap/select implementations that would be present. The `ct_{de}classify` functions could also be made conditional on the cargo feature.
- Alternatively, when the cfg is present but not the feature, we could actually also issue valgrind requests for the "public" values when classify is called on them. This could help catch issues in downstream crates where currently the `check-secret-independence` feature can't be enabled.

Other things I noticed:
- The `Swap` trait implementations currently only assert in debug mode that the lengths are equal. In release mode, they panic when `other.len() < self.len()` and ignore values if `other.len() > self.len()`. Is this intended or should that be changed to a proper `assert_eq`?
- The `Scalar` trait is not sealed. Downstream crates could implement it for their own types. I don't think this currently has any real implications, as even the `De-/Classify` impls for secret types that use transmute should be safe for arbitrary `T`, but it seems like something unintended.

closes #1440


crates/utils/secrets/CHANGELOG.md
jschneider-bensch, line 1: As long as we keep these individual changelogs, the entries here should be categorized under `Added`, `Changed`, `Fixed`, etc headings like in the workspace changelog. (We didn't do this retroactively, after switching to this style.)

crates/utils/secrets/src/mem_requests.rs
jschneider-bensch, line 1: Does it make sense to make use of these in ML-DSA already, rather than having a separate copy?

crates/utils/secrets/src/mem_requests.rs
```diff
@@ -0,0 +1,55 @@
...
+//! When the `cfg`` is not enabled, these operations are no-ops.
+//!
+//! [`Undefined`]: https://docs.rs/crabgrind/latest/crabgrind/memcheck/enum.MemState.html#variant.Undefined
+//! [`Defined`]: https://docs.rs/crabgrind/latest/crabgrind/memcheck/enum.MemState.html#variant.Defined
+
+/// Mark memory as secret.
+///
+/// No-op if `valgrind_ct_test` cfg is not enabled.
+#[inline(always)]
+pub(crate) fn ct_classify<T: ?Sized>(val: &T) {
```
jschneider-bensch, line 20: Perhaps we could take this opportunity to rename these `valgrind_{de}classify`, since, as you point out in the Readme there are specific uses of undefined values that get flagged by valgrind (control flow decisions and memory addressing look like the relevant ones for us) and others that would count as non-constant time do not.

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
...
 
+## Unreleased
+
+### Added
+
+- (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Integrate valgrind requests when cfg `valgrind_ct_test` is set
+
+### Removed
+
+- (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const constructors of secret types
```
jschneider-bensch: Would it make sense to categorize this as "Changed", since the constructors are still present, but no longer `const`?

crates/utils/secrets/src/int.rs
```diff
@@ -145,119 +145,129 @@ pub trait Select {
...
 
     /// This macro implements `Select` for public integer type
     /// `&[$ty]` and its secret version `&[$secret_ty]`.
     macro_rules! impl_select {
-        ($ty:ty, $secret_ty:ty, $is_non_zero:ident) => {
+        ($ty:ty, $secret_ty:ident, $is_non_zero:ident, $cast:path) => {
             impl Select for [$ty] {
                 fn select(&mut self, other: &Self, selector: crate::U8) {
-                    let mask = ($is_non_zero(selector.declassify()) as $ty).wrapping_sub(1);
+                    let mask = $cast($is_non_zero(selector)).wrapping_sub($secret_ty(1));
```
jschneider-bensch: Since you introduced the `ct_swap_mask` function below, could you re-use that here?

jschneider-bensch (commented):
This looks good, overall, I think! I just have some suggestions, and have to double check that there are no bad interactions with any of the extractions.

I think it makes sense to keep the valgrind `cfg` and the `check-secret-independence` feature separate for now.

I'm fine leaving it as a debug_assert at the moment, and I don't think we necessarily need to seal the traits. I would hope that anyone implementing these for themselves is aware that that could be counter-productive.

robinhundt:
Oh, while reviewing your comments @jschneider-bensch I noticed that I accidentally introduced a bug by reworking how the mask is computed for the portable swapping.

The new implementation assumes that the selector is either 0 or 1 and nothing else, whereas the previous implementation differentiated between 0 or anything else. 

While the two's complement negation produces [slightly better assembly](https://rust.godbolt.org/z/7fd9xqK8e) I don't think we can assume that the selector is always 0 or 1.

I'm going to revert the change to the mask.

crates/utils/secrets/src/mem_requests.rs
robinhundt: Mh, that is a good point. However, I think in the future we might want to also add instrumentation for LLVM's memory sanitizer. That could then also be done in these methods and toggle via a different cfg. Should we add that, we'd only need to hopefully change these two methods. If they had valgrind in the name, we'd need to rename and change all call sites.

I'm leaning towards leaving them like this.

CHANGELOG.md
robinhundt: Done

crates/utils/secrets/src/int.rs
robinhundt: I've restored the original mask implementation (in a function rather than a macro) and reuse it now for select and swap.

I've only dropped the superfluous `& 1` at the end of the `is_non_zero` check.

robinhundt:
I've now reverted the mask computation to how it was before, but cleaned up a bit with a function instead of a macro which is also reused for the select and swap implementation.

I also spotted another place where there was a `ct_classify` missing, in the `secret` function constructor in `secret_integer.rs`. The reason for why I removed the `const` qualifier there, but I didn't add the classify...

I've made the change to ML-DSA in a stacked PR #1457

crates/utils/secrets/Cargo.toml
```diff
@@ -14,6 +14,9 @@ exclude = ["/proofs"]
 [dependencies]
 hax-lib.workspace = true
 
+[target.'cfg(valgrind_ct_test)'.dependencies]
+crabgrind.workspace = true
```
robinhundt, line 18: Todo: Check whether this actually does what we expect it to.

robinhundt: Mhh, reading the [manifest format description](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies) I'd have assumed that a target dependency with a custom cfg would not work. However, that does not seem to be the case, as the commented dependency specification with a custom cfg seems to be working as intended.

As this pattern is also used in other crates, and I can't find anything about this being problematic, I assume that it is fine to use.

CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
...
 
+## Unreleased
+
+### Added
+
+- (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Integrate valgrind requests when cfg `valgrind_ct_test` is set
+
+### Changed
+
+- [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const qualifier of secret types constructors
```
jschneider-bensch:
```suggestion
- (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const qualifier of secret types constructors
```

crates/utils/secrets/src/mem_requests.rs
jschneider-bensch: Would you like to do this in a follow-up, perhaps? Or would you rather not expose these as part of the public API of `libcrux-secrets`?

crates/utils/secrets/src/int.rs
```diff
@@ -146,118 +146,116 @@ pub trait Select {
...
+        ($ty:ty, $secret_ty:ident, $mask_fn:ident, $cast:path) => {
             impl Select for [$ty] {
                 fn select(&mut self, other: &Self, selector: crate::U8) {
-                    let mask = ($is_non_zero(selector.declassify()) as $ty).wrapping_sub(1);
+                    let mask: $secret_ty = $cast($mask_fn(selector));
                     for i in 0..self.len() {
-                        self[i] = (self[i] & mask) | (other[i] & !mask);
+                        // if selector == 0, then mask is 0b111..11 and we select self[i],
+                        // otherwise mask is 0b000..00 and we select other[i]
+                        self[i] = ((mask & self[i]) | (!mask & other[i])).declassify();
```
jschneider-bensch, line 195: I think all the uses of `select` that would be relevant for us have `self.len() == other.len()`, so we should also assert that. Should also document  these requirements on the trait impls.

jschneider-bensch (approved):

crates/utils/secrets/src/mem_requests.rs
jschneider-bensch: Ugh, sorry, failed to realized this is what you did in #1457.

crates/utils/secrets/src/int.rs
robinhundt: I've added assert_eqs for all impls. I added a documentation note to the traits instead of the impls, as those are more visible. rust-analyzer also shows the docs for the trait when hovering over a usage of the trait's methods.

jschneider-bensch: Thanks! We implement the traits also on some basic type, but it should be clear from context that the comment covers every implementor that would have a `.len()` function.

PR #1447, opened 2026-05-19
[ML-DSA] Add valgrind testing for Key Generation

### LINKED ISSUE #1443: [ctgrind/ml-dsa] Testing for key generation

jschneider-bensch:
While the key generation declassifications are partly in platform-dependent code, this does not yet ensure #1448, so we should take care that all implementations are covered here: I added classifcations for portable and AVX2 and the under-documented NEON implementation uses the portable code for the declassifications in question.  

resolves #1443 

[skip changelog]

### DIFF
--- a/libcrux-ml-dsa/src/simd/avx2/rejection_sample/less_than_eta.rs
+++ b/libcrux-ml-dsa/src/simd/avx2/rejection_sample/less_than_eta.rs
@@ -61,6 +61,10 @@ pub(crate) fn sample<const ETA: usize>(input: &[u8], output: &mut [i32]) -> usiz
     // - mldsa-native: https://github.com/pq-code-package/mldsa-native/blob/0591fe06832418f8a320d5a1533327df063185cf/dev/x86_64/meta.h#L130
     // - Dilithium reference implementation: https://github.com/pq-crystals/dilithium/blob/6e00625c5b29f516c6de973fe2ee2fbb150973f9/avx2/rejsample.c#L312
     //
+    // It should be noted that the declassification of the comparison
+    // result here does not (and must not) declassify the sampled
+    // coefficients.
+    //
     // While we do not consider protection against power
     // side-channels in scope for libcrux, at this point, a
     // hardened implementation of sampling mod p can be found in
--- a/libcrux-ml-dsa/src/simd/avx2/rejection_sample/less_than_eta.rs
+++ b/libcrux-ml-dsa/src/simd/avx2/rejection_sample/less_than_eta.rs
@@ -1,6 +1,9 @@
-use crate::simd::avx2::{encoding, rejection_sample::shuffle_table::SHUFFLE_TABLE, Eta};
-
 use libcrux_intrinsics::avx2::*;
+
+use crate::{
+    ct_test::ct_declassify,
+    simd::avx2::{encoding, rejection_sample::shuffle_table::SHUFFLE_TABLE, Eta},
+};
 
 // TODO: This code seems to slow the implementation down, but stabilizes
 // benchmarks. Revisit this once the other functions are vectorized.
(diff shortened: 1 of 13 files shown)


libcrux-ml-dsa/src/simd/avx2/rejection_sample/less_than_eta.rs
```diff
@@ -38,6 +41,34 @@ pub(crate) fn sample<const ETA: usize>(input: &[u8], output: &mut [i32]) -> usiz
...
+    // - mldsa-native: https://github.com/pq-code-package/mldsa-native/blob/0591fe06832418f8a320d5a1533327df063185cf/dev/x86_64/meta.h#L130
+    // - Dilithium reference implementation: https://github.com/pq-crystals/dilithium/blob/6e00625c5b29f516c6de973fe2ee2fbb150973f9/avx2/rejsample.c#L312
+    //
+    // While we do not consider protection against power
+    // side-channels in scope for libcrux, at this point, a
+    // hardened implementation of sampling mod p can be found in
+    // Azouaoui et al, "Levelling Dilithium Against Leakage",
+    // section 4.4
+    // (https://csrc.nist.gov/csrc/media/Events/2022/fourth-pqc-standardization-conference/documents/papers/leveling-dilithium-against-leakage-pqc2022.pdf).
+    ct_declassify(&compare_with_interval_boundary);
```
robinhundt, line 74: One concern I have, is that mldsa-native performs the declassify higher in the call stack, above the rejection sampling loop, and then classifies the result again. Here the declassify is performed deeper in the sampling implementation on some intermediate value.

What is not clear to me:
- Which operation is actually the one whose timing depends on `compare_with_interval_boundary`? Is it already the index into the `SHUFFLE_TABLE`? Or only the branch in the sampling loop higher in the call stack?
- Does this declassify only what is needed, or are there bits after the sampling which should be classified but are now declassified? I.e. should there be an accompanying classify somewhere like in the mldsa-native implementation?

franziskuskiefer (approved):
leaving an r+ here already

libcrux-ml-dsa/src/simd/avx2/rejection_sample/less_than_eta.rs
robinhundt: Removing the declassify reveals that the following operations are flagged by valgrind:

- `mm_shuffle_epi8` in `less_than_eta.rs:101`
  - This is a bit weird, this operation should be constant time and I don't know why valgrind flags it.
- the indexing using `sampled_count` in `less_than_eta.rs:117`
- the `mm_storeu_si128_i32` in `less_than_eta.rs:116`
- the `if !done` ind `sample.rs:184`
- the indexing `&mut out[*sampled_coefficients..],` in `sample.rs:187`
- and equivalent places in other sampling methods with a different `eta` or the `sample_four_error_ring_elements` method


I've used valgrind's [vbits](https://docs.rs/crabgrind/latest/crabgrind/memcheck/fn.vbits.html) method to inspect the validity bits of `s1_s2` in the keygen in `ml_dsa_generic.rs` after they're sampled. The output confirms that with the current declassify, all vbits of `s1_s2` are undefined. This makes sense, as all coefficients in `less_than_eta.rs:sample` are derived from `potential_coefficients` which is not declassified.

So I think I resolved my concerns, but maybe that the declassify of `compare_with_interval_boundary` doesn't declassify the sampled coefficients could be added as a comment.

jschneider-bensch: Thanks for double checking! I've elaborated the comment in `less_than_eta.rs` to stress that the declassification does not and must not include the resulting sampled coefficients.

robinhundt (approved):

robinhundt:
Ah, I think this needs an updated extraction, because of the added `ct_{de}classify` functions which are empty in the extraction, but still present.

jschneider-bensch:
Indeed :+1:

jschneider-bensch:
I've updated the extraction for ML-DSA. This introduces one more `.c` file in ML-DSA, but I hope to address that with a Eurydice update soon. (Also created #1463 to keep an eye on extraction quality in the future.)

robinhundt:
@jschneider-bensch The extraction looks a bit weird to me. There are some changes in the monomorphizations of the ct_classify functions.

I'm curious why:
- some of them are marked as inline and other not
- There are changes to the definitions of these functions, but not to their uses. The extracted code that contains e.g. the `bool try_0_comp = try_0 < 15U` does not even contain a call to `ct_declassify` on the bool.
  - This is not really a problem, as ideally we don't want that call there anyway in the extraction, but I'm wondering why is it not there? Does the extraction tool realize that these functions are empty and doesn't even emit their calls at the source level?
  - If these functions are seemingly unused, why are they even emitted?

jschneider-bensch:
Whether they are inlined or not is a matter of the YAML config. For the other questions, tbh I'm not sure if e.g. charon will just extract whatever you configure, and Eurydice does a pass that removes no-op function calls, or if either of those things happen somewhere else.

PR #1453, opened 2026-05-20
Introduce AI-generated submissions policy in CONTRIBUTING.md

franziskuskiefer:
Added a policy on AI-generated submissions to ensure quality and human validation.

[skip changelog]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,3 +88,31 @@ You may use a [path dependency](https://doc.rust-lang.org/cargo/reference/specif
 If your pull request includes such a non-default dependency (e.g., a Git or path-based reference), **please mention it clearly in the PR description**.
 Using a non-[crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) dependency in a PR is acceptable for development. However, **before merging**, the dependency must be updated to point to a released version of `hax`.
 If a new `hax` release is required for your PR to be merged, please [open an issue on the hax repository](https://github.com/cryspen/hax/issues/new?title=Release%20request&body=For%20libcrux%20PR%20%23XX,%20I%20need%20a%20release%20of%20hax.) to request it.
+
+### Policy on AI-Generated Submissions
+
+We are supportive of responsible AI use in software development.
+AI can be a helpful tool, and we do not object to contributors using it as part of their workflow for writing code, drafting issues, or conducting security research.
+
+However, submissions to this project must meet a reasonable quality bar and remain suitable for human review.
+AI-generated content can be produced very quickly, but it frequently contains hallucinations, inaccurate claims, repetitive reports, or non-functional code.
+Reviewing these submissions requires significant maintainer time and attention.
+
+To keep our review process sustainable, we enforce the following guidelines:
+- **Human Validation is Required**: Please ensure that any AI-assisted issue, pull request, or security report has been carefully checked and verified by a human before it is submitted.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -88,3 +88,31 @@ You may use a [path dependency](https://doc.rust-lang.org/cargo/reference/specif
 If your pull request includes such a non-default dependency (e.g., a Git or path-based reference), **please mention it clearly in the PR description**.
 Using a non-[crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) dependency in a PR is acceptable for development. However, **before merging**, the dependency must be updated to point to a released version of `hax`.
 If a new `hax` release is required for your PR to be merged, please [open an issue on the hax repository](https://github.com/cryspen/hax/issues/new?title=Release%20request&body=For%20libcrux%20PR%20%23XX,%20I%20need%20a%20release%20of%20hax.) to request it.
+
+### Policy on AI-Generated Submissions
```
jschneider-bensch:
```suggestion
### Policy on LLM-Assisted Submissions
```

CONTRIBUTING.md
jschneider-bensch, line 1: Since this is about LLM use, I would prefer that we use that term, instead of "AI".

CONTRIBUTING.md
```diff
@@ -88,3 +88,31 @@ You may use a [path dependency](https://doc.rust-lang.org/cargo/reference/specif
...
+
+We are supportive of responsible AI use in software development.
+AI can be a helpful tool, and we do not object to contributors using it as part of their workflow for writing code, drafting issues, or conducting security research.
```
jschneider-bensch: In my opinion LLMs have many, many issues that should be considered besides their concrete utility for some task, so personally I wouldn't say I'm broadly supportive. At the same time I don't think we could ban them for contributors if we wanted to do that and that many of the issues are not to be solved at the level of individuals.
I would be happier if we omitted these two lines, unless we also include a broader discussion of the externalities that's not yet present here.

jschneider-bensch (dismissed):

CONTRIBUTING.md
franziskuskiefer: We can't omit this, as it's setting the tone for the rest. But you can suggest a different wording. Because we do not object and they are very useful tools for some things. We can't solve other issues around LLMs here and only need to define a policy for use in this repo.

CONTRIBUTING.md
```diff
@@ -88,3 +88,31 @@ You may use a [path dependency](https://doc.rust-lang.org/cargo/reference/specif
 If your pull request includes such a non-default dependency (e.g., a Git or path-based reference), **please mention it clearly in the PR description**.
 Using a non-[crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) dependency in a PR is acceptable for development. However, **before merging**, the dependency must be updated to point to a released version of `hax`.
 If a new `hax` release is required for your PR to be merged, please [open an issue on the hax repository](https://github.com/cryspen/hax/issues/new?title=Release%20request&body=For%20libcrux%20PR%20%23XX,%20I%20need%20a%20release%20of%20hax.) to request it.
+
+### Policy on LLM-Assisted Submissions
+
+We are supportive of responsible AI use in software development.
```
jschneider-bensch:
```suggestion
```

CONTRIBUTING.md
jschneider-bensch: I've put in an editing suggestion that leaves our non-objection and recognition of the usefulness for some things.

jschneider-bensch (approved):

PR #1454, opened 2026-05-21
sha3: fix footgun in runtime dispatched sha3 function

robinhundt:
Previously, you could call the `hash` (or aliased sha3) function with a constant LEN parameter not corresponding to the passed algorithm. In that case, the digest would be silently truncated.

I've removed the `#[cfg(not(eurydice))]` as the other sha3 functions also don't have it above their debug_asserts.


jschneider-bensch (approved):
That is a good catch!

jschneider-bensch:
It's not the extraction failure I was worried about, but looks like we don't derive `Copy` with `cfg(eurydice)`, so the debug assertion doesn't compile.

PR #1455, opened 2026-05-21
[kats/sha3] Move cavp crate into libcrux-kats and adapt sha3 tests

### LINKED ISSUE #1404: Integrate `cavp` crate in `libcrux-kats`
`cavp` doesn't really need to be a separate crate.

robinhundt:
This PR:
- moves the cavp parsing code into libcrux-kats
- adapts the sha3 cavp.rs test to use libcrux-kats (aided by claude code)
- tests sha3 avx2 and neon implementation against cavp test vectors (aided by claude code)
  - This uncovered a bug in the avx2 shake256 implementation that causes a panic on out buffers with a `length > 32` and `length % 8 != 0`. The fix is in a follow-up PR.
- removes the cavp crate

The avx2 and neon tests are automatically executed by the `ML-KEM - Build & Test` workflow.

Should removing cavp have a changelog entry in the project changelog? I'd argue that should only be for libcrux the library changes.
[skip changelog]


robinhundt:
I've stacked #1456 on this PR which fixes the bug and removes the `should_panic` annotation.

jschneider-bensch (approved):
Thanks!
I don't think we need a changelog entry here 👍

PR #1456, opened 2026-05-21
[sha3] avx2 shake256 fix out of bounds indexing

robinhundt:
In the rem8 > 0 branch, the index offset was mistakenly calculated
as `start + len - rem8`, not taking into account that previously
start is rebound to be `start + 32 * chunks`. This causes an out
of bounds index panic when an out vector has a length that is not
divisible by 8 and > 32 (as otherwise len == rem).


karthikbhargavan:
Wonderful catch!

robinhundt:
This also needs updated extractions.

robinhundt:
@jschneider-bensch This is ready for review now. The branch also contains my current version of the flake.nix which I used to regenerate the extractions. Using the flake, it is a simple `nix run .#ml-kem-extract` and ML-KEM is extracted using the correct versions of eurydice, charon, etc. I've also fixed a small bug in how the `FSTAR_REV` variable is set in some of the extract scripts, which resulted in the `F*: unset` output.

crates/algorithms/sha3/src/simd/avx2.rs
jschneider-bensch: Perhaps we should avoid this shadowing and rename this to `offset_rem` and `offset` below to `offset_rem8`?

jschneider-bensch (approved):

crates/algorithms/sha3/src/simd/avx2.rs
robinhundt: Yes, I like that better. Less likely to cause an issue in the future.

PR #1457, opened 2026-05-21
[ML-DSA]: use libcrux-secrets ct_de-/classify functions

robinhundt:
In #1446 we integrated valgrind client requests into libcrux-secrets. This PR replaces the valgrind `ct_{de}classify` functions added to libcrux-ml-dsa in #1435 and #1447 and replaces them by the (now public) functions of libcrux-secrets.

I've also had to change the configuration of the c and f* extractions of ml-dsa due to the new dependency. I managed to get them working with the help of claude, but a closer look at whether these are correct/idiomatic would be good.

To run libcrux-ml-dsa with valgrind, it needs to be compiled with `RUSTFLAGS='--cfg valgrind_ct_test'` and executed with valgrind. Normal execution not under valgrind will panic.


robinhundt:
This should wait until #1447 is merged and then update those uses as well.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#1382](https://github.com/cryspen/libcrux/pull/1382): Add support for HMAC-DRBG
 - (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Integrate valgrind requests when cfg `valgrind_ct_test` is set
+- (libcrux-ml-dsa) [#1457](https://github.com/cryspen/libcrux/pull/1457): Add dependency on libcrux-secrets for optional valgrind integration
```
franziskuskiefer, line 25: And using secret integers in future I guess? iot has this already and we should transfer that over here at some point.

libcrux-ml-dsa/Cargo.toml
franziskuskiefer: Do we still need the direct dependency now or should this all go through the secrets crate? No strong opinion, just wondering.

franziskuskiefer (approved):
Thanks. lgtm with two nits/questions

libcrux-ml-dsa/Cargo.toml
robinhundt: Yes, good catch, that should be removed at it's not necessary anymore.

CHANGELOG.md
robinhundt: Yes, in the future the plan is to use libcrux-secrets for that as well. But I wouldn't mention that here in the changelog already, as at the moment, the secrets crate is only used for valgrind requests.

PR #1461, opened 2026-05-26
[secrets]: fix incorrect cmp on aarch64

### LINKED ISSUE #1459: [secrets] Swap/Select can produce wrong result on aarch64

robinhundt:
The previous implementation of select/swap for aarch64 used a compare instruction to check whether the selector is 0. This cmp instruction was used on a 32-bit register into which the 8-bit selector was loaded. The upper `[8..32]` bits of that register where thus unspecified.

If some of these upper bits were set (e.g. because the selector was derived from truncating a larger value using an as cast), the inline asm `cmp` would differ from what the rust code would expect from comparing the u8 selector.

This commit fixes this issue, by using `tst` with a mask that only tests the first 8 bits of the selector, ignoring the upper unspecified bits.

fixes #1459

### DIFF
--- a/crates/utils/secrets/src/int.rs
+++ b/crates/utils/secrets/src/int.rs
@@ -261,11 +261,15 @@ mod aarch64 {
 
     macro_rules! select64 {
         ($lhs:expr, $rhs:expr, $selector:expr) => {
-            // Using https://developer.arm.com/documentation/ddi0602/2021-12/Base-Instructions/CSEL--Conditional-Select-
+            // Using https://developer.arm.com/documentation/ddi0602/2026-03/Base-Instructions/TST--immediate---Test-bits--immediate---an-alias-of-ANDS--immediate--
+            // and https://developer.arm.com/documentation/ddi0602/2026-03/Base-Instructions/CSEL--Conditional-select-
             #[allow(unsafe_code)]
             unsafe {
+                // We use the `tst` instruction to only check the lower byte of the cond register.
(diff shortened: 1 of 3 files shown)


robinhundt:
It seems there are currently no users of libcrux-secret's `Swap` or `Select` traits, so the incorrect aarch64 implementation was not used in downstream crates in libcrux or libcrux-iot.

The ML-KEM implementations have their own portable constant time select implementations.

crates/utils/secrets/src/int.rs
```diff
@@ -427,8 +443,11 @@ mod select {
         [T]: Select,
     {
         let selector: u8 = rng.random::<u8>() & 1;
-        // XXX: Setting `selector` as follows will break the selection in release mode on `aarch64`.
-        // let selector = if selector { 0 } else { rng.next_u32() as u8 };
+        let selector = if selector == 0 {
```
jschneider-bensch, line 446: It's good to know that this was the cause of the breakage, but looking at it now, the previously commented code looks redundant to the line above and I can't remember what the idea was here. 🤔

jschneider-bensch (approved):
This is a great catch!

crates/utils/secrets/src/int.rs
robinhundt: Yes, I was also wondering why this way of sampling the selector was even tried :sweat_smile: But it's good that this was still there, otherwise we wouldn't have found the issue.

PR #1462, opened 2026-05-26
[secrets] More robust casts instead of transmutes

robinhundt:
libcrux-secrets contained transmutes of the form `transmute::<&[T], &[Secret<T>]>`. This is *technically* UB, as the layout of slice fat pointers is not defined (in practice it is always `(ptr, len)` and this is unlikely to change and might be guaranteed in the future, but not at the moment)[^refs]. 

The first commit changes the relevant impls to use the `slice::from_raw_parts` functions. This could also be done via ptr casts, but I find the function a bit more clear.

The second commit changes the correct transmutes to more robust ptr casts.

[^refs]: See e.g. [here](https://lore.kernel.org/all/CAH5fLgibt_BQmOtkfEfo1=48zUeoWBJ-=u5gzw_a3X6Q7=aUSA@mail.gmail.com/) or [here](https://users.rust-lang.org/t/is-a-slice-of-transmutable-elements-transmutable-into-a-slice-of-transmuted-elements/79784/3).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const qualifier of secret types constructors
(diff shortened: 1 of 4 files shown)


jschneider-bensch (approved):
Nice!

PR #1467, opened 2026-06-01
[Combined Extraction] C Extract ML-DSA 44 & 87

### LINKED ISSUE #1328: Update Eurydice for combined header-only C extraction of ML-KEM & ML-DSA

jschneider-bensch:
[skip changelog]


jschneider-bensch:
Should add tests for the new ML-DSA variants as well.

jschneider-bensch:
I've added a C code extraction in addition to the header only extraction and I've also added KAT tests for ML-DSA 44 & 87.

franziskuskiefer:
We merged #1554 instead.

PR #1470, opened 2026-06-03
Workspace: exclude sys

franziskuskiefer:
This PR drops some unused sys crates from the workspace such that we can build and test the entire workspace. Now `cargo test --workspace` should work fine (using nextest would be recommended to speed it up).

There are additional fixes in ctgrind tests and the kats.

[skip changelog]


jschneider-bensch:
I've removed the `sys/hacl` build job that was failing CI in #1471, but I guess the `Cargo.toml` there and in `sys/pqclean` still refers to workspace variables.

jschneider-bensch (approved):

PR #1478, opened 2026-06-11
Upgrade to hax-lib 0.3.7 + F* v2026.03.24

karthikbhargavan:
## What this does
Upgrades the workspace from hax-lib 0.3.6 to 0.3.7 (`cargo-hax-v0.3.7`, rev `d8b5b3d`) and bumps the CI F\* version to v2026.03.24 (required by the 0.3.7 hax proof-libs), with the minimal F\* changes needed to keep the hax proof jobs green under 0.3.7. 

## Changes
- **Toolchain pin** — `hax-lib` → git tag `cargo-hax-v0.3.7` (`[workspace.dependencies]` + `[patch.crates-io]`), specs workspace pin; `hpke` disabled (hpke-rs pulls a crates.io libcrux pinned to 0.3.6). CI F\* set to v2026.03.24 in the mlkem/mldsa/sha3 hax workflows.
- **F\* changes(no impl-logic changes)** —
  - ml-kem `Spec.Utils`: 0.3.7 dropped `Core_models.Ops.Range.impl_index_range_slice` (→ explicit bounds) and turned `f_neg` into a `t_Neg` typeclass with no machine-int instance (→ re-provide `t_Neg i16`).
  - core-models `from_fn` / `BitVec::from_fn`: 0.3.7 emits two type implicits at call sites; add the absorbing slot to the extraction overrides.
  - ml-dsa: `Alloc.Vec.t_Vec` is now a distinct type (was `unfold = t_Slice`); index `to_vec()` ghost snapshots via `.as_slice()`, and rewrite one `cloop!` iterator (`inside_out_shuffle`) as an indexed loop.
- **CI tooling** — the proof scripts now run `make -k` and print an F\* error summary, so a single nightly run reports every failing module rather than stopping at the first.

**REGRESSION:** 22 functions (12 in ml-dsa, 10 in ml-kem) no longer verify under hax-lib 0.3.7 / F\* v2026.03.24 and are admitted here (`panic_free` where panic-freedom still holds, `lax` otherwise); **a subsequent PR re-proves them all with more comprehensive and stable proofs.** (The 0.3.7 hax proof-libs themselves require F\* v2026.03.24 — they don't elaborate under older F\* — so an older toolchain can't be used to sidestep these.)

<details>
<summary>The 22 admitted functions</summary>

**ml-dsa (12)** — `Polynomial.barrett_reduce`; `Simd.{Avx2,Portable}.Arithmetic.barrett_reduce_simd_unit`; `Simd.Avx2.Arithmetic.montgomery_multiply_aux`; `Simd.Avx2.Encoding.{Commitment.serialize, T0.serialize, T1.serialize, Error.serialize_when_eta_is_2, Error.serialize_when_eta_is_4, Gamma1.serialize_when_gamma1_is_2_pow_17, Gamma1.serialize_when_gamma1_is_2_pow_19}`; `Simd.Avx2.Invntt.invert_ntt_at_layer_3`.

**ml-kem (10)** — `Mlkem{512,768,1024}.Rand.{generate_key_pair, encapsulate}`; `Vector.Portable.Vector_type.from_i16_array`; `Vector.Avx2.Arithmetic.to_unsigned_representative`; `Polynomial.subtract_reduce`; `Ind_cca.Unpacked.generate_keypair`.

Causes are a mix of 0.3.7 SMT "incomplete quantifiers" regressions, z3 timeouts (up to ~5 min/query), and the changed `CryptoRng`/`Rand_core` trait model.
</details>


jschneider-bensch:
It might not be necessary to remove the hpke dependency here with `patch.crates-io` section for hax-lib. If possible I would prefer we keep hpke in the repo. (Sooner or later we'll have to have some pre-releases on both sides to get to a releasable state for libcrux and hpke.)

karthikbhargavan:
> It might not be necessary to remove the hpke dependency here with `patch.crates-io` section for hax-lib. If possible I would prefer we keep hpke in the repo. (Sooner or later we'll have to have some pre-releases on both sides to get to a releasable state for libcrux and hpke.)

Should be fixed now. This now fixes the Makefile to find the right hax-lib when verifying the code, so having multiple hax-libs in the environment is no longer a disaster.

jschneider-bensch:
I've taken some of the commits from here and rebased them onto #1505 for #1506. The changes in #1505 mean that we don't have to touch the F* Makefile or `get-hax-ref.yml` workflow. Unfortunately, we can't release crates with path or git dependencies on hax-lib, although I think it would be a good thing for the Makefile not to go through every single dependency looking for hax-lib.

jschneider-bensch:
Closing in favor of #1506.

PR #1480, opened 2026-06-16
Specs for ML-KEM

abentkamp:
This PR adds a hacspec-style reference implementation for ML-KEM. It includes cross-tests with the actual ML-KEM implementation, an F* and a Lean extraction. The PR builds on top of the SHA3-spec PR https://github.com/cryspen/libcrux/pull/1399.

The PR is structured into four clean commits:
* just the spec
* cross-spec tests
* F*
* Lean

This is the basis for the Lean verification of ML-KEM in `libcrux-iot`.


CHANGELOG.md
jschneider-bensch, line 1: No need to add a changelog entry in the libcrux crate for this.

specs/sha3/hax_aeneas.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 
-HAX_VERSION = "1f85fc13b9967080cc657863e2000ba5d4aa8647"
+HAX_VERSION = "ffdf432705d409b62ec025d253a340234b59766f"
```
jschneider-bensch, line 10: The revision in the Readme also needs an update then, right?

specs/ml-kem/src/ind_cca.rs
```diff
@@ -0,0 +1,644 @@
...
+            && DK_PKE_SIZE == RANK * BYTES_PER_RING_ELEMENT
+            && DK_SIZE == DK_PKE_SIZE + EK_SIZE + H_DIGEST_SIZE + 32
+    );
+    let (ek, dk_pke) = ind_cpa::generate_keypair::<RANK, EK_SIZE, DK_PKE_SIZE>(params, d)?;
+
+    // dk ← (dkₚₖₑ ‖ ek ‖ H(ek) ‖ z)
+    let mut dk = [0u8; DK_SIZE];
+    dk[..DK_PKE_SIZE].copy_from_slice(&dk_pke);
+    dk[DK_PKE_SIZE..DK_PKE_SIZE + EK_SIZE].copy_from_slice(&ek);
+    dk[DK_PKE_SIZE + EK_SIZE..DK_PKE_SIZE + EK_SIZE + H_DIGEST_SIZE].copy_from_slice(&H(&ek));
```
jschneider-bensch, line 48: The array bounds calculation here is a bit hard to read. Would it be possible to use a running offset instead? I.e.
```rust
let mut offset = 0;
dk[..DK_PKE_SIZE].copy_from_slice(...);
offset += DK_PKE_SIZE;
dk[offset..offset + EK_SIZE].copy_from_slice(...);
offset += EK_SIZE;
...
```

specs/ml-kem/src/ind_cca.rs
```diff
@@ -0,0 +1,644 @@
...
+    let hashed = G(&to_hash);
+    let (shared_secret, pseudorandomness) = hashed.split_at(32);
+
+    let r: [u8; 32] = pseudorandomness[..32].try_into().unwrap();
+
+    // c ← K-PKE.Encrypt(ek, m, r)
+    let c = ind_cpa::encrypt::<RANK, U_SIZE, V_SIZE, CT_SIZE>(params, ek, m, &r)?;
+
+    let mut k = [0u8; 32];
+    k.copy_from_slice(shared_secret);
```
jschneider-bensch, line 104: Why not `try_into()` this, like the randomness above?

specs/ml-kem/src/ind_cca.rs
```diff
@@ -0,0 +1,644 @@
...
+            && ek.len() == RANK * BYTES_PER_RING_ELEMENT + 32
+    );
+    // (K, r) ← G(m ‖ H(ek))
+    let mut to_hash = [0u8; 64];
+    to_hash[..32].copy_from_slice(m);
+    to_hash[32..].copy_from_slice(&H(ek));
+    let hashed = G(&to_hash);
+    let (shared_secret, pseudorandomness) = hashed.split_at(32);
+
+    let r: [u8; 32] = pseudorandomness[..32].try_into().unwrap();
```
jschneider-bensch, line 98:
```suggestion
    let r: [u8; 32] = pseudorandomness[..32].try_into().expect("G's digest size is 64 and pseudoranomness is the trailing 32 byte slice of that");
```

specs/ml-kem/src/ind_cca.rs
```diff
@@ -0,0 +1,644 @@
...
+        EK_SIZE == RANK * BYTES_PER_RING_ELEMENT + 32
+            && DK_PKE_SIZE == RANK * BYTES_PER_RING_ELEMENT
+            && DK_SIZE == DK_PKE_SIZE + EK_SIZE + H_DIGEST_SIZE + 32
+            && U_SIZE == (RANK * COEFFICIENTS_IN_RING_ELEMENT * params.du) / 8
+            && V_SIZE == (COEFFICIENTS_IN_RING_ELEMENT * params.dv) / 8
+            && CT_SIZE == U_SIZE + V_SIZE
+            && J_INPUT_SIZE == 32 + CT_SIZE
+    );
+    // dkₚₖₑ ← dk[0 : 384k]
+    let dk_pke = &dk[..DK_PKE_SIZE];
```
jschneider-bensch, line 164: Again, I'd find a running offset calculation easier to follow.

specs/ml-kem/src/ind_cca.rs
```diff
@@ -0,0 +1,644 @@
...
+    j_input[..32].copy_from_slice(z);
+    j_input[32..].copy_from_slice(c);
+    let rejection_shared_secret: [u8; 32] = J(&j_input);
+
+    // c′ ← K-PKE.Encrypt(ekₚₖₑ, m′, r′)
+    let c_prime =
+        ind_cpa::encrypt::<RANK, U_SIZE, V_SIZE, CT_SIZE>(params, ek, &m_prime, &r_prime)?;
+
+    // if c ≠ c′ then K′ ← K̃
+    if c[..] == c_prime[..] {
```
jschneider-bensch, line 198: In a production implementation you'd have to use a constant time comparison here, but of course this is a spec. Still, I would like to prevent any misunderstandings that you could use this implementation for anything real. Could you add a short Readme with a disclaimer that strongly discourages anyone from using the implementation for anything other than what we're doing with it?

jschneider-bensch (commented):
Leaving some comments already before the rest of the review.

CHANGELOG.md
abentkamp: Reverted.

specs/sha3/hax_aeneas.py
abentkamp: The Readme didn't exist because I hadn't the latest changes of the other PR. I merged main to get them, and updated the Readme.

specs/ml-kem/src/ind_cca.rs
abentkamp: Shouldn't be a problem, but there is a good chance that this will require fixes in our proof. Can we postpone this one?

jschneider-bensch: It's okay to postpone. It's just a style change, after all.

specs/ml-kem/src/ind_cca.rs
abentkamp: This looks weird indeed. Still, I'd like to postpone proof-altering changes.

specs/ml-kem/src/ind_cca.rs
abentkamp: Postpone?

specs/ml-kem/src/ind_cca.rs
abentkamp: Sure. Postpone?

specs/ml-kem/src/ind_cca.rs
abentkamp: Done.

crates/sys/platform/src/x86.rs
jschneider-bensch, line 1: These changes cause a rustfmt error.

CHANGELOG.md
jschneider-bensch, line 1: This file does not need to be changed at all.

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+
+/// Algorithm 13: K-PKE.KeyGen
+///
+/// Generates an encryption key and a corresponding decryption key.
+///
+/// ```plaintext
+/// Output: encryption key ekₚₖₑ ∈ 𝔹^{384k+32}.
+/// Output: decryption key dkₚₖₑ ∈ 𝔹^{384k}.
+///
+/// d ←$ B
```
jschneider-bensch, line 141: This is the `key_generation_seed` and is not sampled internally, but is an input.
```suggestion
Input: randomness d ∈ 𝔹^{32}
```

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+/// (ρ,σ) ← G(d)
```
jschneider-bensch, line 142: The rank is part of the input to G here.
```suggestion
/// (ρ,σ) ← G(d‖k)
```

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+/// N ← 0
+/// for (i ← 0; i < k; i++)
+///     for(j ← 0; j < k; j++)
+///         Â[i,j] ← SampleNTT(XOF(ρ, j, i))
```
jschneider-bensch, line 146: This line deviates from the presentation in the standard.
```suggestion
///         Â[i,j] ← SampleNTT(ρ‖j‖i)
```
While the current state is what actually happens at the beginning of `SampleNTT`, I prefer the presentation of the standard since it makes clear how (in particular in which order) the function arguments are to be combined  to one input for `SampleNTT`.

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+        EK_SIZE == RANK * BYTES_PER_RING_ELEMENT + 32
+            && DK_PKE_SIZE == RANK * BYTES_PER_RING_ELEMENT
+            && key_generation_seed.len() == 32
+    );
+
+    let (secret_as_ntt, t_as_ntt, _A_as_ntt, seed_for_A) =
+        generate_keypair_unpacked::<RANK>(params, key_generation_seed)?;
+
+    // ekₚₖₑ ← ByteEncode₁₂(t̂) ‖ ρ
+    let t_encoded: [u8; DK_PKE_SIZE] = serialize_secret_key::<RANK, DK_PKE_SIZE>(&t_as_ntt);
```
jschneider-bensch, line 185: I would prefer if `serialize_secret_key` had a generic name such as `byte_encode_12`. That matches the terminology of the standard and avoids confusion, since we're using it here to serialize a part of the public key.

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+/// Input: message m ∈ 𝔹^{32}.
+/// Input: encryption randomness r ∈ 𝔹^{32}.
+/// Output: ciphertext c ∈ 𝔹^{32(dᵤk + dᵥ)}.
+///
+/// N ← 0
+/// t̂ ← ByteDecode₁₂(ekₚₖₑ[0:384k])
+/// ρ ← ekₚₖₑ[384k: 384k + 32]
+/// for (i ← 0; i < k; i++)
+///     for(j ← 0; j < k; j++)
+///         Â[i,j] ← SampleNTT(XOF(ρ, j, i))
```
jschneider-bensch, line 211:
```suggestion
///         Â[i,j] ← SampleNTT(ρ‖j‖i)
```
see comment for key gen

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+///     end for
+/// end for
+/// for(i ← 0; i < k; i++)
+///     r[i] ← SamplePolyCBD_{η₁}(PRF_{η₁}(r,N))
```
jschneider-bensch, line 215: This line introduces a confusing shadowing of the name of the randomness `r`. In the standard, the values are sampled into a vector `y`.
```suggestion
///     y[i] ← SamplePolyCBD_{η₁}(PRF_{η₁}(r,N))
```

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+///     N ← N + 1
+/// end for
+/// for(i ← 0; i < k; i++)
+///     e₁[i] ← SamplePolyCBD_{η₂}(PRF_{η₂}(r,N))
+///     N ← N + 1
+/// end for
+/// e₂ ← SamplePolyCBD_{η₂}(PRF_{η₂}(r,N))
+/// r̂ ← NTT(r)
```
jschneider-bensch, line 223: I suggest renaming the hatted `r` to a hatted `y` as well, but can't easily type it out here.
```suggestion
/// r̂ ← NTT(y)
```

specs/ml-kem/src/matrix.rs
```diff
@@ -0,0 +1,183 @@
...
+        } else {
+            A_as_ntt[i][j] = sampled;
+        }
+    }
+    Ok(())
+}
+
+#[allow(non_snake_case)]
+#[hax_lib::requires(seed_for_A.len() == 32  && RANK <= 4)]
+pub fn sample_matrix_A<const RANK: usize>(
```
jschneider-bensch, line 110: There should be some documentation here, probably a description of the outer sampling loop in CPA keygen/encrypt and a description for the `transpose` parameter. (I realize it's there to match the implementation, but it's something that isn't part of the standard.)

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+    serialize::{
+        compress_then_serialize_message, compress_then_serialize_u, compress_then_serialize_v,
+        deserialize_ring_elements_reduced, deserialize_then_decompress_message,
+        deserialize_then_decompress_u, deserialize_then_decompress_v, serialize_secret_key,
+    },
+};
+
+/// Helper to sample a polynomial from CBD with dynamic eta.
+#[hax_lib::requires(eta == 2 || eta == 3)]
+fn sample_secret(eta: usize, prf_input: &[u8; 33]) -> Polynomial {
```
jschneider-bensch, line 18: Again, I'd prefer this to have name like `sample_poly_cbd` (already taken, unfortunately) because we use it to sample error values as well.

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+        U_SIZE == (RANK * COEFFICIENTS_IN_RING_ELEMENT * params.du) / 8
+            && V_SIZE == (COEFFICIENTS_IN_RING_ELEMENT * params.dv) / 8
+            && CT_SIZE == U_SIZE + V_SIZE
+            && randomness.len() == 32
+    );
+
+    // r[i] ← SamplePolyCBD_{η₁}(PRF_{η₁}(r,N)) ; r̂ ← NTT(r)
+    let r_as_ntt = sample_vector_cbd_then_ntt::<RANK>(params.eta1, randomness, 0);
+
+    // e₁[i] ← SamplePolyCBD_{η₂}(PRF_{η₂}(r,N))
```
jschneider-bensch, line 313: Please add a comment that documents that `N` is the domain separation parameter and has value `RANK` here because to the previous sampling of vector `r` (or `y` as the standard names it).

specs/ml-kem/src/ind_cpa.rs
```diff
@@ -0,0 +1,508 @@
...
+    let error_1 = sample_vector_cbd::<RANK>(params.eta2, randomness, RANK as u8);
+
+    // e₂ ← SamplePolyCBD_{η₂}(PRF_{η₂}(r,N))
+    let mut prf_input = [0u8; 33];
+    prf_input[..32].copy_from_slice(randomness);
+    prf_input[32] = (RANK * 2) as u8;
```
jschneider-bensch, line 319: Similar to above, please document that `prf_input[32]` is the domain separator and has value `RANK * 2` due to the two previous vector sampling computations.

specs/ml-kem/src/parameters.rs
```diff
@@ -0,0 +1,367 @@
...
+
+    pub const H_DIGEST_SIZE: usize = 32;
+
+    #[hax_lib::opaque]
+    pub fn H(input: &[u8]) -> [u8; H_DIGEST_SIZE] {
+        hacspec_sha3::sha3_256(input)
+    }
+
+    #[hax_lib::opaque]
+    pub fn PRF<const LEN: usize>(input: &[u8]) -> [u8; LEN] {
```
jschneider-bensch, line 284: It should be documented what the structure of the input is here and that `LEN` depends on the value of parameter `eta`.

specs/ml-kem/src/parameters.rs
```diff
@@ -0,0 +1,367 @@
...
+}
+
+/// Implicit-rejection hash input length: `32 + cpa_ciphertext_size`.
+#[hax_lib::requires(rank == 2 || rank == 3 || rank == 4)]
+pub const fn implicit_rejection_hash_input_size(rank: usize) -> usize {
+    32 + cpa_ciphertext_size(rank)
+}
+
+#[allow(non_snake_case)]
+pub mod hash_functions {
```
jschneider-bensch, line 270: It would be great to have the definitions for these functions from the standard as documentation here.

specs/ml-kem/src/parameters.rs
```diff
@@ -0,0 +1,367 @@
...
+
+/// An ML-KEM field element:
+/// - after reduction modulo FIELD_MODULUS, it is an integer in the range [0, FIELD_MODULUS - 1]
+/// - it is represented as a u16
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[cfg_attr(not(any(hax, hax_backend_lean)), derive(Debug))]
+#[hax_lib::attributes]
+pub struct FieldElement {
+    #[refine(val < FIELD_MODULUS)]
```
jschneider-bensch, line 307: Does this `refine` annotation ensure that every value of this type has `val < FIELD_MODULUS`? Sorry, not familiar with the attribute.

specs/ml-kem/src/parameters.rs
```diff
@@ -0,0 +1,367 @@
...
+        Self { val }
+    }
+
+    /// Reduce an arbitrary `i16` (e.g. an impl-side coefficient or
+    /// Montgomery-domain value) into a canonical `FieldElement` in
+    /// [0, FIELD_MODULUS).  Used by the impl→spec lift functions
+    /// (`Libcrux_ml_kem.Vector.to_spec_*_t`) to bridge the trait-layer
+    /// `i16` representation to the spec-layer `FieldElement` form.
+    pub const fn from_i16(v: i16) -> FieldElement {
+        let q = FIELD_MODULUS as i32;
```
jschneider-bensch, line 324: Is it necessary to extend to `i32` here? Are there tests or a proof of correctness for this function?

specs/ml-kem/src/parameters.rs
```diff
@@ -0,0 +1,367 @@
...
+        let r = (((v as i32) % q + q) % q) as u16;
+        FieldElement::new(r)
+    }
+
+    /// Addition in ℤ/q.
+    pub fn add(self, other: FieldElement) -> FieldElement {
+        FieldElement::new(((self.val as u32 + other.val as u32) % FIELD_MODULUS as u32) as u16)
```
jschneider-bensch, line 331: It shouldn't be necessary to cast to `u32` here if both operands are  `< FIELD_MODULUS`, right?

specs/ml-kem/src/parameters.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    /// Subtraction in ℤ/q.  Adding q avoids unsigned underflow before reducing.
+    pub fn sub(self, other: FieldElement) -> FieldElement {
+        FieldElement::new(
+            ((self.val as u32 + FIELD_MODULUS as u32 - other.val as u32) % FIELD_MODULUS as u32)
```
jschneider-bensch, line 337: Similar to `add` the casts to `u32` seem unnecessary.

specs/ml-kem/src/ntt.rs
```diff
@@ -0,0 +1,687 @@
...
+/// Use the Cooley–Tukey butterfly to compute an in-place NTT representation
+/// of a `Polynomial`.
+///
+/// Given a `Polynomial` `f`, the NTT representation `f^` is:
+///
+/// ```plaintext
+/// f^ := (f mod(X² - ζ^(2*BitRev₇(0) + 1), ..., f mod (X² − ζ^(2·BitRev₇(127) + 1))
+/// ```
+///
+/// This function implements <strong>Algorithm 8</strong> of the NIST FIPS 203 standard, which
```
jschneider-bensch, line 38: Perhaps this comment should be placed above function `ntt` below.
```suggestion
/// This array holds the precomputed zeta values used in <strong>Algorithm 9</strong> of the NIST FIPS 203 standard, which
```

specs/ml-kem/src/ntt.rs
```diff
@@ -0,0 +1,687 @@
...
+    let p = ntt_layer(p, 3);
+    let p = ntt_layer(p, 2);
+    let p = ntt_layer(p, 1);
+    p
+}
+
+/// Compute the product of two `KyberBinomial`s with respect to the
+/// modulus `X² - zeta`.
+///
+/// This function implements <strong>Algorithm 11</strong> of the NIST FIPS 203 standard, which
```
jschneider-bensch, line 276: `BaseCaseMultiply` is a different algorithm.
```suggestion
/// This function implements <strong>Algorithm 12</strong> of the NIST FIPS 203 standard, which
```

specs/ml-kem/src/matrix.rs
```diff
@@ -0,0 +1,183 @@
...
+
+/// N.B.: According to the NIST FIPS 203 standard (Page 9, Line 519), a matrix is
+/// a set of column vectors.
+///
+/// The NIST FIPS 203 standard can be found at
+/// <https://csrc.nist.gov/pubs/fips/203/ipd>.
+///
+pub fn add_polynomials(p1: &Polynomial, p2: &Polynomial) -> Polynomial {
+    createi(|j| {
+        FieldElement::new(((p1[j].val as u32 + p2[j].val as u32) % FIELD_MODULUS as u32) as u16)
```
jschneider-bensch, line 16: This should use the addition defined on `FieldElement`.

specs/ml-kem/src/matrix.rs
```diff
@@ -0,0 +1,183 @@
...
+    })
+}
+
+pub fn sub_polynomials(p1: &Polynomial, p2: &Polynomial) -> Polynomial {
+    createi(|j| {
+        FieldElement::new(
+            ((p1[j].val as u32 + FIELD_MODULUS as u32 - p2[j].val as u32) % FIELD_MODULUS as u32)
```
jschneider-bensch, line 23: This should use the subtraction defined on `FieldElement`.

specs/ml-kem/src/parameters.rs
```diff
@@ -0,0 +1,367 @@
...
+#[cfg_attr(not(any(hax, hax_backend_lean)), derive(Debug))]
+#[hax_lib::attributes]
+pub struct FieldElement {
+    #[refine(val < FIELD_MODULUS)]
+    pub val: u16,
+}
+
+#[hax_lib::attributes]
+impl FieldElement {
+    #[hax_lib::requires(val < FIELD_MODULUS)]
```
jschneider-bensch, line 313: Why enforce this with a pre-condition here, instead of doing the reduction `% FIELD_MODULUS` in the constructor for `FieldElement`? The bound could be post-condition instead and making the `val` field private instead of `pub` would ensure that no out-of-bounds field element can be created in the Rust domain already as long as the operations preserve this too.

specs/ml-kem/src/serialize.rs
```diff
@@ -0,0 +1,612 @@
...
+    })
+}
+
+/// Deserialize, decompress, then NTT-transform u from ciphertext bytes.
+/// Composes [`deserialize_then_decompress_u`] with [`vector_ntt`].  Used
+/// by `ind_cpa::decrypt` and matches the libcrux-impl
+/// `deserialize_then_decompress_u` function (which fuses the NTT into
+/// the per-element decompress loop).
+#[hax_lib::requires(RANK <= 4 && (du == 10 || du == 11) && ciphertext.len() == (RANK * COEFFICIENTS_IN_RING_ELEMENT * du) / 8)]
+pub fn deserialize_then_decompress_u_then_ntt<const RANK: usize>(
```
jschneider-bensch, line 370: It looks like this function is not used in `ind_cpa::decrypt_unpacked`?

specs/ml-kem/src/ind_cca.rs
```diff
@@ -0,0 +1,644 @@
...
+        ind_cpa::generate_keypair_unpacked::<RANK>(params, d)?;
+
+    // Libcrux stores A in transposed form on `IndCpaPublicKeyUnpacked.A`.
+    let m_A: Matrix<RANK> = transpose(&A_as_ntt);
+
+    // public_key_hash = H(serialize_public_key(t_as_ntt, seed_for_A))
+    let ek: [u8; EK_SIZE] = serialize_public_key::<RANK, EK_SIZE>(&t_as_ntt, &seed_for_A);
+    let public_key_hash: [u8; 32] = H(&ek);
+
+    let mut implicit_rejection_value = [0u8; 32];
```
jschneider-bensch, line 392: This seems unnecessary, since `z` is already a reference to a `[u8;32]` and could be cloned below instead.

specs/ml-kem/Readme.md
jschneider-bensch, line 1: Can you add a section on the F* extraction as well? I think there is no script to drive it and it's not clear what it should prove.

specs/ml-kem/Readme.md
```diff
@@ -0,0 +1,24 @@
...
+
+This is a hacspec-style Rust implementation of ML-KEM, closely following FIPS 203. Its purpose
+is to serve as a reference implementation for verifying functional correctness of more efficient
+implementations.
+
+**Do not use this implementation for other applications! Besides being slow, it may be vulnerable to side-channel attacks.**
+
+## Extraction via HAX
+
+### Lean
```
jschneider-bensch, line 11: Is there anything that is proven in lean about this spec on its own, or is the extraction just used for the functional correctness proof of the impl?

jschneider-bensch (changes requested):
I've left more comments, mostly around documentation and naming of functions. Is renaming a function okay in terms of updating the proofs, would you say?

- The other set of suggestions concern the use of `FieldElement` and its operations, where we could use the Rust type system more effectively to maintain the boundedness invariant, instead of mixing up the type's own operations with direct field accesses and allowing construction of the type independent of its `new` method.

- I think some of the documentation issues are inherited from the impl, where some documentation has outdated references that refer to the initial public draft of FIPS203, instead of the final version. While the documentation deviated from the published standard in some ways I found no issues with the implementation code.

- Can you add test vector testing for the spec here, using `libcrux-kats`, so we have the same test vector coverage here as on the impl side?

specs/ml-kem/src/ind_cca.rs
jschneider-bensch: I'm okay with postponing these changes, but could you make a commit with them anyway and afterwards revert that? This way we don't lose track of the changes that we should make when we get the chance.

specs/ml-kem/Readme.md
abentkamp: There aren't any Lean proofs here, only definitions.

jschneider-bensch: Got it :+1: Can you add a line to the Readme then, that there are no proofs here, only definitions that are inputs in proofs elsewhere?

PR #1484, opened 2026-06-18
[secrets] seal scalar trait and synchronize De-/Classify trait impls for public/secret types

### LINKED ISSUE #1483: [secrets] var-time trait implementations can introduce secret-dependent behavior for `Secret<T>` types
In libcrux-secrets, we currently implement certain traits like `Add` or `Sub` for `Secret<T>` if `T` implements the relevant trait. 

https://github.com/cryspen/libcrux/blob/ffa27958d519929e1bf82fad5d33418eb58374b8/crates/utils/secrets/src/int/secret_integers.rs#L87-L92

For the standard library types we care about like `u8`, we know that their addition and subtraction is constant time. However, as the API of libcrux-secrets is at the moment, a user could use `Secret<TheirType>` with a secret dependent `Add` implementation.

```rust
use std::ops::Add;

struct TheirType(u64);

impl Add for TheirType {
    type Output = TheirType;
    
    fn add(self, other: TheirType) -> TheirType {
        if self.0 == 0 {
            other
        } else {
            self + other
        }
    }
}
```
([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cc7921467517bd95af215e625a0202e6), note that this would likely be optimized to a simple addition which would be constant time, but there could be more complex implementations where that would not be the case)

If a user then constructs a `Secret<TheirType>`, the secret type would exhibit secret dependent addition.

While this concrete example might seem a bit constructed, I see at least one potential type of usage where this could occur: when using arbitrary precision big integers. I could use the `Secret` wrapper with a GMP [`Integer`](https://docs.rs/rug/latest/rug/struct.Integer.html) whose addition timing depends on the variable number of limbs of the operands. 

## Possible solutions
1. The easy solution is restricting all operations on `Secret<T>` types to only those `T` that implement a sealed trait and then implement that for all the primitive types we're interested in. There is already the [`Scalar`](https://github.com/cryspen/libcrux/blob/ffa27958d519929e1bf82fad5d33418eb58374b8/crates/utils/secrets/src/traits.rs#L38) trait which is implemented for primitve integers, but it is not sealed and not used in all bounds.
2. Have `CtAdd`, `CtSub`, etc. traits that must not be implemented for types where those operations are not constant time.

robinhundt:
This seals the `Scalar` trait and makes its usage more consistent. This resolves the potential issue that there could be trait implementations of e.g. `Add` for `Secret<T>` that are not actually constant time as described in #1483.

However, adding the `Scalar` trait bound consistently caused issues in downstream crates, due to the different `De-/Classify` impls depending on the `check-secret-independence` feature being present or not. This PR also adds consistent impls of these traits, so that any code that compiles with the feature present, also compiles without it (this was previously not the case).

While working on this code, I also thought about whether there might be cases where the multiplication of larger integers is not constant time. It seems there are older ARM cores, like the ARM7TDMI, that have an early termination multiplier. It might be worth a note somewhere that we're not targeting devices with non-constant time multipliers.

resolves #1483


crates/utils/secrets/src/int/classify_public.rs
jschneider-bensch, line 1: Thanks for documenting when compiling with and without `check-secret-independence` might not work the same!

crates/utils/secrets/src/traits.rs
```diff
@@ -139,3 +139,48 @@ impl Scalar for core::arch::aarch64::uint64x1_t {}
...
+    macro_rules! impl_sealed {
+        ($($type:ty), *) => {
+            $(
+                impl Sealed for $type {}
+            )*
+        };
+    }
+
+    impl_sealed!(u8, u16, u32, u64, u128, i8, i16, i32, i64, i128);
+    #[cfg(target_arch = "x86")]
```
jschneider-bensch, line 158: I think we need some special handling of `cfg(hax)` here if we want to keep the door open to hax extraction with `check-secret-independence` enabled, since we don't use the `core::arch` types in 
`libcrux-intrinsics` when we build with `cfg(hax)`, e.g. for `x86_64`:

https://github.com/cryspen/libcrux/blob/8470b57c228f12c81b27eb5abcba298729a72826/crates/utils/intrinsics/src/avx2.rs#L3-L9

https://github.com/cryspen/libcrux/blob/8470b57c228f12c81b27eb5abcba298729a72826/crates/utils/core-models/src/core_arch/x86.rs#L158-L161

We might not ever want to do this, but could be worth documenting that there expected issues there.

jschneider-bensch (approved):
Thanks for taking the opportunity to fix this annoying wrinkle of the mismatch in impls! I distinctly remember wrestling with secrets to work around this on IoT. The resulting breakage will only mean we can straighten things out there.

Regarding variable time multiplication: Definitely there are platforms where the assumptions made by `libcrux-secrets` do not apply and we should document that we assume multiplication to be safe. There is an overview of potentially problematic platforms e.g. [here](https://www.bearssl.org/ctmul.html).

crates/utils/secrets/src/traits.rs
robinhundt: I've added a comment that there would likely be an incompatibility with hax extraction with the impls for the SIMD types.

robinhundt:
@jschneider-bensch I noticed that https://github.com/cryspen/libcrux/commit/b9053271b70b7a1b89b1261d8345d05298c74fae (I think from the evit sync) removed the no-std job in the secrets workflow. This broke the secrets workflow as it was still referred to in the status check.

I've added back the no-std job and fixed it so that it actually executes for secrets.

crates/utils/secrets/src/traits.rs
jschneider-bensch: Thanks!

jschneider-bensch:
> @jschneider-bensch I noticed that [b905327](https://github.com/cryspen/libcrux/commit/b9053271b70b7a1b89b1261d8345d05298c74fae) (I think from the evit sync) removed the no-std job in the secrets workflow. This broke the secrets workflow as it was still referred to in the status check.
> 
> I've added back the no-std job and fixed it so that it actually executes for secrets.

Oh, thanks for fixing! The `no_std` job there was configured to run in the AES crate, which looked like an error to me and I had thus removed it, but I missed that it's part of the status check as well. It doesn't hurt to have a `no_std` check for secrets, but maybe we can run it twice, with and without `check-secret-independence`.

robinhundt:
I've added a check with the feature enabled.

robinhundt:
Todo: update c extractions...

robinhundt:
Updating the extractions was more annoying than expected. Karamel changed the layout of the include paths in https://github.com/FStarLang/karamel/commit/bfdc2c11d831d889d3587745e553191de421ec29#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52 .

This lead to the installed version I had of karamel, which was built from the nix flake, to always copy krml into `include/krml/krml`. This seems to be by design in karamel, but we're currently expecting krml to be in `include/krml`. The way CI is setup, it also copies `krml` into `include/krml` as that is also the path in the repository. It is only when installed, that the path is duplicated...

For now, I've simply added `--no-karamel_include` to my flake scripts for re-extracting.

combined_extraction/generated/combined_core.h
jschneider-bensch, line 1: It's kind of handy to see here for which array sizes we get monomorphizations.

jschneider-bensch (approved):
Looks good, thanks! :+1:

PR #1489, opened 2026-06-24
[secrets]: relax rhs bounds for shifting

robinhundt:
Crucially, this again allows shifting with a usize rhs.

This was incidentally forbidden by #1484, which required the rhs to implement `Scalar`, which is not implemented for `usize`.

[skip changelog]


jschneider-bensch (approved):
Ah, I missed that in #1484

PR #1493, opened 2026-07-01
[traits/sha3] Digest traits consume hasher on `finish` & provide Instantiable `libcrux-traits` incremental hashing API

### LINKED ISSUE #1492: [sha3] Remove/Fix `libcrux-traits` incremental APIs

### LINKED ISSUE #1496: [traits] digest traits are not misuse resistant
The current [digest traits](https://docs.rs/libcrux-traits/latest/libcrux_traits/digest/index.html) APIs are not misuse resistant. The following scenarios are possible:

### Update after finish

Because the `finish` methods take `&mut self` and not `self`, it is possible to call `update` after finish. For algorithms such as SHA-3, this is not defined[^blake], as you can't go back to absorbing when you're already squeezing the state. If in an implementation of a the traits you want to guard against this misuse, there are no perfect solutions (see #1493), only different compromises.

### Calling finish twice

It is possible call `finish` twice (again because of it taking `&mut`). It is unclear what the expected behavior should be. Should the same digest be returned? Should the state be reset internally before computing the second digest?

### Solution

I think the solution is obvious: `finish` must take `self` by value instead of `&mut self`. This solves both problems described above. Additionally, we might want a `finish_reset(&mut self) -> Digest` API, that gives you a digest and resets the state in the same operation[^rust-crypto-digest]. 

[^blake]: One exception being BLAKE-3, where it is defined to call `update -> finalize -> update -> finalize ...`
[^rust-crypto-digest]: Prior work is the [`digest::Digest::finalize_reset`](https://docs.rs/digest/0.11.3/digest/trait.Digest.html#tymethod.finalize_reset) method.

jschneider-bensch:
This PR updates the digest traits, so `finish` methods consume the hasher. It also updates implementations of these traits in `libcrux-sha2`, `libcrux-blake2`, `libcrux-hmac` and the initial implementation in `libcrux-sha3` that was the start of this PR.

Resolves #1496 

---
## Original PR description.

This PR provides an instantiable type under the `Sha3_XYXHasher` type aliases, which can be used for incremental hashing via `libcrux-traits`.

I've adapted test-code from the embedded-cal branch of `libcrux-iot` and also attempted to make the incremental hashing API somewhat safer than what is enforced by `libcrux-traits`, in that calling `update` on a hasher that has already finished results in a an error (`UpdateError::Unknown` is unfortunately the best error for this currently provided by `libcrux-traits`) unless the hasher is reset first. Repeatedly calling `finish` on a hasher will always copy out the digest that was computed on the first call to `finish`.

Resolves #1492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- (libcrux-traits, libcrux-sha3, libcrux-sha2, libcrux-blake2) [#1493](https://github.com/cryspen/libcrux/pull/1493): `finish` methods on digest traits consume the hasher
+- (libcrux-traits, libcrux-sha3, libcrux-sha2, libcrux-blake2, libcrux-hmac) [#1493](https://github.com/cryspen/libcrux/pull/1493): `finish` methods on digest traits consume the hasher
 - (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const qualifier of secret types constructors
 - (libcrux-secrets) [#1462](https://github.com/cryspen/libcrux/pull/1462): More robust casts instead of transmutes when checking secret independence
 - (libcrux-sha3) [1454](https://github.com/cryspen/libcrux/pull/1454): `debug_assert` that generic `LEN` matches algorithm in `hash` function
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- (libcrux-traits, libcrux-sha3, libcrux-sha2, libcrux-blake2, libcrux-hmac) [#1493](https://github.com/cryspen/libcrux/pull/1493): `finish` methods on digest traits consume the hasher
+- (libcrux-traits, libcrux-sha3, libcrux-sha2, libcrux-blake2) [#1493](https://github.com/cryspen/libcrux/pull/1493): `finish` methods on digest traits consume the hasher
 - (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const qualifier of secret types constructors
 - (libcrux-secrets) [#1462](https://github.com/cryspen/libcrux/pull/1462): More robust casts instead of transmutes when checking secret independence
 - (libcrux-sha3) [1454](https://github.com/cryspen/libcrux/pull/1454): `debug_assert` that generic `LEN` matches algorithm in `hash` function
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- (libcrux-sha3) [#1493](https://github.com/cryspen/libcrux/pull/1493): Provide instantiable `libcrux-traits` incremental hashing API
 - (libcrux-secrets) [#1460](https://github.com/cryspen/libcrux/issues/1460): Fix incorrect cmp in aarch64 select/swap implementation
 - (libcrux-sha3) [#1456](https://github.com/cryspen/libcrux/pull/1456): Fix out of bounds indexing in avx2 SHAKE-256 implementation
 
 ### Changed
 
+- (libcrux-traits, libcrux-sha3, libcrux-sha2, libcrux-blake2) [#1493](https://github.com/cryspen/libcrux/pull/1493): `finish` methods on digest traits consume the hasher
 - (libcrux-secrets) [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const qualifier of secret types constructors
 - (libcrux-secrets) [#1462](https://github.com/cryspen/libcrux/pull/1462): More robust casts instead of transmutes when checking secret independence
 - (libcrux-sha3) [1454](https://github.com/cryspen/libcrux/pull/1454): `debug_assert` that generic `LEN` matches algorithm in `hash` function
--- a/crates/algorithms/blake2/tests/blake2.rs
+++ b/crates/algorithms/blake2/tests/blake2.rs
@@ -208,17 +208,23 @@ fn test_blake2s() {
 
 #[test]
 fn test_digest_traits_2s() {
+    fn updated_hasher() -> Blake2sHasher<32> {
+        let mut hasher = Blake2sHasher::<32>::new();
+        hasher.update(b"this is a test").unwrap();
+        hasher
+    }
+
     let mut got_hash = [0; 32];
 
     // test unkeyed, with const key and digest len
     let expected_hash = b"\xf2\x01\x46\xc0\x54\xf9\xdd\x6b\x67\x64\xb6\xc0\x93\x57\xf7\xcd\x75\x51\xdf\xbc\xba\x54\x59\x72\xa4\xc8\x16\x6d\xf8\xaf\xde\x60";
-    let mut hasher = Blake2sHasher::<32>::new();
-    hasher.update(b"this is a test").unwrap();
+    let hasher = updated_hasher();
     hasher.finish(&mut got_hash);
 
     assert_eq!(&got_hash, expected_hash);
 
     let mut too_short = vec![0; 31];
+    let hasher = updated_hasher();
     let err = hasher.finish_slice(&mut too_short).unwrap_err();
     assert_eq!(
         err,
@@ -226,6 +232,7 @@ fn test_digest_traits_2s() {
     );
 
     let mut too_long = vec![0; 33];
+    let hasher = updated_hasher();
     let err = hasher.finish_slice(&mut too_long).unwrap_err();
     assert_eq!(
         err,
@@ -234,17 +241,23 @@ fn test_digest_traits_2s() {
 }
 #[test]
 fn test_digest_traits_2b() {
+    fn updated_hasher() -> Blake2bHasher<32> {
+        let mut hasher = Blake2bHasher::<32>::new();
+        hasher.update(b"this is a test").unwrap();
+        hasher
+    }
+
     let mut got_hash = [0; 32];
 
     // test unkeyed, with const key and digest len
     let expected_hash = b"\xe9\xed\x14\x1d\xf1\xce\xbf\xc8\x9e\x46\x6c\xe0\x89\xee\xdd\x4f\x12\x5a\xa7\x57\x15\x01\xa0\xaf\x87\x1f\xab\x60\x59\x71\x17\xb7";
-    let mut hasher = Blake2bHasher::<32>::new();
-    hasher.update(b"this is a test").unwrap();
+    let hasher = updated_hasher();
     hasher.finish(&mut got_hash);
 
     assert_eq!(&got_hash, expected_hash);
 
     let mut too_short = vec![0; 31];
+    let hasher = updated_hasher();
     let err = hasher.finish_slice(&mut too_short).unwrap_err();
     assert_eq!(
         err,
@@ -252,6 +265,7 @@ fn test_digest_traits_2b() {
     );
 
     let mut too_long = vec![0; 33];
+    let hasher = updated_hasher();
     let err = hasher.finish_slice(&mut too_long).unwrap_err();
     assert_eq!(
         err,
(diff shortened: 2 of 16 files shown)


crates/algorithms/sha3/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- [#XXX](https://github.com/cryspen/libcrux/pull/XXX): Provide instantiable `libcrux-traits` incremental hashing API
```
robinhundt:
```suggestion
- [#1493](https://github.com/cryspen/libcrux/pull/XXX): Provide instantiable `libcrux-traits` incremental hashing API
```

crates/algorithms/sha3/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
```
robinhundt, line 11: This change should also be included in the top level changelog for the library.

crates/algorithms/sha3/src/impl_digest_trait.rs
```diff
@@ -1,19 +1,86 @@
...
+#[doc(hidden)]
+/// Incremental hasher state
+///
+/// We implement the `libcrux-traits` traits `InitializeDigestState`
+/// and `DigestIncrementalState` generically for any `RATE`, but
+/// `libcrux_traits::digest::arrayref::DigestIncremental`,
+/// i.e. `finish`, is only implemented for the supported digest
+/// lengths.
+pub struct HasherState<const RATE: usize, const OUTLEN: usize> {
+    // We use a portable XOF state here, but AVX2 and NEON variants
```
robinhundt: I think avx2 and neon sha3 is also not compatible with the current libcrux-traits APIs, right? But I'm not sure whether it would be worth it abstracting over such hashers/xofs that can work on parallel state. As far as I'm aware, we only use that in ML-KEM/ML-DSA where we don't want to be generic over the hasher/xof I think.

crates/algorithms/sha3/src/impl_digest_trait.rs
```diff
@@ -1,19 +1,86 @@
...
+        }
+    }
+}
+
 macro_rules! impl_hash_traits {
-    ($type:ident, $hasher:ident, $len:expr, $method:expr) => {
+    ($type:ident, $hasher:ident, $len:expr, $rate:expr, $method:expr) => {
         #[doc = concat!("A struct that implements [`libcrux_traits::digest`] traits.")]
         #[doc = concat!("\n\n")]
         #[doc = concat!("[`",stringify!($hasher), "`] is a convenient hasher for this struct.")]
```
robinhundt: It is a bit weird that the hasher type alias does not contain this type here anymore. At least this comment is slightly wrong, as it is not a hasher "for this struct".

crates/algorithms/sha3/src/impl_digest_trait.rs
```diff
@@ -1,19 +1,86 @@
...
+    // is necessary until `squeeze` is implemented for `PARALLEL_LANES
+    // > 1`.
+    inner: KeccakXofState<1, RATE, u64>,
+    absorb_state: AbsorbState<OUTLEN>,
+}
+
+enum AbsorbState<const OUTLEN: usize> {
+    Absorbing,
+    Finished([u8; OUTLEN]),
+}
```
robinhundt: This really shows the limitation of the digest traits. In my opinion, the traits themselves should make it impossible to call finish twice (by taking self) and impossible to call absorb after calling finish (also by taking self in finish).

To still have the option of getting a digest and resetting the hasher, there could be a `finish_reset()` method that does both. This is exactly what the [Digest crate offers](https://docs.rs/sha3/latest/sha3/trait.Digest.html#tymethod.finalize_reset).

But without changing the traits, I think this is the best we can do in this PR. However, we should create an issue to track these shortcomings of our digest traits.

crates/algorithms/sha3/src/impl_digest_trait.rs
robinhundt: Thinking about it more, why did you choose to have the `HasherState` as a separate generic struct, instead of instantiating it here inside the macro and making what is currently the ZST $type the `HasherState` (similar as I did in iot)?

We'd have more code inside the macro, which I'm not a huge fan of, but we wouldn't need a generic doc(hidden) struct in that case and would also implement update just for defined rates.

crates/algorithms/sha3/src/impl_digest_trait.rs
```diff
@@ -1,19 +1,86 @@
...
+        *state = Self::IncrementalState::new();
+    }
+
+    fn update(state: &mut Self::IncrementalState, payload: &[u8]) -> Result<(), UpdateError> {
+        match state.absorb_state {
+            AbsorbState::Absorbing => {
+                state.inner.absorb(&[payload]);
+                Ok(())
+            }
+            AbsorbState::Finished(_) => Err(UpdateError::Unknown),
```
robinhundt: I'm not really happy with this solution and wonder whether panicking would not be the better choice here.

The potential problem I see is the following:
- `update(data)` is called on a hasher
- `finish()` is called on the hasher to get the digest
- the hasher is kept around, maybe it is inside a longer lived struct
- reset is **not** called
- later, `update(other_data)` is called, but the `UpdateError::Unknown` is not correctly handled (e.g. just ignored)
- `finish()` is called again, with the expectation that it provides a digest for `other_data`, but it returns the digest for `data`

If `update` would panic, this failure scenario could not occur, as it would be impossible to call update on a finished hasher. Of course, this is also less than ideal, as a panicking update method is also really unexpected and could lead to a DoS vulnerability...

robinhundt (commented):
Thanks for the PR!

It seems that it is impossible to have a misuse resistant implementation of the digest traits as they currently are. I'm wondering whether it might be better to first change the digest traits to something we're more happy with and then implementing these improved traits for sha3, rather than having this implementation for the sub-optimal traits. If we can't get this in before the next release, we could remove the current `Hasher` type-alias as it can't be instantiated.

crates/algorithms/sha3/src/impl_digest_trait.rs
jschneider-bensch: Yes, my main reason was keeping as much code as possible outside the macro. But I can understand your points as well, and have also thought about getting rid of the `*Hasher` alias in favor of just implementing the traits on `$type` instead. My guess on the intended use of the `libcrux-traits` type `Hasher` is that this should implement the `DigestIncremental*` traits somehow generically, but it doesn't and you can't implement them here because of the orphan rule.

crates/algorithms/sha3/src/impl_digest_trait.rs
jschneider-bensch: :100: agree on an overhaul of the traits.

crates/algorithms/sha3/src/impl_digest_trait.rs
jschneider-bensch: Thanks for thinking through this scenario. In this case I think indeed a panic might be preferable to silently subverting expectations, because the panic would at least give an indication that the API is misused (twice!). It does highlight that the API is still flawed to allow the misuse in the first place when a different API could rule it out.

crates/algorithms/sha3/src/impl_digest_trait.rs
robinhundt: I agree, implementing the traits directly on `$type` and removing the `Hasher` alias sounds good.

I've had a quick look at the `Hasher` struct and I also struggle to understand what it provides. I think we could get rid of it when reworking the digest traits.

crates/algorithms/sha3/src/impl_digest_trait.rs
robinhundt: I've created #1496

robinhundt:
Result from call: Change the libcrux-traits digest traits to consume self on finish before this change.

crates/algorithms/sha3/src/impl_digest_trait.rs
jschneider-bensch: I've implemented the traits on `$type` now, but kept the `Hasher` alias. What it provides is a somewhat more convenient receiver API for invoking the digest traits. See the usage in `tests/incremental_traits.rs`.

traits/src/digest/slice.rs
robinhundt, line 110: We could add `Self::IncrementalState` to the `FinishError::InvalidDigestLength` variant. This way, in case of an insufficient long digest buffer, you would not loose the hasher state, and could e.g. realloc the buffer.

This pattern is similar to the `Mutex::lock` API, which contains self inside the `PoisonError` https://doc.rust-lang.org/std/sync/type.LockResult.html.

robinhundt: I'm totally fine with not doing this change now. I think it is clear that the digest traits in their current form are not what we want and we'll likely rework them anyway.

E.g., I have my doubts whether such a fallible method that writes the digest into a slice is even needed. In my opinion, the trait should be minimal and only operate on arrays or references to arrays. Then no errors are needed and everything should be infallible. If a user only has a slice, it is their responsibility to convert this to a reference to an array and handle the potential error.

crates/algorithms/blake2/tests/blake2.rs
```diff
@@ -208,24 +208,31 @@ fn test_blake2s() {
 
 #[test]
 fn test_digest_traits_2s() {
+    fn updated_hasher() -> Blake2sHasher<32> {
```
robinhundt, line 211: If Blake2sHasher was clone, this would be a bit easier.

robinhundt: I created #1500 to track this.

CHANGELOG.md
```diff
@@ -9,11 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- (libcrux-sha3) [#1493](https://github.com/cryspen/libcrux/pull/1493): Provide instantiable `libcrux-traits` incremental hashing API
 - (libcrux-secrets) [#1460](https://github.com/cryspen/libcrux/issues/1460): Fix incorrect cmp in aarch64 select/swap implementation
 - (libcrux-sha3) [#1456](https://github.com/cryspen/libcrux/pull/1456): Fix out of bounds indexing in avx2 SHAKE-256 implementation
 
 ### Changed
 
+- (libcrux-traits, libcrux-sha3, libcrux-sha2, libcrux-blake2, libcrux-hmac) [#1493](https://github.com/cryspen/libcrux/pull/1493): `finish` methods on digest traits consume the hasher
```
robinhundt: `libcrux-hmac` is not needed here. Only the `mut self` was changed to `self` in the `finalize` implementation, which doesn't impact a user of an API.

crates/algorithms/hmac/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#1382](https://github.com/cryspen/libcrux/pull/1382): Add an incremental API
 
+## Changed
+
+- [#1493](https://github.com/cryspen/libcrux/pull/1493): `finish` methods on digest traits consume the hasher
+
```
robinhundt: The change to digests only lead to internal changes in hmac and doesn't require a changelog entry.

robinhundt: And it seems that the digest traits are not re-exported from hmac.

robinhundt (approved):
Thanks! This is much nicer now. I think the more simple implementation of the traits for Sha-3 shows that this is the right direction.

crates/algorithms/blake2/tests/blake2.rs
jschneider-bensch: It's an easy change, but I would postpone it to a follow-up, if that's okay for you.

crates/algorithms/hmac/CHANGELOG.md
jschneider-bensch: True, the HMAC incremental API is exposed via its own `HmacState<OUTLEN>` trait.

traits/src/digest/slice.rs
jschneider-bensch: It's a good idead, but I agree that we should rather have another look at what these traits should be altogether for the next release. You already pointed out more issues in #1503.

PR #1499, opened 2026-07-06
Replace IntOps and EncodeOps traits with inherent methods (#1491)

Arul-Sujith:
This PR resolves issue #1491 by replacing the `IntOps` and `EncodeOps` traits in the `libcrux-secrets` crate with inherent methods directly on the `Secret<T>` type. 

## Rationale
Previously, downstream crates utilizing `check-secret-independence` had to conditionally import `IntOps` (and `EncodeOps`) to perform wrapping arithmetic and byte encodings on `Secret` types. Because these operations are already inherent methods on the standard public integer types (like `u8`, `u32`, `i64`), wrapping them behind a custom trait for the `Secret` type introduced unnecessary friction.

By migrating these methods to be inherent on `Secret<T>`, we achieve better API parity with public integers and eliminate the need for `cfg`-gated trait imports in dependent crates.

## Changes
- Removed the `IntOps` and `EncodeOps` trait definitions from `crates/utils/secrets/src/traits.rs`.
- Updated `crates/utils/secrets/src/int/secret_integers.rs` to implement `Secret<$t>` inherently instead of implementing the traits.
- Modified visibility of the migrated methods (e.g., `wrapping_add`, `to_le_bytes`) to `pub fn`.
- Cleaned up the unused `IntOps` import in `crates/utils/secrets/src/int.rs`.

## Verification
- `cargo test -p libcrux-secrets` passes.
- `cargo test -p libcrux-secrets --features check-secret-independence` passes.
- The full `libcrux` workspace (`cargo test --workspace`) compiles and passes without any downstream trait resolution issues.

### DIFF
--- a/crates/utils/secrets/CHANGELOG.md
+++ b/crates/utils/secrets/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- [#1491](https://github.com/cryspen/libcrux/issues/1491): Replace `IntOps` and `EncodeOps` traits with inherent methods on `Secret`
 - [#1446](https://github.com/cryspen/libcrux/pull/1446): Remove const qualifier of secret types constructors
 - [#1462](https://github.com/cryspen/libcrux/pull/1462): More robust casts instead of transmutes when checking secret independence
 - [#1484](https://github.com/cryspen/libcrux/pull/1484): seal scalar trait and synchronize De-/Classify trait impls for public/secret types
--- a/crates/utils/secrets/src/int.rs
+++ b/crates/utils/secrets/src/int.rs
@@ -157,8 +157,6 @@ pub trait Select {
 mod portable {
     use super::{Select, Swap};
     use crate::Declassify;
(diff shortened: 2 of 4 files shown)


robinhundt (approved):
Great, thanks for the PR! I've tested these changes [in libcrux-iot](https://github.com/celabshq/libcrux-iot/pull/192) where we use libcrux-secrets more than in this repo, an they provide a nice cleanup there.

crates/utils/secrets/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- [#1491](https://github.com/cryspen/libcrux/issues/1491): Replace `IntOps` and `EncodeOps` traits with inherent methods on `Secret`
```
robinhundt, line 16: @jschneider-bensch This change should also be in the top-level changelog. Should we just do it as part of the release Changelog changes?

jschneider-bensch: Yep, we can do that!

PR #1505, opened 2026-07-08
Integrate `hpke-rs` in workspace

jschneider-bensch:
This PR pulls `hpke-rs`  into the `libcrux` workspace under `crates/protocols/hpke`. It should preserve the git history of `hpke-rs`, while the version tags have been renamed to include an `hpke-rs-` prefix (e.g. `v0.6.0` in the original repo is renamed to `hpke-rs-v0.0.6` here).

I also moved CI workflows and updated their default working directories. I'm not sure I fixed up all things in the coverage and GH pages setups already though, so I've disabled those workflows for now.

[skip changelog]


Cargo.toml
```diff
@@ -114,14 +118,20 @@ libcrux-digest = { version = "=0.0.8", path = "crates/primitives/digest" }
...
 
+# HPKE
+hpke-rs = { version = "=0.6.1", path = "crates/protocols/hpke" }
+hpke-rs-crypto = { version = "=0.6.1", path = "crates/protocols/hpke/traits", default-features = false }
+hpke-rs-rust-crypto = { version = "=0.6.1", path = "crates/protocols/hpke/rust_crypto_provider", default-features = false }
+hpke-rs-libcrux = { version = "=0.6.1", path = "crates/protocols/hpke/libcrux_provider", default-features = false }
+
 # For the crates below it is necessary to set `default-features = true`
 # in workspace crates that depend on the default feature set.
+libcrux-kem = { version = "=0.0.8", path = "libcrux-kem", default-features = false }
```
jschneider-bensch, line 133: `libcrux-kem` and `libcrux-traits` are used with `default-features = false` in `hpke-rs-libcrux`. Unfortunately, you can't set that at the crate level for a workspace dependency, so I've set it here and flipped all other workspace dependencies on these to crates to use `default-features = true` which you *can* set at the crate level.

franziskuskiefer:
Thanks! I'll do a full pass on this later. I opened https://github.com/celabshq/hpke-rs/issues/165 to track the move and make sure everything goes smoothly.

crates/protocols/hpke/libcrux_provider/Readme.md
```diff
@@ -0,0 +1,18 @@
...
+Please see [hpke-rs] for more details.
+
+[libcrux]: https://crates.io/crates/libcrux
+[hpkecrypto]: https://github.com/celabshq/hpke-rs/tree/main/traits
+[rustc-image]: https://img.shields.io/badge/rustc-1.56+-blue.svg?style=for-the-badge
+[docs-badge]: https://img.shields.io/badge/docs-rs-blue.svg?style=for-the-badge
+[docs-link]: https://docs.rs/hpke-rs-libcrux
+[crate-badge]: https://img.shields.io/crates/v/hpke-rs-libcrux.svg?style=for-the-badge
+[crate-link]: https://crates.io/crates/hpke-rs-libcrux
+[hpke-rs]: https://github.com/celabshq/hpke-rs
```
franziskuskiefer: We should update all the URLs. Or shall we do that in a follow up to keep this one clean?

Cargo.toml
franziskuskiefer: I think that's fine. I just wonder if there's a reason for that or if we could just have the default features on?

crates/protocols/hpke/rust_crypto_provider/Readme.md
```diff
@@ -0,0 +1,26 @@
+# HPKE Crypto provider using native Rust
+
+[![crates.io][crate-badge]][crate-link]
+[![Docs][docs-badge]][docs-link]
+![Rust Version][rustc-image]
+
```
franziskuskiefer, line 6: Given that this is in the libcrux repo when this gets merged. Let's add something for this provider already that we don't commit to maintaining it right now.


```suggestion

The crate is a community-supported project and is not officially maintained by the core team right now.
However, we are happy to review and merge pull requests. Contributions, bug fixes,
and feature additions from the community are welcome and encouraged!

```

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- [#1505](https://github.com/cryspen/libcrux/pull/1505): Update dependency: `hpke-rs`
```
franziskuskiefer:
```suggestion
- [#1505](https://github.com/celabshq/libcrux/pull/1505): Update dependency: `hpke-rs`
```

.github/workflows/hpke-rust.yml
```diff
@@ -0,0 +1,136 @@
...
+        os:
+          - macos-15-intel
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v6
```
franziskuskiefer: Let's use the most recent version (7) here already.

.github/workflows/hpke-rust.yml
```diff
@@ -0,0 +1,136 @@
+name: HPKE - Build & Test
+
+on:
+  push:
+    branches:
+      - '*'
```
franziskuskiefer: Let's not do on every branch. But on every PR.

.github/workflows/hpke-rust.yml
```diff
@@ -0,0 +1,136 @@
...
+  pull_request:
+    branches: [ main ]
```
franziskuskiefer: Let's also add paths to all of the hpke workflows to run only when hpke changes. Not granular enough, I know. But better than always running everything.

franziskuskiefer (dismissed):
Thanks, I think this generlaly looks good. Just a few small things.

I didn't look at details of the code and expect it to just being moved.

Let's track the gh pages and coverage (and other follow ups) on an issue to see what to do about them.

CHANGELOG.md
jschneider-bensch: Oops, thanks! I have to update my completion snippet for changelog entries.

.github/workflows/hpke-rust.yml
jschneider-bensch: Turns out, this commit already pins [v7](https://github.com/actions/checkout/releases/tag/v7.0.0), the comment was just wrong. I've fixed the comment and pinned all other uses of the action to v7 as well.

.github/workflows/hpke-rust.yml
jschneider-bensch: I've changed it so the workflows are run when there are changes in `crates/protocols/hpke` or in one of the the Rust files in the direct workspace dependencies.

Cargo.toml
jschneider-bensch: `libcrux-traits` doesn't define default features, so indeed it doesnt' make a difference, and I'll change it back to keep these configs simpler.

`libcrux-kem` has `std` as a default feature, so we can't have those set if we want to be `no_std` by default.

franziskuskiefer (approved):
Thanks, let's get this in and update the hpke-rs issue to redirect everything here.

PR #1506, opened 2026-07-08
Upgrade hax-lib to v0.3.7

jschneider-bensch:
This PR rebases some of the commits from #1478 onto #1505. Because `hpke-rs` is now part of the workspace, there is only hax-lib v0.3.7 in the dependency tree, since it's the workspace dependency. This means I've dropped the changes to the F* Makefiles that look for a git dependency of hax-lib.

[skip changelog]

---
# Adapted description from #1478 
## What this does
Upgrades the workspace from hax-lib 0.3.6 to 0.3.7 (`cargo-hax-v0.3.7`, rev `d8b5b3d`) and bumps the CI F\* version to v2026.03.24 (required by the 0.3.7 hax proof-libs), with the minimal F\* changes needed to keep the hax proof jobs green under 0.3.7. 

## Changes
- **Toolchain pin** — ~`hax-lib` → git tag `cargo-hax-v0.3.7` (`[workspace.dependencies]` + `[patch.crates-io]`), specs workspace pin; `hpke` disabled (hpke-rs pulls a crates.io libcrux pinned to 0.3.6).~ CI F\* set to v2026.03.24 in the mlkem/mldsa/sha3 hax workflows.
- **F\* changes(no impl-logic changes)** —
  - ml-kem `Spec.Utils`: 0.3.7 dropped `Core_models.Ops.Range.impl_index_range_slice` (→ explicit bounds) and turned `f_neg` into a `t_Neg` typeclass with no machine-int instance (→ re-provide `t_Neg i16`).
  - core-models `from_fn` / `BitVec::from_fn`: 0.3.7 emits two type implicits at call sites; add the absorbing slot to the extraction overrides.
  - ml-dsa: `Alloc.Vec.t_Vec` is now a distinct type (was `unfold = t_Slice`); index `to_vec()` ghost snapshots via `.as_slice()`, and rewrite one `cloop!` iterator (`inside_out_shuffle`) as an indexed loop.
- **CI tooling** — the proof scripts now run `make -k` and print an F\* error summary, so a single nightly run reports every failing module rather than stopping at the first.

**REGRESSION:** 22 functions (12 in ml-dsa, 10 in ml-kem) no longer verify under hax-lib 0.3.7 / F\* v2026.03.24 and are admitted here (`panic_free` where panic-freedom still holds, `lax` otherwise); **a subsequent PR re-proves them all with more comprehensive and stable proofs.** (The 0.3.7 hax proof-libs themselves require F\* v2026.03.24 — they don't elaborate under older F\* — so an older toolchain can't be used to sidestep these.)

<details>
<summary>The 22 admitted functions</summary>

**ml-dsa (12)** — `Polynomial.barrett_reduce`; `Simd.{Avx2,Portable}.Arithmetic.barrett_reduce_simd_unit`; `Simd.Avx2.Arithmetic.montgomery_multiply_aux`; `Simd.Avx2.Encoding.{Commitment.serialize, T0.serialize, T1.serialize, Error.serialize_when_eta_is_2, Error.serialize_when_eta_is_4, Gamma1.serialize_when_gamma1_is_2_pow_17, Gamma1.serialize_when_gamma1_is_2_pow_19}`; `Simd.Avx2.Invntt.invert_ntt_at_layer_3`.

**ml-kem (10)** — `Mlkem{512,768,1024}.Rand.{generate_key_pair, encapsulate}`; `Vector.Portable.Vector_type.from_i16_array`; `Vector.Avx2.Arithmetic.to_unsigned_representative`; `Polynomial.subtract_reduce`; `Ind_cca.Unpacked.generate_keypair`.

Causes are a mix of 0.3.7 SMT "incomplete quantifiers" regressions, z3 timeouts (up to ~5 min/query), and the changed `CryptoRng`/`Rand_core` trait model.
</details>


.github/workflows/kmac-hax.yml
```diff
@@ -46,8 +36,11 @@ jobs:
 
       - uses: hacspec/hax-actions@main
         with:
-          hax_reference: ${{ github.event.inputs.hax_ref || needs.get-hax-ref.outputs.hax_ref }}
-          fstar: v2025.12.15
+          hax_reference: v${{ env.HAX_LIB_VERSION }}
```
robinhundt: The output of the get-hax-ref action was set as
```
echo "hax_ref=hax-lib-v$HAX_VERSION" >> $GITHUB_OUTPUT
```
So I think you're missing the `hax-lib` prefix here.
```suggestion
          hax_reference: hax-lib-v${{ env.HAX_LIB_VERSION }}
```

robinhundt (commented):
We should add a note to the CONTRIBUTING.md that on a fresh checkout, you might need to do `cargo update hax-lib --precise $HAX_LIB_VERSION, as it could resolve to a newer one than is needed for the proofs.

.github/workflows/secrets-build-test.yml
```diff
@@ -175,7 +175,7 @@ jobs:
 
   secrets-build-test-status:
     if: ${{ always() }}
-    needs: [build, build-no_std, valgrind-ct-test]
```
robinhundt, line 178: Why is the `build-no_std` job removed from the status gate?

.github/workflows/mldsa-hax.yml
```diff
@@ -76,37 +72,35 @@ jobs:
...
-      - uses: actions/download-artifact@v8
-        with:
-          name: fstar-extractions
-          path: .
-
-      - name: 🏃 Prove ML-DSA crate
-        working-directory: libcrux-ml-dsa
-        run: ./hax.sh prove
+  # XXX: Is the caching performed by hacspec/hax-actions enough to get CI minutes down for paid runners?
+  # prove:
```
robinhundt, line 77: Do we already have issues for tracking re-enabling these proofs?

robinhundt (approved):
Thanks!
I think we can also remove the `get-hax-ref` workflow, right? It doesn't seem to be used anymore.

.github/workflows/secrets-build-test.yml
jschneider-bensch: Ah, a removal in error, indeed! I think the job had been removed from the workflow when Karthik made his changes, so this was a fix for a missing job, but subsequently the job was re-instated in the workflow, but I missed updating the no-longer-fix.

.github/workflows/mldsa-hax.yml
jschneider-bensch: We have the documents and tasks in notion, but you're right, it doesn't hurt to have issues here as well.

jschneider-bensch: I've created #1529 for it.

jschneider-bensch:
I've removed the `get-hax-ref.yml` workflow. I thought it was convenient that you could trigger the workflows with a specific hax version if you wanted, but probably we don't actually want to use CI minutes to test a non-standard hax/hax-lib version but such things should be tested locally, if possible.

PR #1507, opened 2026-07-08
[secrets] Fix changelog entry

jschneider-bensch:


robinhundt (approved):
Oops, that was my typo. Thanks for fixing it!

PR #1515, opened 2026-07-09
[CI] Group dependabot updates

jschneider-bensch:
[skip changelog]

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,12 +3,8 @@ updates:
 - package-ecosystem: cargo
   directory: "/"
   schedule:
-    interval: weekly
+    interval: "weekly"
   open-pull-requests-limit: 10
-  groups:
-    all-cargo:
-      patterns:
-        - "*"
 - package-ecosystem: github-actions
   directory: "/"
   schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,16 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+  groups:
+    all-cargo:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,8 +5,16 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+  groups:
+    all-cargo:
+      patterns:
+        - "*"
```
robinhundt: I think it would make sense to distinguish between dev-dependencies and normal/build-dependencies. Grouping the dev-dependencies should be fine (although it might be annoying if one update in a group is breaking and the others are not, then we need to fix that one up before merging, or split manually).

For normal dependencies, I think it would make sense to keep a single PR per update. We don't have a lot of these anyway, and they demand a closer look when they occur.

robinhundt: Mh, according to https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#dependency-type-groups the dependency-type filter for groups is sadly not available with cargo.

jschneider-bensch: Yeah, I also just saw that. But it would be nice to have some time

robinhundt: Given that the dependency-type filter is not available for cargo, I'd lean towards not grouping rust dependency updates. We don't have a lot of external dependencies and I'd rather not accidentally merge an actual dependency if it is grouped with a bunch of dev-dependencies.

I'm fine with grouping action updates.

jschneider-bensch: That's fair.

robinhundt (approved):
Thanks!

Issue #1504, opened 2026-07-07
[c-extraction] cloop! macro step_by variant can result in wrong iteration


robinhundt:
The ml-kem and ml-dsa crates contain a `cloop!` macro that converts iterator adapters into imperative rust code for c extraction. There are bugs in the `step_by` variant of the macro (which are currently not hit by the existing uses).

The ML-KEM variant looks like this:
https://github.com/celabshq/libcrux/blob/e83007e8dc8d8b2189431db69e029ec44ef53b12/libcrux-ml-kem/src/helper.rs#L30-L35

Here, there are two issues:
- the loop is only correct if `$start == 0`
- the loop is only correct if `$end % $step == 0`

If one of them is false, it would result in a wrong loop.

In ml-dsa the step_by variant is slightly different and the second issue was fixed. The first issue is still present. In ml-dsa, there is actually a use of the `cloop!` macro with a step that doesn't divide the end, which is why it was probably fixed already. In ml-kem, there is currently no such use.

https://github.com/celabshq/libcrux/blob/e83007e8dc8d8b2189431db69e029ec44ef53b12/libcrux-ml-dsa/src/helper.rs#L60-L66

I first noticed this issue in libcrux-iot and submitted a PR there to fix it, for which the C extraction currently doesn't compile https://github.com/celabshq/libcrux-iot/pull/191.

PR #1516, opened 2026-07-09
Fix cloop and warnings

Arul-Sujith:
This PR implements a fix for the `cloop!` macro's `step_by` variant as discussed in issue #1504.

## Changes
- The `step_by` variant of the `cloop!` macro in both `ml-kem` and `ml-dsa` was mathematically incorrect for cases where `$start != 0` or `$end % $step != 0`. This replaces the flawed implementation with a rigorously bounded `for` loop `0..(($end) / $step + 1)` that computes the offset internally. This maintains Eurydice's ability to extract cleanly to C (unlike `loop { ... }` approaches) while preventing loop bound overshoots and `$start` scaling bugs.

### DIFF
--- a/libcrux-ml-dsa/CHANGELOG.md
+++ b/libcrux-ml-dsa/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


jschneider-bensch:
This is a welcome change in principle, but will require an update of the C extractions.

PR #1520, opened 2026-07-10
hpke: Add P384 test vectors

### LINKED ISSUE #151: Generate and check KATs for P384
The HPKE RFC doesn't specify any test vectors for P384. It should be easy enough to generate fresh ones using another implementation and verify them against this implementation.

franziskuskiefer:
Test vectors are generated with
https://github.com/hpkewg/hpke 4abc37efc36a4295519964e366f650814b0f3cff.

Update KAT tests to ensure all tests are run and single shot APIs aren't run every time (significantly reduce kat test time).

Fixes https://github.com/celabshq/hpke-rs/issues/151

[skip changelog]


crates/protocols/hpke/tests/test_hpke_kat.rs
```diff
@@ -60,7 +68,11 @@ struct ExportsKAT {
     exported_value: String,
 }
 
-fn kat<Crypto: HpkeCrypto + 'static>(tests: Vec<HpkeTestVector>) {
+/// Run the known-answer tests for all `tests` supported by the `Crypto` backedn,
```
jschneider-bensch:
```suggestion
/// Run the known-answer tests for all `tests` supported by the `Crypto` backend,
```

crates/protocols/hpke/tests/test_hpke_kat.rs
```diff
@@ -26,16 +33,17 @@ struct HpkeTestVector {
...
     psk_id: Option<String>,
     pkRm: String,
     pkSm: Option<String>,
-    pkEm: String,
+    pkEm: Option<String>,
     enc: String,
     shared_secret: String,
-    key_schedule_context: String,
-    secret: String,
+    suite_id: Option<String>,
```
jschneider-bensch, line 44: What's the purpose of the `suite_id`? We don't test anything about it.

jschneider-bensch (approved):
Thanks!
I've filed #1527 as a follow-up.

crates/protocols/hpke/tests/test_hpke_kat.rs
```diff
@@ -26,16 +33,17 @@ struct HpkeTestVector {
     ikmE: String,
     skRm: String,
     skSm: Option<String>,
-    skEm: String,
+    skEm: Option<String>,
```
robinhundt, line 36: Instead of using `hex_to_bytes` wherever these values are used, we could also use [`hex::serde`](https://docs.rs/hex/latest/hex/serde/index.html) in the struct itself, like we do in libcrux-kats.

jschneider-bensch: As far as I can see, this function is only in hpke for testing purposes, so removing it with #1527 would make sense.

robinhundt: Yep, that can be part of that issue. I've added a comment there.

crates/protocols/hpke/tests/test_hpke_kat.rs
franziskuskiefer: That's a new field in the new test vectors, defining the "ciphersuite". We can start checking that as well. https://github.com/hpkewg/hpke/blob/4abc37efc36a4295519964e366f650814b0f3cff/draft-ietf-hpke-hpke.md?plain=1#L723. But it's implicitly tested anyway.

jschneider-bensch: I figured that it would be redundant with the KEM, KDF etc IDs.

PR #1524, opened 2026-07-12
build(deps): only use rand@0.10 in hpke

boxdot:
Port of https://github.com/celabshq/hpke-rs/pull/159 after hpke-rs got integrated into the libcrux repository.


franziskuskiefer (approved):
Thanks! lgtm

PR #1525, opened 2026-07-13
 bump tls_codec to 0.5.0

franziskuskiefer:
There are some changes I want to do on the tls codec release before getting this out. But let's get this first one in already.
[skip changelog]


franziskuskiefer:
ah this fails on the vlbytes serialize as well. Then let's wait until I got that fixed.

jschneider-bensch (approved):

PR #1530, opened 2026-07-14
Run AES RISC-V jobs on GH runners

jschneider-bensch:
This installs nix and supposedly caches the store for subsequent runs as well.

### DIFF
--- a/.github/workflows/aes-riscv-arm.yml
+++ b/.github/workflows/aes-riscv-arm.yml
@@ -48,7 +48,7 @@ jobs:
           cargo build --release --target ${{ matrix.target }}
 
   test:
-    runs-on: self-hosted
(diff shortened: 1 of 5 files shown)


robinhundt:
I also just saw the failure and started on a fix. I'd just use cargo cross personally, which is much simpler and does the setup for us.

robinhundt:
https://github.com/celabshq/libcrux/pull/1528/changes/98ee9e86feb2cba4fdf8449582dcedff0d4f2af4 is what I'd go with. cross also runs the tests under qemu and setting it up seems to be sufficiently fast. I like nix, but here I think cross might be the better tool.

jschneider-bensch:
Simpler sounds great, thanks! :+1:

PR #1539, opened 2026-07-16
hpke: draft-ietf-hpke-pq support

franziskuskiefer:
Adds the post-quantum and PQ/T-hybrid HPKE algorithms from [draft-ietf-hpke-pq-05](https://datatracker.ietf.org/doc/html/draft-ietf-hpke-pq-05), behind a new `draft-ietf-hpke-pq` feature (libcrux provider only).

Added
- Single-stage KDFs `Shake128 = 0x0010` and `Shake256 = 0x0011`, with the one-stage HPKE key schedule (KeySchedule, Export, DeriveKeyPair).
- Pure ML-KEM: `MlKem512 = 0x0040` (plus existing MlKem768/MlKem1024 repointed to the IETF draft).
- PQ/T-hybrid KEMs `MlKem768P256 = 0x0050`, `MlKem1024P384 = 0x0051`, and `MLKEM768-X25519 = 0x647a`, per [draft-irtf-cfrg-concrete-hybrid-kems-04](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-concrete-hybrid-kems-04).

Deprecated
- draft-connolly-cfrg-hpke-mlkem, superseded by draft-ietf-hpke-pq. The two are mutually exclusive (build warning + docs.rs pins a working feature set).

Notes
- MLKEM1024-P384 also needs `rustcrypto-p-curves` for now.
- Adds a `bench_pq` benchmark and `test_hpke_pq` suite with the draft's vectors.
- Updates kats to check for explicit ciphersuite (kem) support.

To test the new code run `cargo test -p hpke-rs -F draft-ietf-hpke-pq,rustcrypto-p-curves`.

Claude helped push some of the code around.


crates/protocols/hpke/CHANGELOG.md
```diff
@@ -14,6 +14,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#146](https://github.com/celabshq/hpke-rs/pull/146): Add support for ML-KEM768 and ML-KEM1024 gated behind the `draft-connolly-cfrg-hpke-mlkem` feature flag.
 
 
+### Added
+
+- Support for the post-quantum and PQ/T-hybrid algorithms of
+  [draft-ietf-hpke-pq](https://datatracker.ietf.org/doc/html/draft-ietf-hpke-pq-04),
+  in the **libcrux provider only**, behind the new `draft-ietf-hpke-pq` feature.
+  Validated byte-for-byte against the draft's published test vectors.
```
keks:
```suggestion
  Validated against the draft's published test vectors.
```

crates/protocols/hpke/benches/bench_pq.rs
```diff
@@ -0,0 +1,328 @@
...
+            }
+            for &kem_mode in PQ_KEM_IDS {
+                if Crypto::supports_kem(kem_mode).is_err() {
+                    continue;
+                }
+                bench_suite::<Crypto>(c, HpkeMode::Base, kem_mode, kdf_mode, aead_mode, true);
+            }
+        }
+    }
+}
```
keks, line 53: I am not 100% sure, but it seems bench.rs and bench-pq.rs are the same except that the IDs that are being iterated over are different and the non-pq version does not have an option to benchmark key generation.

Are there deeper differences? Otherwise I would be in favour of merging the files and just have different top-level functions for pq and non-pq (where non-pq can use the flag to disable benchmarking key generation).

crates/protocols/hpke/src/kdf.rs
```diff
@@ -6,6 +6,57 @@ use crate::util::concat;
...
+) -> Vec<u8> {
+    concat(&[
+        ikm,
+        HPKE_VERSION,
+        suite_id,
+        &length_prefixed(label.as_bytes()),
+        &(len as u16).to_be_bytes(),
+        context,
+    ])
+}
```
keks: I am not convinced that this function is needed. In the draft they also just call concat from labeled_derive. I would just inline it.

https://datatracker.ietf.org/doc/html/draft-ietf-hpke-hpke-03#name-labeled-derivation-function

crates/protocols/hpke/libcrux_provider/src/lib.rs
```diff
@@ -49,7 +60,15 @@ impl HpkeCrypto for HpkeLibcrux {
...
-        let alg = kdf_algorithm_to_libcrux_hkdf_algorithm(alg);
+        // SHAKE256 is a single-stage (XOF) KDF: there is no separate extraction
+        // step, so `Extract(salt, ikm)` derives `Nh` bytes from `salt || ikm`.
+        // See draft-ietf-hpke-pq Section 5.
+        #[cfg(feature = "draft-ietf-hpke-pq")]
+        if let Some(nh) = shake_nh(alg) {
+            return Ok(shake_derive(alg, &[salt, ikm], nh));
+        }
+
+        let alg = kdf_algorithm_to_libcrux_hkdf_algorithm(alg)?;
```
keks: The draft defines separate interfaces for these kinds of algorithms, and I don't think we should force them all into a single API. So maybe we should instead

- define a new enum SingleStageKdfAlgorithm for the Shake algorithms (not a nice name, maybe there is something nicer)
- define a new enum TwoStageKdfAlgorithm for the HKDF-based et al.
- change the trait methods `kdf_extract` and `kdf_expand` to take the TwoStageKdfAlgorithm 
- define a new trait method `kdf_derive(alg:SingleStageKdfAlgorithm, ikm: &[u8], l: usize) -> Result<Vec<u8>, Error>;` and implement that here, and use that in `labeled_derive`
- Give the caller a way to convert the KdfAlgorithm into either SingleStageKdfAlgorithm or a TwoStageKdfAlgorithm -- or maybe just define
  ```rust
  enum KdfAlgorithm {
    SingleStage(SingleStageKdfAlgorithm),
    TwoStage(TwoStageKdfAlgorithm),
  }
  ```
- otherwise leave these functions as-is
- make the callers pick the function to call based on whether it's single-stage or two-stage

Of course changing the trait breaks compatibility, but so does changing the enum, so I think it's fine.

crates/protocols/hpke/libcrux_provider/src/lib.rs
```diff
@@ -540,12 +729,291 @@ fn aead_alg(alg_type: AeadAlgorithm) -> Result<libcrux_aead::Aead, Error> {
...
+        libcrux_sha3::sha256(&input).to_vec()
+    }
+
+    #[inline]
+    fn split(data: &[u8], n: usize, err: Error) -> Result<(&[u8], &[u8]), Error> {
+        if data.len() < n {
+            return Err(err);
+        }
+        Ok(data.split_at(n))
+    }
```
keks, line 699: I am not sure we need this helper instead of `data.split_at_checked(n).ok_or(err)`, but I agree it makes the calling code a little shorter, which is nice. How do you feel about the name `split_at_or`?

crates/protocols/hpke/libcrux_provider/src/lib.rs
```diff
@@ -540,12 +729,291 @@ fn aead_alg(alg_type: AeadAlgorithm) -> Result<libcrux_aead::Aead, Error> {
...
+/// * Rejection-sampled scalars and uncompressed SEC1 point encodings for NIST;
+///   `ss_T` is the x-coordinate of the DH result.
+/// * Wire formats: `ek = ek_PQ || ek_T`, `ct = ct_PQ || ct_T`.
+#[cfg(feature = "draft-ietf-hpke-pq")]
+mod hybrid {
+    use super::*;
+
+    /// The nominal group (traditional component) of a hybrid KEM.
+    #[derive(Clone, Copy)]
+    enum NistCurve {
```
keks: Should this be called NominalGroup? That would leave the door for X25519 open. Actually, why is X25519 not in here? Because it is in a separate X-Wing impl? Maybe we could merge these? :thinking:

crates/protocols/hpke/traits/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - [#154](https://github.com/celabshq/hpke-rs/pull/154): `HpkeCrypto::HpkePrng` now has `rand` v0.10 bound `CryptoRng`, instead of `rand` v0.9 bound `RngCore + CryptoRng`
+- Add `KemAlgorithm::MlKem512 = 0x0040` (per `draft-ietf-hpke-pq`). Repoint the
```
keks: I think this needs a PR number? Also, aren't we adding more algorithms?

crates/protocols/hpke/traits/src/types.rs
```diff
@@ -223,6 +261,42 @@ pub enum KdfAlgorithm {
...
+    /// <https://datatracker.ietf.org/doc/html/draft-ietf-hpke-pq-04>
+    TurboShake256 = 0x0013,
+}
+
+impl KdfAlgorithm {
+    /// Whether `kdf` is a single-stage (XOF) KDF, per draft-ietf-hpke-pq.
+    ///
+    /// Single-stage KDFs use [`labeled_derive`] and a different key-schedule shape
+    /// than the two-stage HKDF KDFs (which use [`labeled_extract`]/[`labeled_expand`]).
+    pub fn is_one_stage(&self) -> bool {
```
keks: This should be `is_single_stage` (or made obsolete by refactoring the whole single-stage vs two-stage mechanism)

keks (dismissed):
I think this is a pretty good start! I left some comments. Most things are pretty minor, but I think the handling of singe stage KDFs could be improved, and that might be a little bit more involved.

crates/protocols/hpke/benches/bench_pq.rs
franziskuskiefer: Yeah, I thought this could make sense but you're right that it ended up with too much duplicate code.

crates/protocols/hpke/libcrux_provider/src/lib.rs
franziskuskiefer: Yeah this could be a bit cleaner. I don't want to change wire types and make that too complicated. But pulling this apart a bit more makes sense.
This is much more churn though. So I'm not sure if it's actually better. I tried some options and aren't super happy with either.

crates/protocols/hpke/libcrux_provider/src/lib.rs
franziskuskiefer: Yes we can use NominalGroup. x25519 isn't used in any combiner like this and is handled pretty differently. So no need for it in here.

crates/protocols/hpke/src/kdf.rs
franziskuskiefer: yeah, it's a bit more verbose than needed.

crates/protocols/hpke/libcrux_provider/src/lib.rs
keks: We could also just add a kdf_derive and we return an error if the wrong kind of method is called. Then at least we are not squeezing it into the wrong interface.

crates/protocols/hpke/libcrux_provider/src/lib.rs
keks: Oh okay. In the RFC it looked like doing the normal combiner with X25519 would just be XWing, but I didn't look to closely.

crates/protocols/hpke/CHANGELOG.md
```diff
@@ -5,6 +5,31 @@ All notable changes to this project will be documented in this file.
...
+  - ML-KEM-512 (`0x0040`), ML-KEM-768 (`0x0041`), and ML-KEM-1024 (`0x0042`) as
+    HPKE KEMs.
+  - The PQ/T-hybrid KEMs `KemAlgorithm::MlKem768P256 = 0x0050`,
+    `KemAlgorithm::MlKem1024P384 = 0x0051`, and `MLKEM768-X25519` (`0x647a`,
+    via X-Wing draft-06), per
+    [draft-irtf-cfrg-concrete-hybrid-kems-03](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-concrete-hybrid-kems-03).
+
+### Deprecated
+
+- The `draft-connolly-cfrg-hpke-mlkem` feature is deprecated in favour of
```
keks: Should this also have a PR number?

crates/protocols/hpke/libcrux_provider/src/lib.rs
```diff
@@ -582,7 +583,7 @@ fn p_curve_key_gen_seed(alg: KemAlgorithm, seed: &[u8]) -> Result<[u8; 32], Erro
         return Err(Error::InsufficientRandomness);
     }
 
-    let kdf_alg: KdfAlgorithm = alg.into();
+    let kdf_alg = two_stage_kdf(alg)?;
```
keks: A comment in the From impl says that the caller can override the KDF algorithm. Can/should we respect that override here?

https://github.com/celabshq/libcrux/blob/83670ff5672a591b98d2f1c6b57dfe645c9556f7/crates/protocols/hpke/traits/src/types.rs#L401-L404

keks (approved):
I think this generally looks good. There seems to be another missing PR number in the change logs, and I have one question about whether something should be change where I expect the answer is "no", so approving with a caveat :+1:

crates/protocols/hpke/libcrux_provider/src/lib.rs
franziskuskiefer: Good point. But this code is never actually used. The plumbing here is a bit weird. I removed the code. Only PQ ciphersuites use the KEM API right now. DH ciphersuites go through the DH API.

PR #1541, opened 2026-07-17
Add set_reseed_interval to support tighter reseed cadences (#1523)

### LINKED ISSUE #1523: hmac-drbg: allow configuring the reseed interval
## Summary

`libcrux-hmac-drbg` currently hardcodes the reseed interval as a crate-level `const RESEED_INTERVAL: u64 = 1 << 48` and keeps `reseed_counter` as a private field. There is no way for a consumer to instantiate a DRBG with a shorter reseed interval, nor to inspect/override the counter. Consumers that want a tighter, self-chosen reseed cadence are forced to track their own counter alongside the DRBG and call `reseed_from_rng` manually.

## Motivation

In OpenMLS's libcrux crypto provider we replaced a `rand` `ReseedingRng<ChaCha20Core, OsRng>` (which reseeded from the OS every 4 GiB of output) with `HmacDrbgSha256`. We'd like to preserve an application-defined reseed cadence (e.g. reseed from the OS every N bytes / N generate calls) as defense-in-depth, independent of the DRBG's very large default interval (`2^48`).

Today this is only achievable by wrapping the DRBG in our own struct that counts bytes and calls `reseed_from_rng` when a threshold is crossed. That works, but it duplicates state the DRBG already maintains and can't influence the DRBG's own `needs_reseed()` decision.

## Proposed API

Any of the following (in rough order of preference) would solve this:

1. **Per-instance configurable interval.** A constructor variant that takes the interval, e.g.
   ```rust
   pub fn new_with_reseed_interval(
       entropy_input: &[u8],
       nonce: &[u8],
       personalization_string: &[u8],
       reseed_interval: u64,
   ) -> Result<Self, InstantiateError>;
   ```
   and/or a setter `pub fn set_reseed_interval(&mut self, interval: u64)`. `needs_reseed()` / `generate()` would then compare against the per-instance value (defaulting to `RESEED_INTERVAL`).

   The same knob would ideally be plumbed through the `rand`-feature wrappers (`HmacDrbgRng` / `new_from_sys_rng` / `SeedableRng`).

2. **Expose the counter.** Keep the fixed interval but make the counter settable, e.g. `pub fn set_reseed_counter(&mut self, value: u64)`, so callers can force `needs_reseed()` to trip earlier.

## Notes / constraints

- The configured interval should still be clamped to `<= RESEED_INTERVAL` (the NIST SP 800-90A maximum) — the feature is about reseeding *more* often, never less.
- Please keep `1 << 48` as the default so existing behavior is unchanged.

Arul-Sujith:
This PR adds support for configuring a custom reseed interval in `libcrux-hmac-drbg`, directly addressing the feature request in issue #1523. It enables consumers to enforce a tighter, application-defined reseed cadence (e.g., reseeding from the OS every N bytes) as a defense-in-depth measure, independent of the DRBG's default global interval.

## Changes

- **Per-Instance Configuration**: Expanded `HmacDrbg` state to maintain a local `reseed_interval`, which defaults to the SP 800-90A maximum (`1 << 48`).
- **New Setters**: Added `pub fn set_reseed_interval(&mut self, interval: u64)` to `HmacDrbg` and the `rand`-featured `HmacDrbgRng` wrapper.
- **NIST Compliance (Clamping)**: The setter strictly bounds the input to the global `RESEED_INTERVAL`. A custom interval can force earlier reseeding, but it can *never* exceed the maximum interval allowed by the spec.
- **Dynamic Logic**: Updated the internal bounds checks in `generate()` and `needs_reseed()` to utilize `self.reseed_interval`.
- **Hax Proofs**: Updated the F* preconditions/postconditions to correctly target `self.reseed_interval` in place of the global constant, ensuring proofs remain sound.
- **Testing**: Added explicit test coverage (`custom_reseed_interval` and `custom_reseed_interval_clamping`) inside `src/tests.rs` to verify the early reseed triggering and the hard-cap logic.

Fixes #1523


robinhundt:
Thanks for the PR! We recently added a [policy on LLM-assisted submissions](https://github.com/celabshq/libcrux/blob/main/CONTRIBUTING.md#policy-on-llm-assisted-submissions).

Did you use any AI tooling? If yes, to what capacity?

Arul-Sujith:
Yes, I used an LLM to assist with this PR. Primarily to understand codebase, apply fixes, and generate PR descriptions. 

Then manually review the fixes and run tests. I will make sure to include this disclosure upfront in future PR descriptions.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- (libcrux-hmac-drbg) [#1523](https://github.com/celabshq/libcrux/issues/1523): Allow configuring a custom reseed interval
```
robinhundt, line 32: Could you please add an ## Unreleased section and add the entry there? 0.0.5 is already released.

robinhundt, line 32: Could you also add this entry to the libcrux-hmac-drbg crate's CHANGELOG under an ## Unreleased section?

CHANGELOG.md
```diff
@@ -37,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - (libcrux-sha3) [#1292](https://github.com/celabshq/libcrux/pull/1292): Add support for incremental CShake (https://github.com/kraemv)
 - (libcrux-hmac) [#1382](https://github.com/celabshq/libcrux/pull/1382): Add an incremental API
 
+
```
robinhundt, line 41: Spurious newline.

crates/algorithms/hmac-drbg/src/tests.rs
```diff
@@ -254,5 +254,26 @@ fn reseed_required_error() {
 fn needs_reseed_true_after_exhaustion() {
```
robinhundt, line 254: Because of the closing brace, this test function lost the assert that actually tested that the Drbg needs reseeding.

Please add the assert back to this function.

crates/algorithms/hmac-drbg/src/rand.rs
```diff
@@ -321,6 +322,13 @@ impl<const OUTLEN: usize, Hmac: HmacAlgorithm<OUTLEN>, ReseedRng: CryptoRng>
         Self { drbg, rng }
     }
 
+    /// Set the reseed interval for the underlying DRBG.
+    ///
+    /// The interval is clamped to a maximum of [`crate::RESEED_INTERVAL`].
+    pub fn set_reseed_interval(&mut self, interval: u64) {
+        self.drbg.set_reseed_interval(interval);
```
robinhundt, line 329: If the user supplied an interval of `0` this would lead to a panic in `HmacDrbgRng::safe_generate_small`. `needs_reseed` would be true, the drbg would be reseeded, but then the counter would still be greater than the interval (`1 > 0`) and generate would return `ReseedRequired`, hitting the unreachable in `safe_generate_small`.

Either we clamp the interval to the range `[1, RESEED_INTERVAL]` or we make this method fallible and return a 
```rust
ReseedIntervalError {
  TooSmall,
  TooLarge
}
```
(maybe there are better variant names).

@franziskuskiefer what would be more convenient for the uses in openmls?

robinhundt (changes requested):
Thanks for the PR! Can you please fix the changelogs and the tests?

For the change to the `set_reseed_interval` method I think @franziskuskiefer input would be good.

@franziskuskiefer I was also wondering whether configuring the `reseed_interval` enables you to more conveniently reseed after N number of bytes as described in #1523. If I understand the HMAC-DRBG correctly, the `reseed_interval` only specifies after how many generate calls, each of which can generate 2^19 bits, reseed must be called. If the output length is not equal per generate call, the `reseed_interval` can't be used to enforce a reseeding after a given number of bytes.

franziskuskiefer:
> @franziskuskiefer I was also wondering whether configuring the `reseed_interval` enables you to more conveniently reseed after N number of bytes as described in #1523. If I understand the HMAC-DRBG correctly, the `reseed_interval` only specifies after how many generate calls, each of which can generate 2^19 bits, reseed must be called. If the output length is not equal per generate call, the `reseed_interval` can't be used to enforce a reseeding after a given number of bytes.

I guess it needs to be extended to allow that, or we need to add another way of tracking it.
The main ask here was to lowering the number of bytes that are produced before reseeding. The default that's defined in the spec is pretty high.

Arul-Sujith:
I'll fix the changelogs and tests shortly.

For `set_reseed_interval(0)`, returning a `ReseedIntervalError` makes sense to avoid panics. 

@franziskuskiefer tracking the exact byte count would definitely be better for OpenMLS. Do you want byte tracking added in this PR or should we leave it for a follow-up?

robinhundt:
> The main ask here was to lowering the number of bytes that are produced before reseeding.

I looked at #1523 again and at the issues in rand involving `ReseedingRng`. There is this issue https://github.com/rust-random/rand/issues/1748 proposing to re-add `ReseedingRng`, with one person mentioning their simplified implementation of `ReseedingRng` for rand@0.20 ([reseeding-rng crate](https://crates.io/crates/reseeding_rng)).

I think using this wrapper (or one written by us) is the better option for ensuring that the HMAC-DRBG is reseeded after a given number of bytes.

#1523 mentioned:
> Today this is only achievable by wrapping the DRBG in our own struct that counts bytes and calls reseed_from_rng when a threshold is crossed. That works, but it duplicates state the DRBG already maintains and can't influence the DRBG's own needs_reseed() decision.

Which, from my reading, is not correct. The `HmacDrbg` currently does not track the number of bytes produced over multiple generate calls. As discussed, the `reseed_interval` tracked by the DRBG is the number of `generate` calls, not the number of bytes.

So we have three options as I see it:
1. Add the number of bytes produced to the state of the HMAC-DRBG and in its API offer the option to reseed after a given number of bytes
2. [reseeding-rng crate](https://crates.io/crates/reseeding_rng)
3. Vendor the reseeding-crate or write our own version

I would not choose option 1. It would be a non-standard extension of the HMAC-DRBG API and if we wanted to have the reseeding functionality for another rng, we'd have to duplicate the work.

I think I'm leaning towards option 3, as the crate is relatively young, does not have a lot of dependents (although it is used by the tor project), and having an external dependency could be a hindrance if we wanted to verify some code in the future that uses this dependency. The question is then, where should this code live? If the only user of such reseeding behavior is currently in openmls, I think it makes sense to implement the `ReseedingRng` there.

A separate question:  
Assuming we add a `RessedingRng` to openmls that achieves the goal of #1523, do we need the extension to HMAC-DRBG implemented in this PR? Are there other valid reasons/use-cases where one might want to manually set the `reseed_interval` of the HMAC-DRBG? If we don't know other uses for this, I'd propose not merging this at the moment, to not add API surface and state to the HMAC-DRBG which is not currently needed.

robinhundt:
I had a call with @franziskuskiefer to discuss this and we settled on the following:

We actually do not need a way of letting the user set the `reseed_interval` (meaning the number of generate calls) on the HMAC-DRBG. What we need is the ability to set a user-configurable number of *bytes* after which the drbg is reseeded. I'm sorry that this wasn't quite clear from the original issue description.

The way this should be implemented is as an additional wrapper type inside libcrux-hmac-drbg, similar to the wrappers already present [here][0]. I would propose adding a new wrapper:
```rust
struct ReseedingHmacDrbg<const OUTPUT_LEN: usize, Hmac: HmacAlgorithm<OUTLEN>, ReseedRng: TryCryptoRng {
  drbg: HmacDrbg<OUTLEN, Hmac>,
  rng: ReseedRng,
  bytes_consumed: usize,
  reseed_bytes_threshold,
}


impl TryRng for ReseedingHmacDrbg {
  type Error = ...;
  
  fn try_next_u32(&mut self) -> Result<u32, Self::Error> {
    // maybe put this in a separate method
    if self.bytes_consumed + 4 > selfreseed_bytes_threshold || self.drbg.needs_reseed() {
      self.drbg.reseed_from_rng(&mut self.rng, &[]).map_err()?;
      self.bytes_consumed = 0;
    }
    let mut out = [0; 4];
    self.drbg.generate(&mut out, &[]).map_err()?;
    Ok(out)
  }
}

impl TryCryptoRng for ReseedingHmacDrbg {}
```
This is a sketch, I've omitted the other required methods and error handling/propagation here. @franziskuskiefer This is different from the existing wrappers in `rand.rs` in that it uses a fallible `TryCryptoRng` and not an infallible `CryptoRng`. From the [usage in openmls][1], I assume that you want to pass the fallible `SysRng` and handle any errors.

EDIT: `HmacDrbg` already implements `TryRng`, so there is no need to replicate the out allocation, generate call, simply call the internal `try_next_u32`.

@Arul-Sujith would you be interested in implementing this wrapper in a new PR? I I'll close this PR as we do not need the `set_reseed_interval` method as implemented here. Sorry about that!


[0]: https://github.com/celabshq/libcrux/blob/e460d70a936de5b2f328d8c8357670334e31af8c/crates/algorithms/hmac-drbg/src/rand.rs#L251
[1]: https://github.com/openmls/openmls/blob/8545241346d3c11c3c5e566e4f87f6a35ad8529c/libcrux_crypto/src/crypto.rs#L52

PR #1546, opened 2026-07-21
use hmac-drbg as prng in libcrux provider (#1545)

### LINKED ISSUE #1545: Use drbg in hpke
Now that we have a DRBG again we should use it in HPKE.

https://github.com/celabshq/libcrux/blob/c5fb80f37530ee9b2df9501ae5ff8cb4a973a4bd/crates/protocols/hpke/libcrux_provider/src/lib.rs#L34-L38

Arul-Sujith:
This PR updates the `hpke-rs-libcrux` crypto provider to use the newly restored `libcrux-hmac-drbg` as its pseudo-random number generator,.

## Changes

- **Dependency Update**: Added `libcrux-hmac-drbg` (with the `rand` feature) to `crates/protocols/hpke/libcrux_provider/Cargo.toml`.
- **PRNG Swap**: Upgraded `HpkeLibcruxPrng` to store `HmacSha256DrbgRng<rand_chacha::ChaCha20Rng>` instead of just `ChaCha20Rng`.
- **Initialization**: Configured the new DRBG inside the `prng()` factory method, using the existing OS-seeded `ChaCha20Rng` as the underlying entropy source for automatic reseeding, and an empty array `&[0u8; 32]` for the personalization string.
- **Changelog**: Logged the HPKE DRBG integration in `CHANGELOG.md`.

All local tests, HPKE integration tests, KATs, and the broader workspace successfully compile and pass.

Fixes #1545

---

*AI Disclosure: Used an LLM to assist with this PR. Apply the DRBG struct replacement within the HPKE provider, and generate this PR description. Manually reviewed the modifications and ran the workspace test suite to verify the changes.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+
+- (hpke-rs-libcrux) [#1545](https://github.com/celabshq/libcrux/issues/1545): Use DRBG as the PRNG in the HPKE provider
+
 ## [0.0.5] (2026-07-15)
 
 ### Fixed
--- a/crates/protocols/hpke/libcrux_provider/src/lib.rs
+++ b/crates/protocols/hpke/libcrux_provider/src/lib.rs
@@ -34,7 +34,7 @@ pub struct HpkeLibcrux {}
 pub struct HpkeLibcruxPrng {
     #[cfg(feature = "deterministic-prng")]
     fake_rng: Vec<u8>,
-    rng: rand_chacha::ChaCha20Rng,
+    rng: libcrux_hmac_drbg::HmacSha256DrbgRng<UnwrapErr<SysRng>>,
 }
 
 impl Zeroize for HpkeLibcruxPrng {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- (hpke-rs-libcrux) [#1545](https://github.com/celabshq/libcrux/issues/1545): Use DRBG as the PRNG in the HPKE provider
```
franziskuskiefer: This should go into the unreleased section.

crates/protocols/hpke/libcrux_provider/src/lib.rs
```diff
@@ -34,7 +34,7 @@ pub struct HpkeLibcrux {}
 pub struct HpkeLibcruxPrng {
     #[cfg(feature = "deterministic-prng")]
     fake_rng: Vec<u8>,
-    rng: rand_chacha::ChaCha20Rng,
+    rng: libcrux_hmac_drbg::HmacSha256DrbgRng<rand_chacha::ChaCha20Rng>,
```
franziskuskiefer: I don't think need another rng in here. We just need something that collects entropy for reseeding. The `SysRng` is good for that. That way we can get rid of the chacha20rng.

franziskuskiefer (changes requested):
Thanks for the contribution.

I think we should use a slightly different setup.
While you're at it. You could also add a function next to `Hpke::new` that allows passing in an rng. That would make the entire thing a little more flexible.

Arul-Sujith:
Thanks for the review! 

- Moved the changelog entry to the unreleased section 
- Swapped out `ChaCha20Rng` for `SysRng` in the provider
- Added the `new_with_prng` constructor to the main `Hpke` struct for better flexibility. 

Let me know if that looks good!

crates/protocols/hpke/libcrux_provider/src/lib.rs
```diff
@@ -358,15 +358,19 @@ impl HpkeCrypto for HpkeLibcrux {
...
             }
         }
 
         #[cfg(not(feature = "deterministic-prng"))]
-        HpkeLibcruxPrng {
-            rng: rand_chacha::ChaCha20Rng::from_rng(&mut UnwrapErr(SysRng)),
+        {
+            let rng = UnwrapErr(SysRng);
+            HpkeLibcruxPrng {
+                rng: libcrux_hmac_drbg::HmacSha256DrbgRng::new(rng, &[0u8; 32]),
```
franziskuskiefer, line 372: Let's use some const bytes for the personalization. Something towards `b"libcrux hpke"`.

crates/protocols/hpke/libcrux_provider/src/lib.rs
```diff
@@ -34,7 +34,7 @@ pub struct HpkeLibcrux {}
 pub struct HpkeLibcruxPrng {
     #[cfg(feature = "deterministic-prng")]
     fake_rng: Vec<u8>,
-    rng: rand_chacha::ChaCha20Rng,
+    rng: libcrux_hmac_drbg::HmacSha256DrbgRng<UnwrapErr<SysRng>>,
 }
 
 impl Zeroize for HpkeLibcruxPrng {
```
franziskuskiefer, line 40: The comment in here doesn't make sense anymore.

franziskuskiefer (changes requested):
Thanks for the update. There's another place where the chacha rng is still used.

Can we use the drbg in `kem_key_gen_derand` as well?

The Chacha rng should not be needed anymore in the libcrux provider. There's also the `fake_rng` that still has it. That could be kept for dev builds if really neede. But it would be nicer if we could drop it there as well.

PR #1554, opened 2026-07-27
combined C extraction - mldsa/mlkem

franziskuskiefer:
On top of/replacing #1467.

- combined extraction of all ML-KEM and ML-DSA variants to C and C header only
- cleaned up tests
- added asan/ubsan

[skip changelog]


combined_extraction/extract-c.yaml
```diff
@@ -0,0 +1,377 @@
...
+        - [libcrux_sha3, sha256_ema]
+        - [libcrux_sha3, sha384_ema]
+        - [libcrux_sha3, sha44_ema]
+        - [libcrux_sha3, portable, sha224]
+        - [libcrux_sha3, portable, sha256]
+        - [libcrux_sha3, portable, sha384]
+        - [libcrux_sha3, portable, sha44]
+        - [libcrux_sha3, portable, keccakx1]
+        - [libcrux_sha3, portable, shake128]
+        - [libcrux_sha3, portable, shake256]
```
robinhundt: nit: I think this bundle is unused and not needed. The patterns of the `libcrux_sha3_portable` bundle already match everything that is listed here, due to the catch all pattern.

combined_extraction/extract-c-header.yaml
robinhundt: nit: Same issue as in extract-c.yaml, so this was pre-existing. Still worth to remove I think, to not have unneeded config in here.

combined_extraction/c/tests/nistkats_pre_hashed-44.json
robinhundt, line 1: **blocking**: This file doesn't seem to be referenced in any tests. Was the intention to use this for testing the pre-hashed APIs?

combined_extraction/c/tests/nistkats_pre_hashed-65.json
robinhundt, line 1: **blocking**: This is also unused.

combined_extraction/c/tests/nistkats_pre_hashed-87.json
robinhundt, line 1: **blocking**: This is also unused.

combined_extraction/c/CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+#         # We benchmark internal functions here that are inlined and thus not available
+#         # in MSVC.
+#         add_executable(sha3_bench
+#             ${PROJECT_SOURCE_DIR}/benches/sha3.cc
+#         )
+#         target_link_libraries(sha3_bench PRIVATE
+#             benchmark::benchmark
+#         )
+#     endif(NOT MSVC)
+# endif(DEFINED ENV{LIBCRUX_BENCHMARKS})
```
robinhundt: question: Do we expect to need this commented out code in the future? If yes, a short comment explaining why it is commented out and why we might need it again would be helpful.

combined_extraction/c/tests/mlkem.cc
```diff
@@ -0,0 +1,400 @@
...
+
+  memset(keygen_rand.data, 0x13, 64);
+  Eurydice_arr_ec encaps_rand = {};
+
+  memset(encaps_rand.data, 0x15, 32);
+
+  // We put this on the heap to avoid blowing the stack.
+  MLKEM_PORTABLE_UNPACKED_TYPE *key_pair = static_cast<
+      MLKEM_PORTABLE_UNPACKED_TYPE *>(malloc(sizeof(MLKEM_PORTABLE_UNPACKED_TYPE)));
+  MLKEM_SYM(_portable_unpacked_generate_key_pair_mut)(keygen_rand, key_pair);
```
robinhundt: question: This directly calls generate_key_pair on the uninitialized memory returned by malloc. It seems the `init_key_pair` functions can't be called here, as they return a default initialized key-pair on the stack and don't take a ptr as a parameter. I'm wondering, why is this key pair allocated on the heap in the first place? It shouldn't be so big that it blows the stack.

combined_extraction/c-header-only/libcrux_mldsa44_avx2.h
```diff
@@ -0,0 +1,177 @@
...
+ * SPDX-FileCopyrightText: 2025 Cryspen Sarl <info@cryspen.com>
+ *
+ * SPDX-License-Identifier: MIT or Apache-2.0
+ *
+ * This code was generated with the following revisions:
+ * Charon: 6f058254eb741c12e9b388df07adaf7cc8aac8ed
+ * Eurydice: fca2e9fbd728e49d677f3fc0da0054b55f3b9973
+ * Karamel: 8c19d41458ce5cbfea029ebc03334ba96d149039
+ * F*: 70671ffb81fa30aba09b9d6e2af275dfbccaa8f8
+ * Libcrux: 97ec9733b27531975cb58960a1a3049193a43e07
```
robinhundt: question: was there a reason for doing the combined extraction with different versions of the tools than we currently use in CI? E.g. CI uses eurydice at aaa9fa657fb6f09802edb890252040d94cd93982 and charon at e656e17bff6ca5efac8ab6919b9b74cb9a8dd8ad

combined_extraction/c/CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+        add_compile_options(/fsanitize=address)
+        message(STATUS "AddressSanitizer ENABLED (MSVC)")
+    endif()
+endif()
+
+if((CMAKE_C_COMPILER_ID STREQUAL "Clang" AND
+    CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "13.0.0") OR
+    (CMAKE_C_COMPILER_ID STREQUAL "AppleClang" AND
+    CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "13.1.6"))
+    # add_compile_options(-Werror -Wframe-larger-than=25344)
```
robinhundt: **blocking**: If we enable the -Werror flag, we get hundreds of `-Wstatic-in-inline` violations in the split header extraction. These are caused by the `KRML_MUSTINLINE` attribute expanding to `inline __attribute__((always_inline))` which has external linkage. But some of these functions use other functions with `static inline` (e.g. in the glue) that have internal linkage. This violates constraint [6.7.4p3][0] of the C11 standard. As far as I understand, this is not huge problem. At the moment, it seems to work because the declarations of these functions in the header do not have an inline annotation. If these were added there, this would likely cause linker errors.

I think this might be possible to fix upstream in karamel by changing `KRML_MUSTINLINE` to also emit `static` but I'm not 100% sure on the ramifications of this.

For our combined extraction, I think it would be good to re-enable `-Werror`, but also specify `-Wno-static-in-inline` with a comment explaining why and `-Wno-unused-function` (eurydice emits some functions that are unused). Then we don't accidentally overlook important errors in the future.

[0]: https://port70.net/~nsz/c/c11/n1570.html#6.7.4p3

robinhundt: question: It seems that the current frame limit is ~300 KB for ML-DSA 87 signing. This seems like quite a lot. Is it expected that the stack usage is so high? Should we enable the `-Wframe-larger-than` option with a high enough thresholds to catch further increases in the future?

combined_extraction/c/CMakeLists.txt
robinhundt, line 1: question: Would it make sense to register the tests with `add_test` here? I know it exists, but don't know about the trade-offs of registering tests in the cmake files.

combined_extraction/c/tests/util.h
robinhundt, line 89: nit: Might be worth asserting here that kats is non-empty and test vectors have actually been loaded. I'm not sure whether a wrong path could silently lead to empty vectors at the moment.

combined_extraction/c/tests/mldsa.cc
```diff
@@ -1,40 +1,77 @@
...
+#elif MLDSA_VARIANT == 65
 #include "libcrux_mldsa65_portable.h"
+#define MLDSA_VERIFICATION_KEY_SIZE 1952U
+#define MLDSA_SIGNING_KEY_SIZE 4032U
+#define MLDSA_SIGNATURE_SIZE 3309U
+#elif MLDSA_VARIANT == 87
+#include "libcrux_mldsa87_portable.h"
+#define MLDSA_VERIFICATION_KEY_SIZE 2592U
+#define MLDSA_SIGNING_KEY_SIZE 4896U
+#define MLDSA_SIGNATURE_SIZE 4627U
```
robinhundt: question: are these constants not exported by the extracted Libcrux? I wasn't able to find them with a quick grep.

combined_extraction/c-header-only/libcrux_mldsa44_avx2.h
```diff
@@ -0,0 +1,177 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Cryspen Sarl <info@cryspen.com>
```
robinhundt: nit: This still says Cryspen.

combined_extraction/c/CMakeLists.txt
robinhundt: Additional note: It seems the static-in-inline warnings are only emitted when compiling with clang and not gcc. It is probably safe to ignore this.

robinhundt (dismissed):
Thanks for the PR! I did a manual + Claude assisted review of the PR. I can locally extract and build and run the test binaries. I've also had claude run a more comprehensive matrix of build and compiler configurations inside a VM. The only issue was the c-header-only version under gcc 14 with sanitizers enabled. For some reason that configuration seemed to exhaust the available RAM in the VM during linking. The debug tests for that configuration were built and ran partially, the release ones did not. All other configurations worked.

CI: I think it would be good to build and test the extraction in CI on changes to the extracted files. If we re-extract at some point, those tests will be automatically run and we don't have to rely on manual testing which can be forgotten. 

I would probably not add a job that checks that the current extraction of the rust code matches the committed extraction. This job is quite annoying for the individual extractions and I think it makes sense to deliberately update extractions when needed, as opposed to for every rust change.

combined_extraction/c/tests/nistkats_pre_hashed-44.json
franziskuskiefer: We don't test pre-hashed on the C side anywhere right now

combined_extraction/c/CMakeLists.txt
franziskuskiefer: Yes it's expected to be pretty high. Unless someone cares about this, I wouldn't touch this.

combined_extraction/c/tests/nistkats_pre_hashed-44.json
franziskuskiefer: Added

combined_extraction/c/CMakeLists.txt
franziskuskiefer: I added something. It's not really something we need to focus on though.

combined_extraction/c/CMakeLists.txt
franziskuskiefer: done, but it's not really used

combined_extraction/c-header-only/libcrux_mldsa44_avx2.h
franziskuskiefer: we should update. but that's not for this pr, using the older version

combined_extraction/c/tests/mlkem.cc
franziskuskiefer: doesn't make a real difference other than making the stack bigger. changed. let's see if it makes trouble.

robinhundt (approved):
Great! Thanks for adding the CI config and the other changes.

PR #1558, opened 2026-07-28
[hmac-drbg]: fix panic in reseeding wrappers fill_bytes method

### LINKED ISSUE #1451: [hmac-drbg/CI] Build & test on CI

### LINKED ISSUE #1556: [hmac-drbg] rand auto-reseeding wrappers fill_bytes panics for outputs divisible by `MAX_GENERATE_BYTES` (65_536)
The HMAC-DRBG wrappers in `libcrux-hmac-drbg` which auto-reseeds after the maximum reseed interval has been reached panic for outputs whose length is divisible by `MAX_GENERATE_BYTES` (65_536).

The Wrappers here:
https://github.com/celabshq/libcrux/blob/e460d70a936de5b2f328d8c8357670334e31af8c/crates/algorithms/hmac-drbg/src/rand.rs#L254-L262

Implement `try_fill_bytes` by chunking the bytes into chunks of size `MAX_GENERATE_BYTES` and call `safe_generate_small` on the chunks and the rest:
https://github.com/celabshq/libcrux/blob/e460d70a936de5b2f328d8c8357670334e31af8c/crates/algorithms/hmac-drbg/src/rand.rs#L375-L382

`safe_generate_small` calls `self.drbg.update` and calls unreachable if that method returns a `RequestInvalid` error:
https://github.com/celabshq/libcrux/blob/e460d70a936de5b2f328d8c8357670334e31af8c/crates/algorithms/hmac-drbg/src/rand.rs#L345-L353

Contrary to the comment, it's *not* fine, as this error is returned if `output` is empty. This happens when the `as_chunks_mut` rest is empty when the output is divisible by `MAX_GENERATE_BYTES`, leading to the `unreachable` being reached and a panic.

The following tests fail with a panic:
```rust
#[test]
fn sha256_fill_bytes_panics() {
    let mut rng = make_sha256();
    let mut output = [0; libcrux_hmac_drbg::MAX_GENERATE_BYTES];
    rng.fill_bytes(&mut output);
}

#[test]
fn sha384_fill_bytes_panics() {
    let mut rng = make_sha384();
    let mut output = [0; libcrux_hmac_drbg::MAX_GENERATE_BYTES];
    rng.fill_bytes(&mut output);
}

#[test]
fn sha512_fill_bytes_panics() {
    let mut rng = make_sha512();
    let mut output = [0; libcrux_hmac_drbg::MAX_GENERATE_BYTES];
    rng.fill_bytes(&mut output);
}
```

```
thread 'sha512_fill_bytes_panics' (255441) panicked at crates/algorithms/hmac-drbg/src/rand.rs:350:58:
internal error: entered unreachable code
```

robinhundt:
fixes #1556
closes #1451

I chose to remove the error variant in the case of 0 byte output generate calls. From my reading of [SP 800-90A][0], this is not disallowed and I see no reason why it should be. This alone would have fixed the bug, but would have caused different behavior between the reseeding wrapper for outputs divisible by the max amount and the non re-seeding `HmacDrbg::try_fill_bytes`. 

I've also added fuzzing for the reseeding wrappers (generated by claude opus 5 and revised by me) which catches the panic and is green after the fix. I also fixed the `new` fuzz target which had a fault.

I noticed that there no CI tests yet for hmac-drbg (#1451 was still open), so I added a basic CI workflow including fuzzing based on the aes workflow.

One question I have: Can the hax annotations that are present be consumed by hax or are they just comments intended for future annotations with `hax_lib`? One of the hax annotations was wrong.

[0]: https://nvlpubs.nist.gov/nistpubs/specialpublications/nist.sp.800-90ar1.pdf

### DIFF
--- a/.github/workflows/hmac-drbg.yml
+++ b/.github/workflows/hmac-drbg.yml
@@ -55,11 +55,11 @@ jobs:
 
       - name: 🏃🏻‍♀️ Test
         run: |
-          cargo hack test --each-feature --exclude-features testing --verbose
+          cargo hack test --each-feature --exclude-features _testing-apis --verbose
 
       - name: 🏃🏻‍♀️ Test Release
         run: |
-          cargo hack test --each-feature --exclude-features testing --verbose --release
+          cargo hack test --each-feature --exclude-features _testing-apis --verbose --release
 
   fuzz:
     strategy:
--- a/.github/workflows/hmac-drbg.yml
+++ b/.github/workflows/hmac-drbg.yml
@@ -1,0 +1,103 @@
+name: HMAC-DRBG
+
+permissions: {}
+
+on:
+  merge_group:
+  pull_request:
+    branches: ["main", "dev", "*"]
+    paths:
+      - "crates/algorithms/hmac-drbg/**"
+      - .github/workflows/hmac-drbg.yml
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          - windows-latest
+
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+        working-directory: crates/algorithms/hmac-drbg
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@43aecc8d72668fbcfe75c31400bc4f890f1c5853 # v2.83.2
+        with:
+          tool: cargo-hack
+
+      - name: 🔨 Build
+        run: |
+          cargo build --all-targets --verbose
+
+      - name: 🔨 Build Release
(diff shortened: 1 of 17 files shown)


jschneider-bensch:
> One question I have: Can the hax annotations that are present be consumed by hax or are they just comments intended for future annotations with `hax_lib`? One of the hax annotations was wrong.

At the moment, these are just comments. My understanding was that they could be useful for future annotations, like you hypothesize.

.github/workflows/hmac-drbg.yml
```diff
@@ -0,0 +1,103 @@
...
+          cargo build --all-targets --verbose
+
+      - name: 🔨 Build Release
+        run: cargo build --release --all-targets --verbose
+
+      # Test ...
+
+      - name: 🏃🏻‍♀️ Test
+        run: |
+          cargo hack test --each-feature --exclude-features testing --verbose
```
jschneider-bensch: As I understand it the `testing` feature does not provide extra tests, and is only used in fuzzing to provide non-production APIs for resetting state etc, so it makes sense to exclude it here. But maybe if it were named `testing-apis` it would be clearer why it's okay to exclude it when testing.

jschneider-bensch (approved):
Thanks!

.github/workflows/hmac-drbg.yml
robinhundt: Done! I wanted to rename it anyway to prefix it with `_` so that it doesn't show up on docs.rs but forgot.

jschneider-bensch: Great, thanks! :+1:

PR #1561, opened 2026-08-03
hmac-drbg: use chunks_mut for try_fill_bytes

robinhundt:
Was there a reason why the initial implementation was not already done with chunks_mut? 

[skip changelog]


robinhundt:
Stacked on #1558 . I wanted to try the new stacked PR feature, but it seems we don't have access yet? I don't see an option to convert this into a stack.

jschneider-bensch (approved):

jschneider-bensch:
I'm not aware of a specific reason not to use `chunks_mut`. The original version with `as_chunks_mut` seems to be the same amount of trouble in a potential hax extraction. :)

Issue #1544, opened 2026-07-20
libcrux-provider compatibility with wasm32-unknown-unknown?


rocodes:
hi team, thank you for all your work on libcrux.

If I understand correctly, https://github.com/celabshq/libcrux/commit/8dfe4ec38a5a82cda5af85d7667fa68bae88e3b1 introduced a required `sys_rng` / getrandom dependency to hpke-libcrux-rs.  We build a crate that uses the libcrux provider and targets multiple architectures including `wasm32-unknown-unknown`, and ideally we'd avoid `rand` and `getrandom` in that crate and let the consumer crates that we are building supply their own architecture-specific randomness choice rather than require it in our shared crate.

Is this something you'd be open to supporting again in libcrux_provider? Thank you and let me know if you need any more information!

jschneider-bensch:
Would the `HpkeLibcruxPrng::from_seed` and `Hpke::new_with_prng` functions from #1562 help in this regard? That PR doesn't touch the mandatory `sys_rng` dependency, though.

rocodes:
Hi @jschneider-bensch, thank you for your reply!  I could have been wrong about the specific PR, the main issue is the mandatory dependency and whether no_std support for `hpke-libcrux-rs` is something you'd support as you do with the libcrux-rs crate and others in this repo, so that hpke-libcrux-rs can be randomness-agnostic and the crates that use it can supply their own randomness source.

Our use-case is a no_std [`protocol`](https://github.com/freedomofpress/securedrop-protocol/tree/main/securedrop-protocol/protocol-minimal) crate that uses hpke-libcrux-rs among other dependencies, and can in turn be used by different client crates meant for different target architectures including wasm32-unknown-unknown and x86_64-unknown-linux-gnu.

jschneider-bensch:
We're definitely interested in making as much of libcrux and hpke-rs usable in no_std configurations as possible. :+1:

Would you want to avoid `rand` entirely, or would a no_std compatible `rand` dependency work for you as well? I.e. if the `sys_rng` feature was only optionally enabled, would that be enough?
Currently `HpkeCrypto` has the `rand_core::CryptoRng` bound on its `HpkePrng` associated type.

rocodes:
Thanks for the reply, this is great to hear, and yep we also use `rand_core::CryptoRng`, no_std compatible rand would be just fine :)

PR #1562, opened 2026-08-04
hpke: wasm crate for web demo

franziskuskiefer:
This is what powers https://celabs.eu/blog/pq-hpke/.

This should also address #1544.
There's some overlap with #1546.

[skip changelog]


crates/protocols/hpke/wasm/pkg/index.html
```diff
@@ -0,0 +1,355 @@
...
+        hpke_open,
+      } from "./hpke_rs_wasm.js";
+
+      await init();
+
+      // The ciphersuite in use. Built from the dropdown selection below, which
+      // defaults to Base / DHKEM(X25519, HKDF-SHA256) / HKDF-SHA256 / ChaCha20Poly1305.
+      let config;
+
+      function update_config() {
```
jschneider-bensch, line 218: This function should reset all the input field as well, otherwise the app can crash when trying to seal to a left-over, incompatible public key for a different config.

crates/protocols/hpke/wasm/pkg/index.html
```diff
@@ -0,0 +1,355 @@
...
+      document.getElementById("suite").onchange = update_config;
+      update_config();
+
+      function seal() {
+        let info = new TextEncoder().encode(document.sender_form.info.value);
+        let aad = new TextEncoder().encode(document.sender_form.aad.value);
+        let pt = new TextEncoder().encode(document.sender_form.payload.value);
+        let pkR = hex_to_bytes(document.receiver_form.pkR.value);
+
+        let rand = new Uint8Array(32);
```
jschneider-bensch: I don't understand what this is used for.

jschneider-bensch (dismissed):

crates/protocols/hpke/wasm/pkg/index.html
franziskuskiefer: It's not used anymore. We did the sampling in here before and I didn't delete it apparently.

jschneider-bensch (approved):
This looks good to me, now, thanks!

Issue #1341, opened 2026-02-24
Check usage of `cfg(not(hax))` vs `hax_lib::exclude`


jschneider-bensch:
> do we have other crates where we should fix this? Also, should we add a dummy crate (or get #1259 in) to have a test in the repo to make sure we don't break this again?

It is worth checking to avoid dependency trouble in the future, for sure. We should get #1259 as well.

_Originally posted by @jschneider-bensch in https://github.com/cryspen/libcrux/issues/1312#issuecomment-3851538561_

jschneider-bensch:
It would still be preferrable to use `hax_lib::exclude` over `cfg(not(hax))` when possible. Situations where it may not be possible is when we have mutually exclusive code paths for hax extraction and otherwise.

PR #1578, opened 2026-09-02
Make hax-lib a `target.'cfg(hax)'` dependency

### LINKED ISSUE #1575: Make `hax-lib` a target dependency
The hax/hax-lib release 0.4 will include https://github.com/cryspen/hax/pull/2117, which should make it possible to replace the unconditional hax-lib dependency we currently have in many crates with a `target.'cfg(hax)'` dependency.

For Lean extractions there might be some issues still with the `cfg` handling (https://github.com/cryspen/hax/pull/2069) but for F*, everything should work with `cfg(hax)`.

jschneider-bensch:
This PR makes `hax-lib` a `target.'cfg(hax)` dependency in `libcrux-intrinsics`, `libcrux-secrets`, `libcrux-sha3`, `libcrux-kmac`, `libcrux-ml-kem`, `libcrux-ml-dsa` and the main `libcrux` crate.
The one other crate that depends on `hax_lib`, `core-models`, is already a `target.'cfg(hax)` dependency in `libcrux-ml-dsa`, the only place where it is used. Therefore I did not touch it.

Default-feature builds on my machine of the crates listed above seem to work. So far so good!

TODO:
- [x] Validate functional builds beyond default features and for all target architectures. (I don't anticipate problems here, but want to make sure.)
- [x] Validate `cfg(hax)` builds still work and result in good extractions to F*.
- [ ] Look into #1341 while checking on the quality of F* extractions

Resolves #1575 

[skip changelog]

### DIFF
--- a/.github/workflows/kmac-hax.yml
+++ b/.github/workflows/kmac-hax.yml
@@ -18,7 +18,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   FSTAR_REV: v2026.03.24
-  HAX_LIB_VERSION: 0.3.7
+  HAX_LIB_VERSION: 0.4.0
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/mldsa-hax.yml
+++ b/.github/workflows/mldsa-hax.yml
@@ -16,7 +16,7 @@ on:
 env:
   CARGO_TERM_COLOR: always
   FSTAR_REV: v2026.03.24
-  HAX_LIB_VERSION: 0.3.7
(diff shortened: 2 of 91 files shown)


jschneider-bensch:
I've bumped the hax-lib version to 0.4.1 and rebased to main. Since the fix for https://github.com/cryspen/hax/issues/2243 is included in this new release I've also undone the hax exclusions that were introduced in ML-KEM in #1584.

jschneider-bensch:
Looks like there are some issues in SHA3 with Bundles that weren't extracted on previous versions.

PR #1579, opened 2026-09-02
[psq] Align `nonce-control` handshake nonces with default nonces

### LINKED ISSUE #1577: `nonce-control` affecting handshake keys
The idea behind `nonce-control` feature flag was to allow the caller manual nonce control over **transport** keys (via `Transport::set_sender_nonce`/`set_receiver_nonce`) making the `increment_nonce()` into a no-op. However, this no-op turned to be global on all calls to `AEADKeyNonce::increment_nonce()` which make it also apply to the **handshake** keys (k0/k1/k2), which changes the handshake wire format: a peer built with `nonce-control` cannot complete a handshake with a peer built without it.

## Details

Handshake keys are derived fresh with nonce `[0u8; 12]` (`derive_k0`/`derive_k1_*`/`derive_k2_*` -> `AEADKeyNonce::new`). All handshake AEAD operations go through `handshake_encrypt`/`handshake_decrypt` -> `encrypt`/`decrypt` -> `increment_nonce()`:

- without `nonce-control`: increment-before-use, so every handshake payload is sealed/opened at nonce **1**;
- with `nonce-control`: the increment is a no-op and nothing ever sets a handshake key's nonce, so every handshake payload uses nonce **0**.

The result is a bidirectional interop failure: e.g. a `nonce-control` initiator seals msg1's outer payload with k0 at nonce 0, and a default-build responder's `handshake_decrypt` opens at nonce 1, so the tag check fails and the handshake aborts before msg2 is produced. The reverse direction fails identically,
as do k1 (inner payload) and k2 (msg2).

There is no way to compensate from outside the crate: `set_nonce` is only exposed on `Transport`, and the handshake keys are derived and consumed inside private methods with no interception point.

## Suggested fix

Scope the feature to transport keys. The seam is already clean: transport uses `encrypt`/`decrypt_out` directly, while the handshake exclusively uses the `handshake_encrypt`/`handshake_decrypt` wrappers. Since handshake keys are one-shot, having those two wrappers set the nonce to 1 (or increment
unconditionally) under `nonce-control` restores byte-compatibility with default builds at no security cost.

jschneider-bensch:
This follows the suggested fix from #1577, by explicitly setting the nonce to an aligned value in `handshake_encrypt, handshake_decrypt` when feature `nonce-control` is used.
I'll file a follow-up issue about adding interoperability tests for different feature sets.


libcrux-psq/src/aead.rs
```diff
@@ -10,8 +10,23 @@ use tls_codec::{
 /// Length of an AEAD nonce in bytes.
 pub const NONCE_LEN: usize = 12;
 
+/// If not using feature `nonce-control`, nonces are incremented
+/// automatically on encryption/decryption operations until this value
+/// is reached, at which point the nonce must no longer be used.
 #[cfg(not(feature = "nonce-control"))]
 const NONCE_MAX: [u8; NONCE_LEN] = [0xff; NONCE_LEN];
```
franziskuskiefer, line 17:
```suggestion
const NONCE_MAX: [u8; NONCE_LEN] = [0xff; NONCE_LEN];

```

franziskuskiefer (approved):
Thanks, lgtm.

But can you add a test before merging?

PR #1581, opened 2026-09-03
[psq] Integration test for wire format agreement (`nonce-control`)

jschneider-bensch:
An integration test for interoperability between `nonce-control` and non-`nonce-control` peers. This only exercises one ciphersuite, but the AEAD usage is the same across ciphersuites and nonce handling is in turn the same across AEAD choices.

Generated by LLM and reviewed by me.

[skip changelog]


franziskuskiefer (approved):
:shipit:

PR #1584, opened 2026-09-09
Fallible rng

franziskuskiefer:
This PR allows fallible rngs in some more places. We should do it everywhere and also update the code underneath to actually propagate errors. But this is a first, non-intrusive pass.

This also adds some DER functionality so that we can actually use the P256 EcDSA crate.

PR #1585, opened 2026-09-10
DER for p256 ecdsa

franziskuskiefer:
Minimal DER de/encoding for p256 ecdsa signatures.

Split out of #1584 because it really didn't belong there.


crates/algorithms/ecdsa/src/der.rs
```diff
@@ -0,0 +1,264 @@
...
+///
+/// Long-form lengths are rejected outright. DER requires the minimal encoding,
+/// so the long form only ever encodes a length of at least 128 — which no
+/// P-256 `ECDSA-Sig-Value` can reach: an INTEGER that long has more than 32
+/// content bytes, and a SEQUENCE that long cannot be filled by two such
+/// INTEGERs.
+fn read_len(bytes: &[u8], pos: &mut usize) -> Option<usize> {
+    let first = *bytes.get(*pos)?;
+    *pos += 1;
+    if first < 0x80 {
```
jschneider-bensch: Could you make this stricter by giving the maximum length as a const generic? E.g. for the sequence length in `der_to_raw` you really expect the length to be at most `MAX_DER_LEN - 2` and for the individual integer lenghts in `read_integer_32` you expect them to be at most `33`.

The check for `value.len() > 32` in `read_integer_32` should ensure that there can't be one invalid overly long value balanced with a short value, but you could such a case here already.

crates/algorithms/p256/tests/wycheproof.rs
jschneider-bensch, line 1: Why are we testing ecdsa a second time here in the first place?

jschneider-bensch (approved):
I think it looks good.

crates/algorithms/ecdsa/src/der.rs
franziskuskiefer: done

crates/algorithms/p256/tests/wycheproof.rs
franziskuskiefer: I guess because it was there I just redid it ... I removed it.

PR #1586, opened 2026-09-10
[p256] const visibility change and derive_ecdh doc addition

robinhundt:
This PR makes the constants that are used in the `Ecdh` trait implementations public.

It also adds a comment on the `derive_ecdh` method that clarifies that this returns the complete point `x||y`.

While this now documents what the method returns, I would still argue for changing the return of this method to just provide `x`. This would be in-line with the definition in NIST SP 800-56 ([5.7.1.2][sp]) or SEC 1 ([3.3.1][sec1]) and would then reflect what is stated on the `Ecdh` trait,
https://github.com/celabshq/libcrux/blob/5caed1132ad6f0f10f329a2fe665b7a6ca9b8fa5/traits/src/ecdh/arrayref.rs#L35-L36

as the shared secret is defined as `x` in the linked references.

[sp]: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf
[sec1]: https://www.secg.org/sec1-v2.pdf


jschneider-bensch (approved):
Thank you for adding the documentation! :+1:

PR #1591, opened 2026-09-22
[platform] fix x86 data race and latent feature detection bugs

robinhundt:
Based on #1590 

The version of the previous PR used a spin loop to guard access to the CPU_ID static mut. In cases where the thread holding the lock is descheduled, this could lead to extensive busy waiting of the other threads. This PR takes a different approach by having multiple threads potentially race to execute `cpuid` and overwrite their results atomically.
This is okay, as each thread executing `cpuid` will return the same results and atomic overwrites of a `static` are allowed.

This also fixes two other latent bugs in the previous implementation:
- The code now checks whether the target is Intel SGX and returns no feature bits in that case
- The code now checks whether cpuid is implemented (max_basic_leaf > 0)
- The code first checks whether extended features are available (max_basic_leaf >= 7) before executing `__cpuid_count(7, 0)`

The implementation is partially based on the one in the [Rust standard library](https://github.com/rust-lang/rust/blob/e5b95097d9a14bdec7cd9101dde67ee3aad2578a/library/std_detect/src/detect/os/x86.rs#L27).


franziskuskiefer (approved):
let's get this in

PR #1592, opened 2026-09-22
[platform] fix avx/avx2 missing xgetbv check

robinhundt:
While working on #1591 I other pre-existing issue in the x86 feature detection code.

The checks only checked for the existence of the xsave/osxsave, but not the return of `xgetbv`, which is required to check whether the OS actually saves the registers for e.g. AVX2 instructions. This could have previously caused an issue on a machine/VM/hypervisor where the hardware reports xsave, but actual saving of the registers is not enabled by the OS, leading to a fault.

For avx2, the previous code did not even check whether `xsave/osxsave` was enabled. This is done now by checking for `avx` in the `avx2` branch which implies `xsave/osxsave`.

I suspect there are other issues with the avx512 handling which are not reachable today.


franziskuskiefer (approved):
We probably won't see a difference. But it doesn't hurt.

PR #1597, opened 2026-09-22
[sha3] Enforce correct digest length for `hash::<LEN>`

jschneider-bensch:


franziskuskiefer (commented):
I'm not sure I'm super convinced by this one actually. I don't like this function. These failure cases are real but panicking isn't great either. Making the function fallible is also not great though. But maybe that's what it should be

jschneider-bensch:
I'm fine with leaving this out since it's a minor change anyway. The F* failure on CI is interesting though.

PR #1598, opened 2026-09-22
[hmac] Fix documentation of output length of `hmac`

jschneider-bensch:
[skip changelog]


franziskuskiefer (approved):

jschneider-bensch:
I've changed the base to main, and merged here the two approved PRs #1599 and #1600. The actual change that was present here was obsoleted by the HMAC change in #1603. I'll thus update the description here.

PR #1602, opened 2026-09-22
hpke: update rust-crypto provider dependencies

### LINKED ISSUE #1583: [hpke-rs-rust-crypto] macOS builds don't use AES HW-accelerated instructions
The current version of the HPKE rust crypto provider relies on the version 0.10 of the `aes-gcm` crates, which itself relies on the version 0.8 of the `aes` crate. As a result, when building `hpke-rs` with Rust crypto and relying on AES-GCM, by default, the fixed-slice software-only implementation of AES is used, with suboptimal performances. 

A possibility consists in bumping the `aes-gcm` dependency to 0.11, which itself depends on `aes` v0.9 that implements underlying HW capabilities auto-detection on macOS.

In the meanwhile, a workaround consists in explicitly setting a configuration flag at the compiler level, possibly with side effects wrt to other crates: 
```
[target.aarch64-apple-darwin]
rustflags = ["-C", "target-cpu=native", "--cfg", "aes_armv8"]
```

robinhundt:
closes #1583


crates/protocols/hpke/rust_crypto_provider/CHANGELOG.md
jschneider-bensch, line 1: I think these don't necessarily need changelog entries, or can you say why that would be preferrable?

crates/protocols/hpke/rust_crypto_provider/Cargo.toml
```diff
@@ -12,26 +12,32 @@ repository = "https://github.com/celabshq/libcrux"
 [dependencies]
 hpke-rs-crypto.workspace = true
 # Rust crypto
+elliptic-curve = { version = "0.14" }
```
jschneider-bensch, line 15: Is it necessary to pull this in as a new dependency? AFAICT all of the curve crates re-export `sec1::ToSec1Point` and `Generate`.
Perhaps this version is preferrable because it makes explicit where the common traits originate.

jschneider-bensch (approved):
Thanks!

crates/protocols/hpke/rust_crypto_provider/Cargo.toml
robinhundt: Previously it was also a dependency, but just not direct. The code before used the re-exported elliptic-curve crate from `k256`, but then used these traits for the other crates like `p256` as well. This can make it a bit harder to diagnose version mismatches between the different versions of elliptic-curve used by the crates.

Having it as an explicit dep makes it more clear I think.

crates/protocols/hpke/rust_crypto_provider/CHANGELOG.md
robinhundt: These were updates of the dependencies to new breaking releases. I'm not sure whether that requires us to also make a breaking release next, but I think it is good to have it documented in the changelog.

